PR #10, opened 2020-05-26
Add CORS headers for production, resolve #8

### LINKED ISSUE #8: Adjust CORS settings
Adjust cors settings to allow remote frontend urls (in dev and prod).

Ni42:
Add the headers for the relevant URLs for deployment. Also add the new planned dashboard URLs.

### DIFF
--- a/web/index.ts
+++ b/web/index.ts
@@ -1,11 +1,11 @@
 import * as express from "express";
-import * as path from "path";
 import * as fs from "fs";
 import * as http from "http";
 import * as https from "https";
 import * as bodyParser from "body-parser";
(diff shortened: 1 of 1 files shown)


Ni42:
Hmm, yeah. In that case we should really only allow the corona-school.de domains for production, since the Heroku domains are (supposedly) dev only.

BlobbyBob:
Yeah. I don't think that non-production software should work on the production database. We might instead just deploy the dev branch as dev version of the backend working on a different database.

Ni42:
Lets split like this then I guess.

Issue #5, opened 2020-05-26
Add route GET /courses


BlobbyBob:
According to specification

PR #20, opened 2020-05-28
Route GET /courses

BlobbyBob:
Implements #5


BlobbyBob:
A new router for courses needs to be implemented. I didn't do this, so we won't run into to many conflicts when merging

PR #21, opened 2020-05-28
Add maxRecipients to course entity

Ni42:


Ni42:
No idea, massive brainlag I guess.

Issue #7, opened 2020-05-26
Add route POST /course


BlobbyBob:
According to specification

PR #23, opened 2020-05-30
POST /course

BlobbyBob:
Implements #7


BlobbyBob:
**We should merge #20 before this!**

BlobbyBob:
Still waiting for category list. If it doesn't arrive soon, we should just merge and fix later.

Issue #16, opened 2020-05-27
Implement certificate pdf modification


BlobbyBob:
Wrapper code is already implemented, only the modification of the PDF is missing.

**Edit**: We also need to adjust the route; all the required data should be included in the query parameters.

PR #25, opened 2020-05-30
Certificate generation

BlobbyBob:
Implements #16


Ni42:
(TODO: fix layout bugs in HTML source file)
(TODO: length checks for string arguments, parseInt for the rest)

PR #39, opened 2020-06-02
Feat/dummy courses

LeonErath:


BlobbyBob:
@LeonErath Since the entities are now updated, you can rewrite your sample data to fit the new format.

PR #56, opened 2020-06-09
Add HTTP 409 for existing users

Ni42:


Ni42:
I guess it's fine like this, or do we want to make it a completely optional field?

BlobbyBob:
I guess it's fine

PR #59, opened 2020-06-14
Screening reminders

geroembser:
This pull request will implement screening reminders as specified in corona-school/project-user#54. 

This implementation of screening reminders will use the refactoring of the jobs part and I hope that it is therefore easier to understand. At the same time, someone might use this as some kind of template for implementations of similar jobs that should automatically send emails. But depending on how many of what kind of emails need to be scheduled and sent automatically, it is probably recommendable to further simplify that process of implementing automatic emails.


geroembser:
Good point @BlobbyBob, but I’m not a law expert. That is probably important for a couple of emails we’re sending, especially the reminders (e.g. screening or upcoming courses, i.e. corona-school/project-user#54 or corona-school/project-user#25).

I did some quick Googling and research. And it seems sufficient if someone can just write us an email in reply, informing us that she wants her Corona School account to be deactivated.
But what every user can also always do is fairly simple: just go into the user-section/dashboard my.corona-school.de and deactivate the account there. Someone whose account is disabled, will of course never again receive emails from Corona School. Furthermore, I also believe that account deactivation is associated with the cancellation of the screening reminders. 

But I think, that we’ll also need to differ between two kind of emails, i.e. marketing emails and those that are inherent to our „business“ and what the Corona School does. If someone registers for Corona School, she should know that a screening interview is necessary to be part of Corona School. So as long as someone is registered for Corona School, I think it’s legitimate to expect them to appear for a screening interview. And if someone is no longer interested he can always deactivate his account and even if they don’t deactivate it, they will receive a maximum of 5 screening reminders. The screening is not an additional feature or product we’re selling, but is instead an important, probably the most important part, of the service we offer. The screening process is essential for every registration and for every _active_ user! All active users on our platform that want to teach 1-on-1 are forced be screened. Thus I personally don’t think that we should differentiate between deactivation of accounts and log-out of screening reminders.

But what I have done now ist that I have added a small hint to the footer of that email template, referring to that relationship of an active account in a non-screened state and the reminders, also noticing that someone can deactivate his account at every time (in his user section or by writing us an email). 

@tobiasbork what do you think from the product’s perspective? Is that legitimate? Or do we need a special one click solution to disable the account directly from a screening reminder mail? Or probably even cancel screening reminders while still having an active account?

PR #60, opened 2020-06-14
Add mail templates for courses

geroembser:
This pull request will add the required **templates** for mails that were sent in context of Corona School's courses: 
* notify participants that a course was cancelled
* remind an instructor of an upcoming course
* remind a participant of an upcoming course


geroembser:
@tobiasbork has edited and approved the template for cancelled course mail and created the remaining two templates for the emails that should remind participants and instructors before upcoming courses.

PR #62, opened 2020-06-14
Fix types and screener testuser password

Jonasdoubleyou:
Fixes things I've came across while building the dev-env:

(1) The latest tsc fails to compile without the typecasts, so adding them was the easiest fix
(2) A testuser for screening would be quite useful, thus a proper hashed password has to be set.

### DIFF
--- a/web/controllers/certificateController/index.ts
+++ b/web/controllers/certificateController/index.ts
@@ -54,14 +54,13 @@ export async function certificateHandler(req: Request, res: Response) {
         // todo rename parameter pupil to match
         if (req.params.student != undefined && req.params.pupil != undefined && (res.locals.user instanceof Student || res.locals.user instanceof Pupil)) {
 
-            // TODO: typehint this
             let params = {
-                endDate: req.query.endDate || moment().format("X"),
-                subjects: req.query.subjects || "Mathematik, Biologie, Altgriechisch",
-                hoursPerWeek: Number.parseInt(req.query.hoursPerWeek, 10) || 0,
-                hoursTotal: Number.parseInt(req.query.hoursTotal, 10) || 0,
-                medium: req.query.medium || "Taschenrechner mit Programmierfunktion",
-                categories: req.query.categories || "Liste\nListe2\nListe3"
+                endDate: req.query.endDate as string || moment().format("X") as string,
+                subjects: req.query.subjects as string || "Mathematik, Biologie, Altgriechisch",
+                hoursPerWeek: Number.parseInt(req.query.hoursPerWeek as string, 10) || 0,
+                hoursTotal: Number.parseInt(req.query.hoursTotal as string, 10) || 0,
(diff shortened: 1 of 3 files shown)


Ni42:
Which TS version are you running that fails the type check?

Jonasdoubleyou:
The typescript version hasn't changed as that's fixed in tsconfig.json, I
guess this has only to do with the tsc (typescript@3.9.5). Could also be
that this is just a warning, however `npm run build:clean` fails with a
nonzero exit code (and thus causes the container startup to terminate).

Ni42 <notifications@github.com> schrieb am Mo., 15. Juni 2020, 00:08:

geroembser:
I have no problems running `npm run build:clean using` using tsc version 3.9.5. It will exit with 0 without any problems.

Jonasdoubleyou:
That's weird. Could it be that some of the @types/ dependencies introduced
a breaking change?

geroembser <notifications@github.com> schrieb am Mo., 15. Juni 2020, 00:25:

geroembser:
It seems like you’re using Node 10 while at the same time running `npm install` without the committed `package-lock` file. That combination is the problem, I think.
If I’ll remove the `package-lock` file (which someone should not do, I think, because we wanna have the exact same dependency tree) and run `npm install` with Node 10, I’ll experience the exact same problem. 
But if `npm install` can use the `package-lock` file to restore the dependency tree, I have no problems neither with Node 10 nor with (which I usually use) Node 13. Haven't tested Node 14 yet...

So probably the problem would be fixed without type casts when you also copy the `package-lock` file. Or just use a more recent node version. Or just both – that’s what I do at the moment. Using `package-lock` file is probably always recommendable.

erics465:
Could reproduce with Node 14.4.0 after running `npm update`. That should be avoided. Respecting the `package-lock.json` and just using `npm install` however leads to an error-free build.

Ni42:
Yes, please don't ignore the package-lock.json, it's in the repo for a reason :)

Small notice: npm will update the package-lock if a new version of an installed package appears that still matches package.json. If you ONLY want to install from package-lock.json there is `npm ci`, too.

Jonasdoubleyou:
Yeah, actually checked for package-lock in another repo, but the other teams seem to use yarn so was too lazy to install yarn and further check, but thanks for reproducing that three times :). Will change the docker config, but still I don't see any harm in typing those things properly (cause one is an implicit any, and the other ones will break in the future - as shown).

Ni42:
So, basically you are right. It's a breaking change...
DefinitelyTyped/DefinitelyTyped/pull/43434

Jonasdoubleyou:
Containers fixed with https://github.com/corona-school/dev-env/commit/cb5e25bee9aaa68c8977b0eb7d5f3cef9774de17 (except for the screener backend, that's causing some trouble, and I don't want to mess with package.json as that might break some CD?), could someone review this PR?

PR #63, opened 2020-06-15
Implement instructor mail to all participants, resolve #54

### LINKED ISSUE #54: Instructor mail to all participants
Create a new route, that allows sending a mail via MailJet to all participants of a course.  
We should consider
- Participants should not be able to see each others mail addresses
- The instructors mail should the in the Reply-To field

Ni42:


web/controllers/courseController/index.ts
```diff
@@ -2242,4 +2242,70 @@ async function leaveSubcourse(pupil: Pupil, courseId: number, subcourseId: numbe
...
+
+    const entityManager = getManager();
+    const course = await entityManager.findOne(Course, {id: courseId});
+
+    if (course == undefined) {
+        logger.warn("Tried to send group mail to invalid course");
+        return 404;
+    }
+
+    const subcourse = await entityManager.findOne(Subcourse, {id: subcourseId});
```
BlobbyBob, line 2281: You fail to check if the subcourse belongs to the course. Use
```ts
{id: subcourseId, course: course}
```
as filter

PR #70, opened 2020-06-16
Course Screening API

Jonasdoubleyou:
Adds the following API endpoints:

- /api/screening/courses - to retrieve courses in different states & with search functionality
- /api/screening/course/:id/update - to update courses in various ways, e.g. to update the state or change fields


Jonasdoubleyou:
@geroembser creating the migration would be a lot of work for me, as the backend and postgres are running in containers ... Could you create the migration?

web/index.ts
```diff
@@ -166,6 +166,15 @@ createConnection().then(() => {
             "/screener/:email",
             screeningController.updateScreenerByMailHandler
         );
+        screenerApiRouter.post(
```
geroembser, line 169: Probably `GET` is preferred instead of `POST` ?

web/controllers/screeningController/index.ts
```diff
@@ -301,3 +303,106 @@ export async function updateScreenerByMailHandler(
         next();
     }
 }
+
+/**
+ * @api {POST} /courses getCourses
```
BlobbyBob, line 308: The path should be `/screening/courses` and method should probably be `GET`.

web/controllers/screeningController/index.ts
```diff
@@ -301,3 +303,106 @@ export async function updateScreenerByMailHandler(
...
+ *
+ * @apiName getCourses
+ * @apiGroup Screener
+ *
+ * @apiUse Authentication
+ *
+ * @apiExample {curl} Curl
+ * curl -k -i -X POST -H "Token: <AUTHTOKEN>" https://dashboard.corona-school.de/api/screening/courses
+ *
+ * @apiParam (JSON Body) {string=} courseState the course state ("created", "submitted", "allowed", "denied", "cancelled")
```
BlobbyBob, line 325: What is the `=` for?  
Also: The docs won't compile because of that

web/controllers/screeningController/index.ts
```diff
@@ -301,3 +303,106 @@ export async function updateScreenerByMailHandler(
...
+ * 
+ * Only screeners with a valid token in the request header can use the API.
+ *
+ * @apiName getCourses
+ * @apiGroup Screener
+ *
+ * @apiUse Authentication
+ *
+ * @apiExample {curl} Curl
+ * curl -k -i -X POST -H "Token: <AUTHTOKEN>" https://dashboard.corona-school.de/api/screening/courses
```
BlobbyBob, line 323: The current url is `api.corona-school.de`, not `dashboard`. (The old docs in this file have the same issue.)  
Also method should be GET

web/controllers/screeningController/index.ts
```diff
@@ -301,3 +303,106 @@ export async function updateScreenerByMailHandler(
...
+ *
+ * @apiParam (JSON Body) {string=} courseState the course state ("created", "submitted", "allowed", "denied", "cancelled")
+ * @apiParam (JSON Body) {string=} search A query text to be searched in the title and description
+ */
+export async function getCourses(req: Request, res: Response) {
+    try {
+        const { courseState, search } = req.body;
```
BlobbyBob, line 330: Since this should be a GET request, you should use query parameters instead of a request body.

web/controllers/screeningController/index.ts
```diff
@@ -301,3 +303,106 @@ export async function updateScreenerByMailHandler(
...
+
+        return res.json({ courses });
+    } catch (error) {
+        logger.warn("/screening/courses failed with", error.message);
+        return res.status(500).send("internal server error");
+    }
+}
+
+/**
+ * @api {POST} /course/:courseID/update updateCourse
```
BlobbyBob, line 360: See comments on previous apidoc block.

web/controllers/screeningController/index.ts
Jonasdoubleyou: It's supposed to be an optional parameter (https://jsdoc.app/tags-param.html) should've checked if it works

web/controllers/screeningController/index.ts
Jonasdoubleyou: Then maybe the host should be omitted?

web/controllers/screeningController/index.ts
Jonasdoubleyou: Well actually this was rather by design (https://dropbox.tech/developers/limitations-of-the-get-method-in-http, https://evertpot.com/dropbox-post-api/) ... as the other APIs are also not quite RESTful i thought I could be lazy and use a JSON body instead of doing urlencoding. But I'm fine changing that if you prefer GET.

web/controllers/screeningController/index.ts
Jonasdoubleyou: I just typed it as string|undefined, that should convey the same meaning

PR #82, opened 2020-06-23
Add Feedback Request Mails, Resolve project-user#86

imkgerC:


common/migration/1592927037226-CleanupMails.ts
```diff
@@ -0,0 +1,72 @@
...
+        await queryRunner.query(`ALTER TABLE "match" DROP COLUMN "automaticMailsForFeedbackDisabled"`, undefined);
+        await queryRunner.query(`ALTER TABLE "match" DROP CONSTRAINT "REL_3d73a8a5c5e8a120b63df389b6"`, undefined);
+        await queryRunner.query(`ALTER TABLE "match" DROP COLUMN "confirmationToStudentMailId"`, undefined);
+        await queryRunner.query(`ALTER TABLE "match" DROP CONSTRAINT "REL_fef5c21e7977af1f5fdf2153e0"`, undefined);
+        await queryRunner.query(`ALTER TABLE "match" DROP COLUMN "confirmationToPupilMailId"`, undefined);
+        await queryRunner.query(`ALTER TABLE "match" DROP CONSTRAINT "REL_72eb9ce4ff7e5e1a2d668bb267"`, undefined);
+        await queryRunner.query(`ALTER TABLE "match" DROP COLUMN "dissolutionToStudentMailId"`, undefined);
+        await queryRunner.query(`ALTER TABLE "match" DROP CONSTRAINT "REL_24e84ce75ede0251f5c4f30e73"`, undefined);
+        await queryRunner.query(`ALTER TABLE "match" DROP COLUMN "dissolutionToPupilMailId"`, undefined);
+        await queryRunner.query(`ALTER TABLE "pupil" ALTER COLUMN "verification" SET DEFAULT null`, undefined);
```
Ni42, line 22: This ALTER COLUMN and the following ones which only adjust default values appear because of typeorm/typeorm#3076
You can verify which these are by applying and rerunning migrations. They should be removed from the up() and down() functions to prevent clutter.

common/entity/Match.ts
```diff
@@ -76,44 +75,15 @@ export class Match {
...
         default: false
     })
-    automaticMailsForDissolutionDisabled: boolean;
+    feedbackToPupilMail: boolean;
 
     @Column({
         default: false
     })
-    automaticMailsForFeedbackDisabled: boolean;
+    feedbackToStudentMail: boolean;
```
geroembser, line 86: Those columns are added but do not appear in the migration. I think they'll need to be added there.

Issue #75, opened 2020-06-17
Instructor Screening


BlobbyBob:
Duplicate of Screening

PR #84, opened 2020-06-25
Add APIs for retrieving instructors and adding review to instructors

Jonasdoubleyou:


web/controllers/screeningController/index.ts
```diff
@@ -405,4 +406,125 @@ export async function updateCourse(req: Request, res: Response) {
...
+ * @apiName getInstructors
+ * @apiGroup Screener
+ *
+ * @apiUse Authentication
+ *
+ * @apiExample {curl} Curl
+ * curl -k -i -X GET -H "Token: <AUTHTOKEN>" [host]/api/screening/instructors
+ *
+ * @apiParam (URL Query) {string} screeningStatus get instructors with a certain screeningStatus
+ * @apiParam (URL Query) {string} search fuzzy search inside the instructors name and email
```
geroembser, line 428: Probably just add that the search will respect the Postgres Pattern matching characters. For example underscores are quite common in emails, so someone knows what the API does.

web/controllers/screeningController/index.ts
```diff
@@ -405,4 +406,125 @@ export async function updateCourse(req: Request, res: Response) {
...
+            return res.status(400).send("invalid value for parameter 'search'");
+
+        search = `%${search}%`; // fuzzy search
+
+        let instructors: {}[];
+
+        if (screeningStatus === ScreeningStatus.Accepted) {
+            instructors = await getManager()
+                .createQueryBuilder(Student,"student")
+                .leftJoinAndSelect("student.screening", "screening")
```
geroembser, line 447: https://github.com/corona-school/backend/blob/master/common/entity/InstructorScreening.ts should be used for storing instructor screenings. We would like to decouple the normal tutor screening from the instructor screening. The `isInstructor` and `isStudent` booleans were set during registration and will indicate what a user deserves to be, not what he will allowed to be.

geroembser: So if someone will be allowed to be instructor and student/tutor, he will have two screenings, one `InstructorScreening` and a normal `Screening`. This is how it is modeled in the backend right now.

web/controllers/screeningController/index.ts
```diff
@@ -405,4 +406,125 @@ export async function updateCourse(req: Request, res: Response) {
...
+            return res.status(400).send("Invalid instructor update!");
+
+        const instructor = await getManager().findOne(Student, { where: { id: +id, isInstructor: true } });
+
+        if (!instructor)
+            return res.status(404).send("Instructor not found");
 
+        await instructor.addScreeningResult(screeningResult);
+
+        instructor.isStudent = isStudent;
```
geroembser, line 521: So if someone will be allowed to be instructor and student/tutor, he will have two screenings, one InstructorScreening and a normal Screening. This is how it is modeled in the backend right now.

Jonasdoubleyou: That's not what Christopher told us. Instructors can also be enabled as one on one with one checkbox.

web/controllers/screeningController/index.ts
Jonasdoubleyou: Oh cool, thanks for notifying the team working on this so early! :)

geroembser: Sorry about that, I thought, you had see that we have this. #75 was related to that. It was modeled in the backend before you started working on that instructor screening tool. And I thought it was merged before you started working on the instructor screening feature. At the end, it is just a question of how the others modeled that in the backend. Sorry for that miscommunication. It wasn't initially planned to force tutors/students to have an additional screening for being instructor. That was the reason why this was added in such a strange way and probably also wasn't clearly communicated.

web/controllers/screeningController/index.ts
geroembser: That's true what Christopher said from the user's perspective. For the screener it should just be an additional checkbox. But in the backend, the current modeling will not work this way. Therefore the screening somewhere needs to be duplicated.

PR #91, opened 2020-06-29
Setup mocha test framework

Blackhawk312:
Changes:
- replaced jest with mocha test framework
- rewrote existing jest tests for mocha
- refactored test structure
- added unit and integration tests to CI
- added test coverage tools
- upgrade node module 'pg' to v8.3.0


Blackhawk312:
The main purpose of this change was replacing jest with mocha and adding it to the ci tasks.
Therefore I didn't do a refactoring of the test setup involving the database. Since this code is indeed ugly, I'll add a To-Do item for this.
This refactoring will most likely also include creating a setup for writing integration tests for the REST API.

However, I'd prefer doing this in a separate pull request to ensure that writing additional unit tests is already possible while the refactoring is still ongoing.

PR #92, opened 2020-06-29
implement match follow-up mail, resolve project-user#90

imkgerC:


jobs/periodic/match-follow-up/index.ts
```diff
@@ -0,0 +1,97 @@
...
+    const sevenDaysAgo = moment(now).subtract(7, "days").toDate();
+    return matches.filter(m => {
+        return m.createdAt <= sevenDaysAgo;
+    });
+}
+
+async function sendFollowUpsToStudents(manager: EntityManager, matches: Match[]) {
+    try {
+        for (const m of matches) {
+            sendMatchFollowUpStudent(m.student, m.pupil);
```
geroembser, line 68: I think, we should add an `await` here (or use `.then()` and catch) because otherwise if sending fails, we still save it in the database as sent, resulting in not trying it again next time.

jobs/periodic/match-follow-up/index.ts
```diff
@@ -0,0 +1,97 @@
...
+            return;
+        }
+        throw e;
+    }
+}
+
+async function sendFollowUpsToPupils(manager: EntityManager, matches: Match[]) {
+    try {
+        for (const m of matches) {
+            sendMatchFollowUpPupil(m.student, m.pupil);
```
geroembser, line 85: Same here as in `sendFollowUpsToStudents()`

Issue #96, opened 2020-07-04
Add mentor registration


Ni42:
See corona-school/project-user/issues/53

Add route `POST register/mentor` and implement controller for registration.

PR #101, opened 2020-07-06
Add mentor registration Apidoc

Ni42:


kevinbischof:
What is the status here actually? I wanted to go on with Issue #96 but it would be better to finish this pull request first.

Ni42:
Since this is based on an old version, it might be better to reimplement using new requirements (if any).
If you want to keep anything, the docs were finished based on the old structure, but there are no additional depencies if you changed something in the entity already.

kevinbischof:
Ok. I also thing it's better to reuse the code in a new branch. Therefore I will close this PR.

PR #105, opened 2020-07-07
Add published state to GetCourses subcourse results

Ni42:


BlobbyBob:
What exactly is that for?

Ni42:
Yes, it was supposed to be quick fix, I'm absolutely up for changing this up more. Since we didn't merge it yesterday we should probably take some time to evaluate...

geroembser:
Closing this due to inactivity... 😢

PR #107, opened 2020-07-11
Verification Link for certificates, resolve project-user#74

geroembser:
This will add a verification link to the certificates which can be used to verify the information on the certificate on an official Corona School website. See corona-school/project-user#74.

It basically introduces the following changes:
* The data of every generated certificate will be stored in the database. 
* Add a simple verification HTML page which displays the information from the certificate. It is part of corona-school/backend. 
* A link to the verification page is printed on each generated certificate.


web/dev.ts
```diff
@@ -109,6 +111,22 @@ export async function setupDevDB() {
         console.log("Inserted Dev Match " + i);
     }
 
+    let pc = new ParticipationCertificate();
+    //pc.uuid = "000000001-0000-0000-0701-1b4c4c526384";
+    pc.uuid = randomBytes(5).toString('hex').toUpperCase();
+    pc.pupil = pupils[0];
+    pc.student = students[0];
+    pc.subjects = JSON.stringify(["Englisch", "Deutsch"]);
```
geroembser, line 119: This does not fit to how subjects where stored [here](https://github.com/corona-school/backend/pull/107/files#diff-3abc81a34a2dbec883e655069ec042c0R150) and displayed [there](https://github.com/corona-school/backend/pull/107/files#diff-3abc81a34a2dbec883e655069ec042c0R217). We should either change the dev data or the way how subjects are stored and displayed.

common/entity/ParticipationCertificate.ts
```diff
@@ -0,0 +1,52 @@
...
+    hoursTotal: number;
+
+    @ManyToOne((type) => Student)
+    @JoinColumn()
+    student: Student;
+
+    @ManyToOne((type) => Pupil)
+    @JoinColumn()
+    pupil: Pupil;
+
```
geroembser, line 50: The `medium` is missing here. See this from the certificate generation function:  https://github.com/corona-school/backend/blob/308bbcc8f78dcb973e96be883f183f7b0782e67b/web/controllers/certificateController/index.ts#L189 
It should be also stored in the database when creating a certificate. And of course it should be displayed on the certificate verification page as well.

web/index.ts
```diff
@@ -166,6 +167,23 @@ createConnection().then(() => {
...
+                return next();
+            }
+            certificateController.confirmCertificateHandler(req, res)
+        });
+        participationCertificateRouter.use((req, res, next) =>{
+            if(req.subdomains.includes("verify")){
+                return res.status(404).end()
+            }
+            next();
+        });
```
geroembser, line 183: Catching the `verify.` subdomain this way will work – but this will lead to some errors. For example if a user goes to https://verify.corona-school.de/1DC0F8AA10 then the browser will usually also request https://verify.corona-school.de/favicon.ico (which isn't handled appropriately). This will then log an error every time, which should be avoided.

web/controllers/certificateController/index.ts
```diff
@@ -151,4 +201,30 @@ function createPDFBinary(student: Student, pupil: Pupil, startDate: Date, params
...
-}
\ No newline at end of file
+}
+
+async function viewParticipationCertificate(certificateId) {
+    const entityManager = getManager();
+    let certificate = null;
+    let html = readFileSync("./assets/verifiedCertificatePage.html", "utf8");
+    try {
+        certificate = await entityManager.findOne(ParticipationCertificate, { uuid: certificateId },  { relations: ["student", "pupil"] });
```
geroembser, line 211: Probably we can add some case insensitivity here when querying the database for a certificate ID. This is only additional and not necessary ("nice to have"), but case insensitive names could probably simplify typing the url from a printed certificate paper.

web/dev.ts
fyx99: changed dev data [https://github.com/corona-school/backend/commit/442838bde60b2004c7362ce80a02293802544f9b](https://github.com/corona-school/backend/commit/442838bde60b2004c7362ce80a02293802544f9b)

common/entity/ParticipationCertificate.ts
fyx99: added column [https://github.com/corona-school/backend/commit/442838bde60b2004c7362ce80a02293802544f9b](https://github.com/corona-school/backend/commit/442838bde60b2004c7362ce80a02293802544f9b)

web/controllers/certificateController/index.ts
fyx99: done [https://github.com/corona-school/backend/commit/442838bde60b2004c7362ce80a02293802544f9b](https://github.com/corona-school/backend/commit/442838bde60b2004c7362ce80a02293802544f9b)

web/index.ts
fyx99: Added favicon to assets [https://github.com/corona-school/backend/commit/445c87302c6acb68e8a0ee0aa7db222242263dcb](https://github.com/corona-school/backend/commit/445c87302c6acb68e8a0ee0aa7db222242263dcb) -- any other concerns?

PR #109, opened 2020-07-15
Add redirectTo parameter to registration, and pass it on with verify link

Jonasdoubleyou:
Can't test locally though, let's test in dev.


geroembser:
I think the pull request should go into dev. Here in the backend "dev" is the branch that is deployed to the dev environment.

PR #112, opened 2020-07-16
Feat/bbb api

chrisp1999:


web/controllers/courseController/index.ts
```diff
@@ -2308,4 +2309,175 @@ async function groupMail(student: Student, courseId: number, subcourseId: number
...
+                        meeting = bbbMeetingCache.get(req.params.id);
+                        res.send({
+                            url: meeting.attendeeUrl(`${user.firstname}+${user.lastname}`)
+                        });
+                    } else {
+                        status = 400;
+                        logger.error("BBB-Meeting has not startet yet");
+
+                    }
+                }
```
BlobbyBob, line 2386: Error handling is missing, if the getCourse doesn't return the course (eg. unauthorized instructor, course not found, etc.). You also should not use getCourse to check for validity, since it is a really bloated method for a specific API endpoint.
And you also don't want optional authentication on this route.

web/middleware/bbb.ts
```diff
@@ -0,0 +1,133 @@
+import { hashToken } from "../../common/util/hashing";
```
BlobbyBob, line 1: This shouldn't be in the middleware folder, since it is no express middleware.

web/middleware/bbb.ts
```diff
@@ -0,0 +1,133 @@
...
+import { getLogger } from 'log4js';
+import axios from "axios";
+import { Parser } from "xml2js";
+
+const parser = new Parser();
+const logger = getLogger();
+
+export let bbbMeetingCache: Map<string, BBBMeeting> = new Map<string, BBBMeeting>();
```
BlobbyBob, line 9: As far as I can see, this is not initialized after a server restart.

web/controllers/courseController/index.ts
```diff
@@ -2308,4 +2309,175 @@ async function groupMail(student: Student, courseId: number, subcourseId: number
...
+                let obj = await getCourse(
+                    authenticatedStudent ? res.locals.user : undefined,
+                    authenticatedPupil ? res.locals.user : undefined,
+                    Number.parseInt(req.params.id, 10)
+                );
+                if (typeof obj == 'number') {
+                    status = obj;
+                } else {
+                    course = obj;
+                }
```
BlobbyBob, line 2455: This code is not used at all. And you also don't want optional authentication here.

web/controllers/courseController/index.ts
```diff
@@ -2308,4 +2309,175 @@ async function groupMail(student: Student, courseId: number, subcourseId: number
...
+ * @apiExample {curl} Curl
+ * curl -k -i -X GET -H "Token: <AUTHTOKEN>" https://api.corona-school.de/api/course/<ID>/meeting
+ *
+ * @apiUse StatusOk
+ * @apiUse StatusBadRequest
+ * @apiUse StatusForbidden
+ * @apiUse StatusInternalServerError
+ */
+export async function getCourseMeetingHandler(req: Request, res: Response) {
+    bbbMeetingCache.forEach((value, key) => console.log(value))
```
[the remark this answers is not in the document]
chrisp1999: Was removed.

web/controllers/courseController/index.ts
```diff
@@ -2308,4 +2309,175 @@ async function groupMail(student: Student, courseId: number, subcourseId: number
...
+ * This endpoint provides the BBB-Meeting of a course.
+ *
+ * @apiName GetCourseMeeting
+ * @apiGroup Courses
+ *
+ * @apiUse OptionalAuthentication
+ *
+ * @apiUse Course
+ * @apiUse Subcourse
+ * @apiUse CourseTag
```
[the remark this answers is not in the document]
chrisp1999: Was removed

web/controllers/courseController/index.ts
```diff
@@ -2308,4 +2309,175 @@ async function groupMail(student: Student, courseId: number, subcourseId: number
...
+ * The other participants can only join after the instructor created the meeting with this endpoint
+ *
+ * @apiName JoinCourseMeeting
+ * @apiGroup Courses
+ *
+ * @apiUse OptionalAuthentication
+ *
+ * @apiUse Course
+ * @apiUse Subcourse
+ * @apiUse CourseTag
```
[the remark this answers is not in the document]
chrisp1999: Was removed

web/controllers/courseController/index.ts
chrisp1999: The getCourse function is needed to provide the BBB-Meeting with the correct meeting name (the name of the corresponding course). The code was restructured, so the function will only be called if the meeting has not been created yet. Error handling should work now.

web/middleware/bbb.ts
chrisp1999: Moved it to the utils folder for now.

web/controllers/courseController/index.ts
chrisp1999: Removed unnecessary code and optional authentication.

PR #114, opened 2020-07-22
Add initial migration

Ni42:


geroembser:
Was implemented with 9f0b958937647f6a5ba79b8a5f0ad3ff62374565 instead...

PR #125, opened 2020-09-02
Fixes apidoc and using of subcourse param, changes endpoint from post…

kevinbischof:
… to get in courses router, improves calculation of pupil's attendance time


web/controllers/courseController/index.ts
```diff
@@ -2334,34 +2334,34 @@ async function groupMail(student: Student, courseId: number, subcourseId: number
...
+ * @apiParam (URL Parameter) {int} subid ID of the subcourse
+ *
+ * @apiName GetCourseMeeting
  * @apiGroup Courses
  *
  * @apiUse Authentication
- * @apiUse Course
- *
- * @apiParam (JSON Body) {int} subcourseId ID of the subcourse of the course that should be joined
+ * @apiUse BBBMeeting
```
geroembser, line 2351: I don't think that we will need the BBBMeeting in the API doc, because as far as I remember the BBBMeeting is only used internally and never exposed to the API?

common/util/bbb.ts
```diff
@@ -184,29 +164,21 @@ function mapJSONtoAttendee(o: any): Attendee {
...
+    const now = new Date();
     // check if lecture is running now (lecture.start + duration > now)
     for (const lecture of lectures) {
-        if (!lessThanDate(lecture.start, new Date())
-            && (lecture.start.getTime() + (lecture.duration * 60000)) > new Date().getTime()) {
+
+        if (lecture.start.getFullYear() == now.getFullYear() &&
+            lecture.start.getMonth() == now.getMonth() &&
+            lecture.start.getDate() == now .getDate() &&
+            (lecture.start.getTime() + (lecture.duration * 60000)) > new Date().getTime()) {
```
geroembser, line 181: Looks good, but we could also use `now` here instead of `new Date()`, as in the lines above.

web/controllers/courseController/index.ts
kevinbischof: Yeah thats true.

common/util/bbb.ts
kevinbischof: You're right.

web/controllers/courseController/index.ts
```diff
@@ -2334,34 +2334,32 @@ async function groupMail(student: Student, courseId: number, subcourseId: number
 }
 
 /**
- * @api {POST} /course/:id/meeting/join JoinCourseMeeting
+ * @api {GET} /course/:id/subcourse/:subid/meeting GetCourseMeeting
```
geroembser, line 2337: I think that is not consistent with that name https://github.com/corona-school/backend/blob/01abe7dc56d0d5d6b08785ed7018d5c33234bb8c/web/controllers/courseController/index.ts#L2360 and how it's accessible via the API, see this https://github.com/corona-school/backend/blob/01abe7dc56d0d5d6b08785ed7018d5c33234bb8c/web/index.ts#L139

web/controllers/courseController/index.ts
```diff
@@ -2334,34 +2334,32 @@ async function groupMail(student: Student, courseId: number, subcourseId: number
...
- * @apiUse Course
- *
- * @apiParam (JSON Body) {int} subcourseId ID of the subcourse of the course that should be joined
  *
  * @apiExample {curl} Curl
- * curl -k -i -X POST -H "Token: <AUTHTOKEN>" https://api.corona-school.de/api/course/<ID>/meeting/join
+ * curl -k -i -X GET -H "Token: <AUTHTOKEN>" https://api.corona-school.de/api/course/<ID>/subcourse/<ID>/meeting
  *
  * @apiUse StatusOk
  * @apiUse StatusBadRequest
```
geroembser, line 2356: I think what the API returns here is again missing. I think that was accidentally removed https://github.com/corona-school/backend/pull/125/commits/01abe7dc56d0d5d6b08785ed7018d5c33234bb8c#diff-41dab649f245dd7fb8b0ea1406ee7122L361-L366

geroembser (changes requested):

geroembser (approved):

PR #136, opened 2020-09-28
[#173] Administrator edit course date and time

lsc0498:
This is the backend part necessary for editing course date and times during course update as a screener. This is realized by adding or removing lectures respectively.

### DIFF
--- a/web/controllers/screeningController/index.ts
+++ b/web/controllers/screeningController/index.ts
@@ -347,7 +347,7 @@ export async function updateCourse(req: Request, res: Response) {
         }
 
         if (newLectures !== undefined) {
-            if (Array.isArray(newLectures) && newLectures.every(l => (Number.isInteger(l.subcourse.id) && !!Date.parse(l.start) && Number.isInteger(l.duration)))) {
+            if (Array.isArray(newLectures) && newLectures.every(l => (Number.isInteger(l.subcourse.id) && !!Date.parse(l.start) && Number.isInteger(l.duration) && Number.isInteger(l.instructor.id)))) {
                 const status = await handleNewLectures(newLectures, +id);
                 if (status != 200) {
                     return res.status(status).send("Adding lectures failed.");
@@ -372,7 +372,7 @@ export async function updateCourse(req: Request, res: Response) {
     }
 }
 
-async function handleNewLectures(lectures: { subcourse: { id: number }, start: Date, duration: number }[], courseId: number) {
+async function handleNewLectures(lectures: { subcourse: { id: number }, start: Date, duration: number, instructor: { id: number } }[], courseId: number) {
     const entityManager = getManager();
 
     for (let i=0; i<lectures.length; i++){
@@ -386,11 +386,21 @@ async function handleNewLectures(lectures: { subcourse: { id: number }, start: D
             logger.warn(`No subcourse found with ID ${lectures[i].subcourse.id}`);
             return 404;
         }
+        const instructor = await entityManager.findOne(Student, { id: lectures[i].instructor.id });
+        if (instructor == undefined) {
+            logger.warn(`No instructor with ID ${lectures[i].instructor.id}`);
+            return 404;
+        }
+        if (subcourse.instructors.findIndex(i => i.id === instructor.id) == -1){
+            logger.warn(`Instructor with ID ${lectures[i].instructor.id} is not in subcourse ${subcourse.id}`);
+            return 403;
+        }
 
         const lecture = new Lecture();
         lecture.subcourse = subcourse;
         lecture.start = lectures[i].start;
         lecture.duration = lectures[i].duration;
+        lecture.instructor = instructor;
 
         try {
             await entityManager.save(Lecture, lecture);
--- a/web/controllers/screeningController/index.ts
+++ b/web/controllers/screeningController/index.ts
@@ -324,6 +324,8 @@ export async function getCourses(req: Request, res: Response) {
  * @apiParam (JSON Body) {string|null|undefined} imageUrl the new image url, or null if no image should be set
  * @apiParam (JSON Body) {Object[]|undefined} instructors the instructor ids of this course
  * @apiParam (JSON Body) {number|undefined} instructors.id the instructor ids of this course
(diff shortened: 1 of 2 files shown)


web/controllers/screeningController/index.ts
```diff
@@ -347,6 +374,67 @@ export async function updateCourse(req: Request, res: Response) {
...
+        if (subcourse == undefined) {
+            logger.warn(`No subcourse found with ID ${lectures[i].subcourse.id}`);
+            return 404;
+        }
+        const instructor = await entityManager.findOne(Student, { id: lectures[i].instructor.id });
+        if (instructor == undefined) {
+            logger.warn(`No instructor with ID ${lectures[i].instructor.id}`);
+            return 404;
+        }
+        if (subcourse.instructors.findIndex(i => i.id === instructor.id) == -1){
```
Jonasdoubleyou, line 396: `!...some(it => it.id === instructor.id)` would be more elegant

web/controllers/screeningController/index.ts
```diff
@@ -347,6 +374,67 @@ export async function updateCourse(req: Request, res: Response) {
...
+async function handleNewLectures(lectures: { subcourse: { id: number }, start: Date, duration: number, instructor: { id: number } }[], courseId: number) {
+    const entityManager = getManager();
+
+    for (let i=0; i<lectures.length; i++){
+        const course = await entityManager.findOne(Course, { id: courseId });
+        if (course == undefined) {
+            logger.warn(`No course found with ID ${courseId}`);
+            return 404;
+        }
+        const subcourse = await entityManager.findOne(Subcourse, { id: lectures[i].subcourse.id, course: course });
```
Jonasdoubleyou, line 386: This could possible be moved to the Course Model, like `const subcourse = (await course.getSubcourses())[0]`, to make it reusable

web/controllers/screeningController/index.ts
```diff
@@ -347,6 +374,67 @@ export async function updateCourse(req: Request, res: Response) {
...
+        if (instructor == undefined) {
+            logger.warn(`No instructor with ID ${lectures[i].instructor.id}`);
+            return 404;
+        }
+        if (subcourse.instructors.findIndex(i => i.id === instructor.id) == -1){
+            logger.warn(`Instructor with ID ${lectures[i].instructor.id} is not in subcourse ${subcourse.id}`);
+            return 403;
+        }
+
+        const lecture = new Lecture();
```
Jonasdoubleyou, line 401: Maybe we can add a constructor to Lecture, like `new Lecture(lecture, subcourse)` or `subcourse.addLecture(lecture)`?

web/controllers/screeningController/index.ts
```diff
@@ -347,6 +374,67 @@ export async function updateCourse(req: Request, res: Response) {
...
+        if (course == undefined) {
+            logger.warn(`No course found with ID ${courseId}`);
+            return 404;
+        }
+        const subcourse = await entityManager.findOne(Subcourse, { id: lectures[i].subcourse.id, course: course });
+        if (subcourse == undefined) {
+            logger.warn(`No subcourse found with ID ${lectures[i].subcourse.id}`);
+            return 404;
+        }
+        const instructor = await entityManager.findOne(Student, { id: lectures[i].instructor.id });
```
Jonasdoubleyou, line 391: same here, `await subcourse.getInstructor()` could be better, especially if the db model in the future.

web/controllers/screeningController/index.ts
lsc0498: I added a method addLecture to Subcourse which seemed more intuitive to me.

web/controllers/screeningController/index.ts
lsc0498: There is exactly one instructor specified for each lecture while there can be many for a subcourse. Hence extracting the instructor from the subcourse wouldn't make sense here. Nevertheless, in the lecture constructor I added to the subcourse I take the instructor from the subcourse matching the given instructor id.

web/controllers/screeningController/index.ts
lsc0498: The data modelling theoretically allows more than one subcourse with lectures though this possibility is not used at the moment. So finding the correct subcourse should always take into account the lecture here. 
However adding a function getSubcourse taking a lecture id to the course model would be valid, nevertheless this use case is in my opinion too fine-grained and specific for moving that function to the course model. I would leave it here.

web/controllers/screeningController/index.ts
```diff
@@ -347,6 +374,53 @@ export async function updateCourse(req: Request, res: Response) {
...
+        }
+        const subcourse = await entityManager.findOne(Subcourse, { id: lecture.subcourse.id, course: course });
+        if (subcourse == undefined) {
+            logger.warn(`No subcourse found with ID ${lecture.subcourse.id}`);
+            return 404;
+        }
+
+        try {
+            await subcourse.addLecture(lecture);
+            await entityManager.save(subcourse);
```
Jonasdoubleyou, line 394: what has changed in the subcourse that we need to save it?

web/controllers/screeningController/index.ts
Jonasdoubleyou: Ah, right, `Subcourse.forLecture(lecture)` would then be some way to move it to the Model. However I'm fine leaving it as is.

web/controllers/screeningController/index.ts
Jonasdoubleyou: Yeah, your new version looks reasonable.

web/controllers/screeningController/index.ts
lsc0498: We added the lecture to the subcourse and save that change. This could be dropped in favour of specifying the lectures subcourse on construction. But then the lecture's constructor would be somewhat independent of the subcourse model and could be located elsewhere. In the spirit of "adding a lecture to a subcourse" I decided to append the lectures array in the subcourse by the new lecture. But this is certainly up to taste 🙂

geroembser: Sorry for the delay! This looks good to me as well!

geroembser (approved):

PR #137, opened 2020-09-29
bbb cache replacement fixes #129 and fixes 132

### LINKED ISSUE #129: Add Readme on how to create a database migration
Our documentation lacks some explanations on how to create database migrations for our code. Since this is a quite common task, we should somewhere describe how to best generate (and even test locally) database migrations if someone modifies the entities. This should make it even simpler for new team members to generate and commit migrations as they're working on our code. 

Thanks to @kevinbischof for bringing this to my attention.

### LINKED ISSUE #132: Problems with BBB-Meetings
See description in commit https://github.com/corona-school/backend/commit/68e23397a5f6d0dd3d63968b3f9ecf115898a2b0

Unfortunately there is currently no exact description or documentation of the problem. I suspect that there is an overlap of two actions:
1. the automatic termination of BBB meetings by BBB even after the last user has left a meeting. The time limit is 15 minutes. 
2. The backend fetches all active meetings from BBB every 14 minutes and keeps them in cache.

kevinbischof:


kevinbischof:
I don't know why lint fails here. Can't solve this problem.

Ni42:
In backend/common/util/bbb.ts, from line 117-123 and 170-171 the indentation needs to be adjusted to match the indent of the opening bracket of the constructor (it used to be right before the changes).

common/util/bbb.ts
```diff
@@ -220,26 +215,17 @@ export async function createOrUpdateCourseAttendanceLog(pupil: Pupil, ip: string
...
         if (activeLecture) {
             const logToUpdate = await getCourseAttendanceLog(activeLecture.id, pupil.id);
-            // Update log
             if (logToUpdate) {
-                // To prevent a high attendance time through rejoining check the absence time
-                // If absence time (difference between now and updatedAt) is shorter than interval time, add abscence time to attended time
-                // Else add courseAttendanceLogInterval to attendedTime
-                const absenceTime = now.getTime() - logToUpdate.updatedAt.getTime();
-                if (absenceTime < courseAttendanceLogInterval) {
-                    logToUpdate.attendedTime += absenceTime;
```
geroembser, line 238: Why is this check removed?

geroembser (dismissed):
This looks good! Today I had time to test this in more detail (so that we won't have any new problems when deploying this so that our course platform will be stable at all times). So far, no problems have occurred in my test environment. Good job!
But there is this one question (which does not really affect the functionality) but rather a question I asked myself during the review of the changes before approval...

common/util/bbb.ts
kevinbischof: Thank you! 
You're totally right. I can't remember why I deleted this part. I re-implemented this part and tested it. Should work like before, but it would be good if you could test it too, so there will be no surprise in the future :)

geroembser (approved):
I've checked this again. Looks good so far!

PR #146, opened 2020-11-08
Fix/screening issues

lsc0498:


web/controllers/screeningController/index.ts
```diff
@@ -260,14 +260,16 @@ export async function updateStudentByMailHandler(req: Request, res: Response, ne
 
 export async function updateStudentInformation(student: Student, info: StudentEditableInfoDTO, screener: Screener) {
     //update static fields on student
+    student.email = info.email.toLowerCase();
```
geroembser, line 263: Just as a note: if the email address already exists, the `save` call later will fail (due to uniqueness constraints). Thus the whole API request will fail, without proper indicating an error. So we should probably fix this in the future

geroembser (approved):

web/controllers/screeningController/index.ts
lsc0498: See #148

Issue #149, opened 2020-11-22
Project expert list


lsc0498:
As a project coach I want to be able to publish my profile to students including some information on my expertise.
As a project coachee I want to be able to search for experts and contact them. 

- [x] Add a project expert entity linked to student entity. Must contain contact-email, description, tags, active boolean, allowed boolean

**API Endpoints**

- [x] Contact expert

- [x] get experts

- [x] put Expert

- [x] get User (add expert information to existing endpoint)

PR #150, opened 2020-11-23
Project expert list

lsc0498:
This implements #149.


web/controllers/expertController/index.ts
```diff
@@ -0,0 +1,328 @@
...
+        status = 500;
+    }
+    res.status(status).end();
+}
+
+async function postContactExpert(id: string, user: Pupil | Student, apiContactExpert: ApiContactExpert): Promise<number> {
+    const entityManager = getManager();
+    const transactionLog = getTransactionLog();
+
+    const expert = await entityManager.findOne(ExpertData, { id: Number(id) });
```
geroembser, line 75: We cannot be sure that `id` is a string that really contains a number. If not, this will result in `NaN` and a failure.

web/controllers/expertController/index.ts
```diff
@@ -0,0 +1,328 @@
...
+
+    const expertiseTags: ExpertiseTag[] = await GetExpertiseTagEntities(info.expertiseTags);
+
+    let expertData = await entityManager.findOne(ExpertData, { student: student });
+    if (!expertData) {
+        expertData = new ExpertData();
+        expertData.student = student;
+    }
+
+    logger.debug(expertData);
```
geroembser, line 241: What's the purpose of this logging statement?

web/dev.ts
```diff
@@ -817,6 +820,47 @@ export async function setupDevDB() {
...
+    const expert1 = new ExpertData();
+    expert1.student = students[5];
+    expert1.contactEmail = "contact@jufo-tufo.de";
+    expert1.description = "JuFo is great!";
+
+    experts.push(expert1);
+
+    const expert2 = new ExpertData();
+    expert2.student = students[6];
+    expert2.contactEmail = "contact@jufo-tufo.de";
```
geroembser, line 853: We can discuss, whether we allow the same contact email address for several experts. But I think it might have its use cases.

geroembser (commented):
Good job. I really like it. But before we're ready for merging, a database migration is still missing.

common/entity/ExpertData.ts
```diff
@@ -49,7 +50,7 @@ export class ExpertData {
     active: boolean;
 
     @Column({
-        default: false
+        default: ExpertAllowedIndication.PENDING
     })
-    allowed: boolean;
+    allowed: ExpertAllowedIndication;
```
geroembser: In the past I have forgotten this a few times too, so I want to mention it here because here I just noticed it: Enum defines without explicitly specifying the enum type were saved as a string in the database, and not as a Postgres Enum, which would be preferred, I guess. Also see documentation [here](https://github.com/typeorm/typeorm/blob/master/docs/entities.md#enum-column-type).

PR #151, opened 2020-12-01
API for user certificates & Cleanup

Jonasdoubleyou:
As we start working on the frontend tomorrow, I'd like to deploy this to dev today. 

The added API /api/certificates exposes all certificates the user is involved in (either as a student or pupil).

### DIFF
--- a/assets/certificateTemplate.html
+++ b/assets/certificateTemplate.html
@@ -14,7 +14,7 @@
     Text Schüler: %SCHUELERFREITEXT%<br/>
     Stunden pro Woche: %SCHUELERPROWOCHE%<br/>
     Stunden gesamt: %SCHUELERGESAMT%<br/>
-    Mium: %MEDIUM%<br/>
+    Medium: %MEDIUM%<br/>
     Screnningdatum: %SCREENINGDATUM%<br/>
 </body>
 </html>
--- a/assets/certificateTemplate.html
+++ b/assets/certificateTemplate.html
@@ -1,0 +1,20 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>[TEST] Corona School Bescheinigung</title>
+</head>
+<body>
+    Name des Studenten: %NAMESTUDENT%<br/>
+    Name des Schülers: %NAMESCHUELER%<br/>
+    Heutiges Datum: %DATUMHEUTE%<br/>
+    Startdatum: %SCHUELERSTART%<br/>
+    Enddatum: %SCHUELERENDE%<br/>
+    Fächer: %SCHUELERFAECHER%<br/>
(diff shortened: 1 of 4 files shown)


web/controllers/certificateController/index.ts
```diff
@@ -118,92 +135,159 @@ export async function certificateHandler(req: Request, res: Response) {
...
+        const { certificateId } = req.params;
+        const entityManager = getManager();
+
+        if (typeof certificateId !== "string")
+            return res.status(400).send("Missing parameter certificateId");
+
+        const certificate = await entityManager.findOne(ParticipationCertificate, { uuid: certificateId.toUpperCase() }, { relations: ["student", "pupil"] });
+
+        if (!certificate)
+            return res.status(404).send("<h1>Zertifikatslink nicht valide.</h1>");
```
lsc0498, line 148: It is discussable whether an invalid certificate should lead to a 404 here. Since the use case implies somehow the possible existence of invalid certificates, it is thus no technical error.

lsc0498 (approved):
Looks good to me 👍

web/controllers/certificateController/index.ts
geroembser: I think 404 is fine here... I think it doesn't matter from an enduser's perspective who just opens the link through her browser.

assets/certificateTemplate.html
```diff
@@ -0,0 +1,20 @@
...
+    Name des Studenten: %NAMESTUDENT%<br/>
+    Name des Schülers: %NAMESCHUELER%<br/>
+    Heutiges Datum: %DATUMHEUTE%<br/>
+    Startdatum: %SCHUELERSTART%<br/>
+    Enddatum: %SCHUELERENDE%<br/>
+    Fächer: %SCHUELERFAECHER%<br/>
+    Text Schüler: %SCHUELERFREITEXT%<br/>
+    Stunden pro Woche: %SCHUELERPROWOCHE%<br/>
+    Stunden gesamt: %SCHUELERGESAMT%<br/>
+    Mium: %MEDIUM%<br/>
```
geroembser, line 17: Probably "Medium", but doesn't really matter 😂

assets/certificateTemplate.html
```diff
@@ -0,0 +1,20 @@
...
+    Screnningdatum: %SCREENINGDATUM%<br/>
```
geroembser, line 18: I guess `CERTLINK` and `CERTLINKTEXT` would be more appropriate than the unused `SCREENINGDATUM` here, so someone can also see this information during development or while testing this locally.

assets/certificateTemplate.html
Jonasdoubleyou: To not accidentially commit the live version I used `git update-index --assume-unchanged` ... Actually fixed that typo, just didn't appear in the diff ... done now :)

geroembser (approved):

PR #154, opened 2020-12-04
Add phone to registration

MS-SAP:


lsc0498 (commented):
Very good work!🎉

web/controllers/registrationController/index.ts
```diff
@@ -52,6 +52,7 @@ export async function postTutorHandler(req: Request, res: Response) {
         if (typeof req.body.firstname == 'string' &&
             typeof req.body.lastname == 'string' &&
             typeof req.body.email == 'string' &&
+            typeof req.body.phone == 'string' || typeof req.body.phone == 'undefined' &&
```
Jonasdoubleyou, line 55: this won't work due to operator predescendence. It reads as (firstname && lastname && email && phone) || (phone && ..). You might want to put this or in braces.

web/controllers/registrationController/index.ts
```diff
@@ -323,6 +335,7 @@ export async function postTuteeHandler(req: Request, res: Response) {
...
```
Jonasdoubleyou, line 338: same here

web/controllers/registrationController/index.ts
```diff
@@ -604,6 +628,7 @@ export async function postMentorHandler(req: Request, res: Response) {
...
```
Jonasdoubleyou, line 631: same here

web/controllers/registrationController/index.ts
```diff
@@ -753,6 +788,7 @@ export async function postStateTuteeHandler(req: Request, res: Response) {
...
```
Jonasdoubleyou, line 791: same here

Jonasdoubleyou (changes requested):

web/controllers/registrationController/index.ts
MS-SAP: Thanks for alerting. Fixed now.

Jonasdoubleyou (approved):

PR #156, opened 2020-12-09
Certificates: List all certificates in settings, support for multiple languages

Jonasdoubleyou:
Migrate certificate generation to EJS, add an "ongoingLessons" property to certificates, 
 rename the current "get certificate" to "create certificate" and add a new API endpoint which actually gets the certificate. 
Add certificates in multiple languages. 

TODO:

- [x] Create Certificate lang query doesn't work
- [x] Lessons are for some reason formatted differently (possibly on frontend), investigate. 
- [x] Update from dev
- [x] Create templates with EJS & in English

TO VERIFY:

- [ ] Create Certificate
- [ ] Create Certificate with "ongoingLessons"
- [ ] View Certificate in English
- [ ] Verify Certificate (in English)
- [ ] View Settings as Student AND User

WHEN DEPLOYING TO PROD:

- [ ] Create Database Migration and run it to add "ongoingLessons" column
- [ ] At the same time, deploy the [frontend](https://github.com/corona-school/web-user-app/pull/104) as the paths changed
- [ ] Verify that old certificate link still works (without ?lang=)

resolves https://github.com/corona-school/project-user/issues/115
resolves https://github.com/corona-school/project-user/issues/169
partly solves https://github.com/corona-school/project-user/issues/139
resolves https://github.com/corona-school/project-user/issues/156
resolves https://github.com/corona-school/project-user/issues/155


common/entity/ParticipationCertificate.ts
```diff
@@ -56,5 +56,6 @@ export class ParticipationCertificate {
     @JoinColumn()
     pupil: Pupil;
 
-
+    @Column({ default: false })
+    ongoingLessons: boolean;
```
lsc0498, line 60: We will probably need a migration before deployment @geroembser?

lsc0498 (commented):
Looks great so far. Thank you very much! 🎉

common/entity/ParticipationCertificate.ts
geroembser: Yes, we should do that. It would be great if that would be part of this PR as well. But as we've done in the past, I'd also be happy to do the migration for you again if there's possibly all the overhead with the containers. So just let me know, it'll be done quickly with my setup if everything else is ok.

Jonasdoubleyou: @geroembser yes, that would be nice, the device on which I could possibly do that myself is currently 600km away ...

geroembser (approved):

PR #160, opened 2020-12-19
Expose certificate verification endpoint for testing

Jonasdoubleyou:
It is supposed to be publicly reachable, so authentication should not be required


Jonasdoubleyou:
not needed atm, can be done with regular next deployment

PR #167, opened 2021-01-08
Improve instructor search by firstname

Jonasdoubleyou:
As far as I can see, the instructor review page for which this search was originally built was replaced with something else.
This API endpoint is now only used for the "Manage Instructors" dialogue.
In that dialogue, it seems as if one can search instructors by full name, however that is not quite true, as the current implementation searches by lastname only.
So I thought to finish up the "search in review" topic, I'll align the behavior of this search with the one in the courses search, allowing users to search by fullname.

### DIFF
--- a/web/controllers/screeningController/index.ts
+++ b/web/controllers/screeningController/index.ts
@@ -777,7 +777,7 @@ export async function getInstructors(req: Request, res: Response) {
         instructors = await getManager()
             .createQueryBuilder(Student, "student")
             .leftJoinAndSelect("student.instructorScreening", "instructor_screening")
-            .where(`student.isInstructor = true AND ${condition} AND (student.email ILIKE :search OR (student.firstname ILIKE :firstname AND student.lastname ILIKE :lastname))`, { email, firstname, lastname })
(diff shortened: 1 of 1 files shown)


geroembser (approved):
Ok, thanks again for the implementation and also for thinking along with this search. 
And sorry for the delay in reviewing this. But to give you some context, you may be wondering why these endpoints are temporarily commented out (and not used in production). This is primarily due to some performance issues we had last week, especially with the `getCourseTags` endpoint. Too much data in production, bad frameworks that we're not really using efficiently, and lots of concurrent users accessing our public course page🥳. And since we also expect a lot of users on our course platform for the duration of the next 2 weeks, many of whom are relying on it to get them up and running quickly, we've decided to temporarily lessen the impact on our users by disabling this feature since almost all courses and their instructors have already completed the review process for this period. Maybe at some point we'll find the time to fix this better. So this won't be active again for another two weeks or so. 
But anyway: So thank you for your good work!

Jonasdoubleyou:
@geroembser yeah, joining all coursetags with all their courses doesn't seem to be a good idea (performance wise) ... and I wonder wether that information is actually used in the frontend ... (and thanks for the context :))

PR #171, opened 2021-01-27
Backend support for automatic signature signing process

Jonasdoubleyou:
The new "automatic" flow is opt-in, so currently no end user will be able to see the new flow, until enabled in the frontend. 

A database migration will be required.

The next steps are then:
- Add the new certificate template and verify that the old and new versions work
- Add the email templates to the backend 
- Finalize the frontend, this is currently close to being finished (see feat/signature)
- Enable the automatic flow in the frontend


web/controllers/certificateController/index.ts
```diff
@@ -3,7 +3,7 @@ import { Request, Response } from 'express';
 import { Pupil } from '../../../common/entity/Pupil';
 import { Student } from '../../../common/entity/Student';
 import { getTransactionLog } from '../../../common/transactionlog';
-import { getManager } from 'typeorm';
+import { getManager, getRepository } from 'typeorm';
```
geroembser, line 6: Looks like this import is unused 🤭 (But it's not the only one in this file and in the entire code, so not really important)

web/index.ts
```diff
@@ -127,7 +127,7 @@ createConnection().then(() => {
         certificateRouter.get("/create/:student/:pupil", authCheckFactory(), certificateController.createCertificateEndpoint);
         certificateRouter.get("/:certificateId", authCheckFactory(), certificateController.getCertificateEndpoint);
         certificateRouter.get("/:certificateId/confirmation", /* NO AUTH REQUIRED */ certificateController.getCertificateConfirmationEndpoint);
-
+        certificateRouter.post("/:certificateId/sign", authCheckFactory(), certificateController.signCertificateEndpoint);
```
geroembser, line 130: So with this implementation a pupil (or her parents) need to be authenticated to sign a certificate. Following corona-school/project-user#150 this would mean we need some other kind of authentication included in the link that ensures the user is not forced to login again. This would be ok too, but I could image that making this endpoint somehow publicly available or adding another authentication mechanism here might be simpler. But we can also go with including the _current_ authtoken in the email for now.

common/entity/ParticipationCertificate.ts
```diff
@@ -58,4 +58,19 @@ export class ParticipationCertificate {
 
     @Column({ default: false })
     ongoingLessons: boolean;
+
+    @Column({ default: "manual" })
+    state: "manual" | "awaiting-approval" | "approved";
+
+    @Column({ nullable: true, type: "bytea", select: false })
```
geroembser, line 65: If we’re storing the image data as base64 (this is how I understand the rest of the code), why do we not just use a simple `text` column in the database instead? Is `bytea` performing better than _text_?

web/dev.ts
```diff
@@ -15,7 +15,7 @@ import { InstructorScreening } from "../common/entity/InstructorScreening";
 import { CourseAttendanceLog } from "../common/entity/CourseAttendanceLog";
 import { Division, Expertise, Mentor } from "../common/entity/Mentor";
 import { School } from "../common/entity/School";
-import { State } from "../common/entity/State";
+import { State, stateCooperationSubdomainPart } from "../common/entity/State";
```
geroembser, line 18: [Just another unused import, probably not the only one, but one I've noticed 😂]

geroembser (commented):
And should I create the corresponding database migration for you again?

common/entity/ParticipationCertificate.ts
Jonasdoubleyou: No, I think performance wise they are more or less equal. That was based on "In short, binary strings are appropriate for storing data that the programmer thinks of as "raw bytes", whereas character strings are appropriate for storing text." in the Postgres Docs. But I guess TEXT is also fine.

web/dev.ts
Jonasdoubleyou: Huh, did I add that? Sometimes VS Code seems to import stuff I don't even use, weird ...

web/index.ts
Jonasdoubleyou: Ah, I actually missed that requirement. I guess we have similar requirements at other places, so including the current auth token would be a "general solution". The /login frontend already takes the fitting query parameters, right?

web/controllers/certificateController/index.ts
```diff
@@ -162,7 +167,12 @@ export async function getCertificateEndpoint(req: Request, res: Response) {
         if (typeof certificateId !== "string")
             return res.status(400).send("Missing parameter certificateId");
 
-        const certificate = await entityManager.findOne(ParticipationCertificate, { student: requestor, uuid: certificateId.toUpperCase() }, { relations: ["student", "pupil"] });
+        /* Retrieve the certificate and also get the signature columsn that are usually hidden for performance reasons */
+        const certificate = await entityManager.findOne(ParticipationCertificate, { uuid: certificateId.toUpperCase(), student: requestor }, {
+            relations: ["student", "pupil"],
+            /* Unfortunately there is no "*" option which would also select the signatures. The query builder also does not cover this case */
+            select: ["uuid", "categories", "certificateDate", "endDate", "hoursPerWeek", "hoursTotal", "id", "medium", "ongoingLessons", "signatureParent", "signaturePupil", "startDate", "state", "subjects", "uuid"]
```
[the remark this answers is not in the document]
Jonasdoubleyou: Good catch, somehow forgot to add these new fields to the test template and thus did not notice ...

web/index.ts
Jonasdoubleyou: The certificate (in german) now gets attached to the Mail itself, also the link attached contains the login token, and then opens the settings page (in theory, cannot test the mails :)) so that other languages can be downloaded.

web/controllers/certificateController/index.ts
```diff
@@ -494,18 +492,17 @@ function createPDFBinary(certificate: ParticipationCertificate, link: string, la
...
-            if (err) {
-                reject(err);
-            } else {
-                resolve(buffer);
-            }
-        });
+    const ASSETS = __dirname + "/../../../../assets";
+    return await generatePDFFromHTMLString(result, {
+        includePaths: [
+            path.resolve(ASSETS, "creiners.png"),
```
geroembser, line 502: As an alternative to manually including all those files (and thus adding the empty test files to the assets in this repository) we could also think about simply including the entire assets directory and adding `<base href="./assets/">` to the certificates templates. Having those empty files will also complicate the current deployment process because this is basically a `git pull` which then, on every deployment, overwrites the actual assets files with those empty files.

web/controllers/certificateController/index.ts
```diff
@@ -87,6 +89,8 @@ export async function createCertificateEndpoint(req: Request, res: Response) {
         if (requestor.wix_id != req.params.student)
             return res.status(403).send("Students may only retrieve certificates for themselves");
 
+        let state = req.params.automatic ? State.awaitingApproval : State.manual;
```
geroembser, line 92: So having this information in the URL parameters, instead of including it in the query parameters, means that we also need to change the definition of the endpoint. Or should that be included later and disabled for now?

geroembser (commented):
Everything seems to work. The emails also look good. I'm only unsure if automatic process should be activated (see the comment). Otherwise only the database migration is missing before we could merge.

web/controllers/certificateController/index.ts
Jonasdoubleyou: Ah, right, not sure why I chose `req.params` here ... `req.query` makes much more sense

web/controllers/certificateController/index.ts
Jonasdoubleyou: @geroembser So using `path.resolve(ASSETS)` copies all the files in an `assets/` subfolder? Then that makes a lot of sense

Jonasdoubleyou:
@geroembser yes, it would be very nice if you could add the migration :) My local setup is even more broken now (due to Pupeteer et. al.)

geroembser (approved):
👍 Thank you!

PR #172, opened 2021-02-09
Add Secret Assets during Heroku Pipeline

Jonasdoubleyou:
To be able to run the production backend on Heroku, we need to add the assets from their secret repository.
This will now be done as part of a Shell script running as part of the build stage.

A custom Buildpack does not seem necessary, as Heroku's default stack is based on Ubuntu and already has all the tools we need (git).


setup.sh
```diff
@@ -0,0 +1,44 @@
...
+  echo '> Cloning the secret asset repo into it'
+  ssh-agent bash -c 'ssh-add - <<< "${SECRET_ASSETS_KEY}"; git clone "${SECRET_ASSETS_REPOSITORY}" .'
+
+  echo '> These files are now there:'
+  ls -R
+else
+  echo '> No Asset repository set in ENV, skipping step'
+fi
+
+echo ' -------------- Setup.sh ended successfully -------------- '
```
geroembser, line 44: I think it might be useful to return the result/exit code of `git clone` as the script's exit value to fail the script in case cloning fails, because I think the build should fail then too and Heroku should mention that appropriately.
So adding a `retvalue=$?` after the ssh call and exit with `exit $retvalue` or something similar should work.

web/controllers/certificateController/index.ts
```diff
@@ -439,9 +439,6 @@ function loadTemplate(name, lang: Language, fallback: boolean = true): EJS.Clien
 
     let path = `./assets/${name}.${lang}.html`;
 
-    if (process.env.NODE_ENV == 'dev')
-        path += `.example`;
-
```
geroembser, line 444: If we remove these lines, why do we still need the sample files? If we remove them, users running the backend on their machines will see an exception. I think we should rename these files then and remove the .example at the end. Or just leave this line in and somehow add `&& IS_HEROKU !== TRUE`.

And by the way, the Drehtür certificates also use this `.example` stuff, which then should be changed too, whatever solution we choose, see https://github.com/corona-school/backend/blob/fc7d7b701430d0d1dbb84d71119948c7cfe0163c/common/drehtuer/certificates.ts#L18-L20

geroembser (commented):
Awesome work! Bash scripting is fun 😂 Did not know about the `heroku-postbuild` feature... They surprise me again and again 🥳

web/controllers/certificateController/index.ts
Jonasdoubleyou: true, that was laziness ...

setup.sh
Jonasdoubleyou: I went for `|| { exit 1; }` as that scales better, but yeah, proper error handling (aborting the pipeline) was missing

Jonasdoubleyou:
@geroembser can you have a closer look at the other certificate? Can't verify, as I can't send out emails

geroembser (commented):
Thanks for your work so far! 
One other thing that came to my mind: I remember [this](https://coronaschoolgermany.slack.com/archives/G01F5P4PAF5/p1608399630005400?thread_ts=1608392900.001500&cid=G01F5P4PAF5) conversation on Slack where we had some problems with the favicon. In the certificates repository no favicon is included. And if our setup script replaces the entire assets folder, we will no longer have a favicon and we will run into this old problem again. Also see corona-school/backend@9de7831f98bb091e4914f85618cad89e7df292ef that has fixed the problem in the past.

This also leads to the question of what we’re doing with the git-ignored assets. We could unignore them, but ignoring the entire folder also has the advantage of not accidentally committing any assets and thus thinking they are included even though they are replaced during promotion to production. What do you think?

Jonasdoubleyou:
@geroembser ignoring the whole folder sounds good, added the favicon to coronaschool-certificate

geroembser:
Ok, "one more thing" 😂: I think it is very useful to add a custom config property to the `package.json` or a file (similar to how git implements its submodules) that contains the certificate version in the form of the corresponding commit hash of the certificate repository. This way, we could uniquely associate a version of our backend with a specific version of our certificates, making rebuilding reproducible and more deterministic, plus giving us improved versioning. The simplest solution is probably just to add a `.certificate-version` file that just contains the corresponding commit hash and check that out, right after cloning the repository. So parsing the json wouldn't be necessary... Could also be added to tests etc, but that would be way too much for now.

Jonasdoubleyou:
@geroembser yeah "one last thing", we're slowly getting into the territory of overengineering :) Stable versioning makes sense, though I think it would also have worked without ...

geroembser:
You are right, it would have worked without, but at what cost. The reason I decided to bring this up was last night, because I noticed that the current certificate version on master doesn't render to A4 (and the font looks much smaller than on the previous one). So there were three options in my opinion: Introduce a lot of reverts to the master branch of the certificates OR fix the certificates OR add versioning. Fixing the certificates seemed to be a task that I can't estimate, at least approximately, how long it would take me. And versioning, as you thankfully implemented, is a three-liner that we could benefit from in the future as well. So to me, this looked like the best solution if we want to try to have a stable Heroku build this weekend. 
But anyway, I should have noticed the problem with the certificate paper format, but it was too late/early for another topic. So I really think that versioning is something that we're going to benefit from starting this weekend by allowing us to simply deploy a stable version of our certificates.

geroembser (dismissed):
Thanks a lot, really good work! 👍

geroembser (dismissed):

geroembser (approved):

PR #173, opened 2021-02-13
Clean up the createSignature endpoint and make final adjustments for automatic signature process

Jonasdoubleyou:
This is a breakign change, and must be deployed at the same time with the frontend adaptions


lsc0498 (approved):
Thanks a lot!🎉

PR #180, opened 2021-04-11
Setting open match request count to 0 for pupils if requestsAutoMatch is set to false

realmayus:
Backend part of https://github.com/corona-school/project-user/issues/265


realmayus:
The old code already sends the pupil an email once they're matched with a tutor, doesn't it? If not, then I'd have to implement that as well.

lsc0498 (approved):
Looks good 👍

lsc0498:
@realmayus Concerning the email you don't have to care about anything at this point I guess 😉

PR #182, opened 2021-04-17
Added /register/checkEmail endpoint for checking if a specified email already exists

realmayus:
(see https://github.com/corona-school/project-user/issues/120)

PR #184, opened 2021-04-18
/register/checkEmail endpoint + rate-limiting

realmayus:
Incorporates changes from https://github.com/corona-school/backend/pull/182
+ rate-limiting

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2475,11 +2475,6 @@
       "resolved": "https://registry.npmjs.org/date-easter/-/date-easter-0.2.5.tgz",
       "integrity": "sha512-pMvZSqQgPBqnUQEHhRMy7f8Dk3KYY/NPFMFHGK4mU+xM1+SX+zc9IoQMcNtS3HuJjCn3mp0GQwBzsZXz3Thjww=="
     },
-    "date-format": {
-      "version": "3.0.0",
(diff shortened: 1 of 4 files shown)


realmayus:
@geroembser How do I add the new endpoint to the docs? The CI gives me this error: https://github.com/corona-school/backend/pull/184/checks?check_run_id=2375186324

All I added docs-wise so far is this:
https://github.com/corona-school/backend/blob/feat/%23183/web/controllers/registrationController/index.ts#L1115-L1134

I also saw that some kind of API version is mentioned in the docs. Do I need to increase that version number, since I added that endpoint?

geroembser:
> The CI gives me this error: https://github.com/corona-school/backend/pull/184/checks?check_run_id=2375186324

@realmayus It looks like `StatusOK` is the problem. `StatusOk` should work instead, as it was already used in the past.

geroembser:
@realmayus if you wanna test the docs locally, you can run `npm run web:docs`. It may output some warnings, but as long as the return code is ok, the Github Action will accept this. So basically `(npm run web:docs > /dev/null && echo 'DOC OK') || echo 'DOC ERROR'` should be a simple test that you can use on your computer to "check" the docs. Probably better than pushing and waiting for Github :joy:

geroembser (approved):
Looks good to me 👍

PR #186, opened 2021-04-28
Backend part of interest-confirmation for pupils

geroembser:
This PR will introduce a new (temporary) feature to support a new intermediate step in the tutoring matching process. 

It adds the required new backend APIs and changes to the existing APIs to work with an additional and necessary interest-confirmation before pupils will get assigned a match. 

The idea behind this interest-confirmation: Let pupils confirm their interest before they get a new match, such that Corona School can ensure a higher matching quality. At the moment, our waiting time for pupils is quite long, so it is very likely that not every pupil registered two weeks ago still needs our support. 

Note: this restrictions do not apply for pupils coming from one of our partner schools/organizations. We then always know/assume that they are still interested in our project. 

## TODO
- [x] add database migration

### DIFF
--- a/common/interest-confirmation/tutoring/receiver/people/db-queries.ts
+++ b/common/interest-confirmation/tutoring/receiver/people/db-queries.ts
@@ -1,0 +1,56 @@
+import { EntityManager, SelectQueryBuilder } from "typeorm";
+import { tuteesToMatchQuery } from "../../../../administration/match-making/tutoring/people/tutees";
+import { RegistrationSource } from "../../../../entity/Person";
+import { Pupil } from "../../../../entity/Pupil";
+import { InterestConfirmationStatus } from "../../../../entity/PupilTutoringInterestConfirmationRequest";
+
+// Raw Queries:
+// ⎺⎺⎺⎺⎺⎺⎺⎺
+function allMatchablePupilsThatRequireInterestConfirmationQuery(manager: EntityManager) {
+    //only pupils not registered through partner schools require interest confirmation
+    return tuteesToMatchQuery(manager).andWhere("p.registrationSource != (:rs)", { rs: RegistrationSource.COOPERATION});
+}
+function allMatchablePupilsWithoutInterestConfirmationRequestQuery(manager: EntityManager) {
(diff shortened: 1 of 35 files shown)


common/interest-confirmation/tutoring/receiver/people/db-queries.ts
```diff
@@ -0,0 +1,56 @@
+import { EntityManager, SelectQueryBuilder } from "typeorm";
+import { matchableTuteesQuery } from "../../../../administration/match-making/tutoring/people/tutees";
+import { RegistrationSource } from "../../../../entity/Person";
+import { Pupil } from "../../../../entity/Pupil";
+import { InterestConfirmationStatus } from "../../../../entity/PupilTutoringInterestConfirmationRequest";
+
+// Raw Queries:
+// ⎺⎺⎺⎺⎺⎺⎺⎺
+function allMatchablePupilsThatRequireInterestConfirmationQuery(manager: EntityManager) {
```
Jonasdoubleyou, line 9: @geroembser could you please take the time to clean this up? It took us about 1 hour to figure out what the hell this code does. I also doubt that having one function for each step improves readability.

geroembser: I don’t see why. As I’ve said, this was just a temporary solution. As far as I know this interest confirmation system will be completely replaced by a more sophisticated _match_ confirmation system in the very near future. (actually it should have been replaced a few weeks ago, but that's another story...)
If you have an idea how to clean this up easily – I’m open to that. For example, rewriting that stuff with Prisma would definitely help. As this would solve the problem why I have introduced that separate functions: working with the queries being pure strings and have no checks for the underlying model (until runtime) made it difficult to develop and test since I had to change this string in different places (and that annoyed because I’m really not familiar with typeorm's raw sql query syntax). 
So I really don't see why I should bother at this point. Then I would directly implement the new match confirmation system. The "specification" process here should be completed now… (but I guess it hasn't left Slack until now...)

Jonasdoubleyou: Ah, didn't know that. I'm fine leaving it as is for now if it is replaced anyway in the (hopefully) near future.

PR #187, opened 2021-05-01
Added login check to see if user is verified

realmayus:
This adds a check to the login procedure which checks if the user is verified. If they aren't yet, a new verification email will be sent to [enforce e-mail verification](https://github.com/corona-school/project-user/issues/267)

### DIFF
--- a/web/controllers/tokenController/index.ts
+++ b/web/controllers/tokenController/index.ts
@@ -15,6 +15,7 @@ import {
     sendFirstScreeningInvitationToProjectCoachingJufoAlumni,
     sendFirstScreeningInvitationToTutor
 } from "../../../common/administration/screening/initial-invitations";
+import { generateToken, sendVerificationMail } from "../../../jobs/periodic/fetch/utils/verification";
 
 const logger = getLogger();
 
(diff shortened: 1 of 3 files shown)


web/controllers/tokenController/index.ts
```diff
@@ -178,7 +182,11 @@ export async function getNewTokenHandler(req: Request, res: Response) {
             }
 
             if (person !== undefined) {
-                if (allowedToRequestToken(person)) {
+                if(person.verifiedAt == null) {
+                    status = 409;
```
lsc0498, line 186: What happens in the frontend if the api call returns `409`? The user should be encouraged to check his mails and a possible error message may be misleading except it states that a verification token is sent.

realmayus: The frontend catches this error and shows an error message saying that the user should check their emails

lsc0498: Alright, sounds good👍

lsc0498 (approved):
Looks good, thank you! Please don't forget to fix linter errors before merging 😉

lsc0498 (approved):
Great, thank you! 😊

PR #188, opened 2021-05-06
Update certificate version

Jonasdoubleyou:
- footer now looks like one
- signature is a bit smaller
(changes requested by @tobiasbork )


lsc0498 (approved):
Top, thank you!

PR #191, opened 2021-05-16
Added dupe check when editing email of student as a screener

### LINKED ISSUE #148: Possible violation of email uniqueness constraint
When editing a student's e-mail address as a screener we currently don't check if the new e-mail already exists which mail violate the uniqueness constraint of e-mail addresses without notice (also see #146)

realmayus:
Closes #148

### DIFF
--- a/web/controllers/screeningController/index.ts
+++ b/web/controllers/screeningController/index.ts
@@ -247,7 +247,6 @@ export async function updateStudentByMailHandler(req: Request, res: Response, ne
     if (studentInfo.isValid()) {
         //save old state of student info
         const prevState = await StudentInfoDTO.buildFrom(student);
-        
 
         if (studentInfo.email.toLowerCase() !== req.params.email.toLowerCase()) {
             // Check if email already exists, but only if it differs from the request params (which happens when we don't change the email address)
@@ -256,7 +255,7 @@ export async function updateStudentByMailHandler(req: Request, res: Response, ne
                 logger.error("Email already in use");
                 res.status(409).send("Email already in use!");
                 return;
-            }    
+            }
         }
 
         //update student info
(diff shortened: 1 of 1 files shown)


realmayus:
I assume that there is some kind of frontend work that has to be done. Does editing a student's email happen in the screening frontend or in the web-user-app?

web/controllers/screeningController/index.ts
```diff
@@ -245,7 +247,15 @@ export async function updateStudentByMailHandler(req: Request, res: Response, ne
     if (studentInfo.isValid()) {
         //save old state of student info
         const prevState = await StudentInfoDTO.buildFrom(student);
-
+        
+        // Check if email already exists
+        const email_dupe_res = await entityManager.findOne(Student, {email: studentInfo.email.toLocaleLowerCase()});
```
lsc0498, line 252: This will find the student we are editing unless we don't change the email address und will thus throw an error in all cases in which we edit something but the mail address 😉

lsc0498:
> I assume that there is some kind of frontend work that has to be done. Does editing a student's email happen in the screening frontend or in the web-user-app?

Yes, in the student editing form there is the possibility to edit the mail address. The form is reachable either through the search line or via the route `student/:email`. The thrown error should be interpreted there correctly such that the screener knows what to do in case of an error.

web/controllers/screeningController/index.ts
```diff
@@ -245,7 +247,15 @@ export async function updateStudentByMailHandler(req: Request, res: Response, ne
...
+        if (email_dupe_res !== undefined) {
+            logger.error("Email already in use");
+            res.status(409).send("Email already in use!");
+            return
```
lsc0498, line 256: Don't forget to fix these linter errors ;-)

realmayus:
> This will find the student we are editing unless we don't change the email address und will thus throw an error in all cases in which we edit something but the mail address 😉
@lsc0498 
This can be avoided by checking whether the email address passed in the request params is different from that provided in the request body, as the student is identified by the email in the params, right? I just pushed a commit implementing a check for that, which basically skips the dupe check if the URL in the params doesn't differ from the one provided in the body, hoping that it prevents the false positive.

lsc0498 (approved):
This looks good now. Thank you! 👍
We can deploy this but as a next step a dedicated error message in the screening frontend would be nice such that the screener who types in an existing address knows why they get an error.

PR #192, opened 2021-05-16
Now filtering out minGrade and maxGrade when pupil tries to put a subject

### LINKED ISSUE #141: Pupils are allowed to set minGrade and maxGrade for their subjects
The putSubjectsHandler function doesn't check if a pupil or a student wants to update their subject. So it's possible for pupils to set minGrade and maxGrade, but pupils only need to seed the subject's name. On the other hand students need to be able to update name, minGrade and maxGrade.
A check like `ìf res.locals.user instanceof pupil/student` should be implemented.

Additionally the frontend should be checked and possibly updated that requests to update a pupil's subject is only allowed to contain the name.

https://github.com/corona-school/backend/blob/2ddde97b3c5669e63c2fcd4bac3d980ea71f49c6/web/controllers/userController/index.ts#L212-L257

realmayus:
closes #141

### DIFF
--- a/web/controllers/userController/index.ts
+++ b/web/controllers/userController/index.ts
@@ -862,6 +862,13 @@ async function putSubjects(wix_id: string, req: ApiSubject[], person: Pupil | St
         return 403;
     }
 
+    // filter out mingrade/maxgrade if user is pupil, since they aren't allowed to change that
+    if (person instanceof Pupil) {
+        req = req.map(x => {
+            return { name: x.name };
+        })
+    }
+
     const oldPerson = Object.assign({}, person);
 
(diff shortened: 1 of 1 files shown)


Jonasdoubleyou (changes requested):
If the user is not authorized to perform a certain operation, the backend should clearly indicate that by responding with an Error. Also this seems to be the wrong place to filter out minGrade / maxGrade, as there are a few other APIs that still return it.

Jonasdoubleyou:
The GraphQL API validates this properly and as we probably throw away this API soonish, I guess we don't need this PR anymore ...

PR #194, opened 2021-05-23
replaced verifiedAt with verification in StudentInfoDTO

realmayus:
Now passing the verification string to the screening backend, as this is a more reliable way of telling if a student has a verified email address.


lsc0498 (approved):
Thank you!

PR #196, opened 2021-06-07
Refactor Course Controller

Jonasdoubleyou:
While looking for whatever keeps the backend crashing, I came across the beautiful course controller and had the urgent need to start refactoring it ... Do whatever you like with this PR (once it's ready)

What I did:
- Removed unnecessary index based loops and replaced them with for...of loops
- Unnested nested if..elses 
- Moved Validation of parameters and body into a reusable function
- Removed all "returns an error code as number or the result" and replaced with a thrown HTTPError or returned result
- Updated Typescript to be able to use string template types

This reduced the file size of the course controller by ~30%.

What I did not (but noticed):
- Inside the transactions a lot of logic gets processed. If a lot of users join the same course, this will lead to rollback / retries, which means that all the queries will potentially happen multiple times. This seems very inefficient. Locking per user / lecture on serverside should be sufficient instead
- The retrieval of Courses / Subcourses / Lectures could be moved into reusable functions
- The validation of entities could be moved into the entity (or directly use a framework that embeds it, e.g. some GraphQL stuff), just like most of the logic, only keeping input validation / authorization in the endpoint itself

### DIFF
--- a/web/controllers/courseController/index.ts
+++ b/web/controllers/courseController/index.ts
@@ -57,6 +57,8 @@ import { CourseGuest, generateNewCourseGuestToken } from '../../../common/entity
 import { getCourseCertificate } from '../../../common/courses/certificates';
 import InstructorIssuedCertificateEvent from '../../../common/transactionlog/types/InstructorIssuedCertificateEvent';
 import { addCleanupAction } from '../../../common/util/cleanup';
+import { handleError, HTTPError } from '../error';
+import * as Validation from "../validation";
 
 const logger = getLogger();
 
@@ -98,8 +100,7 @@ const logger = getLogger();
  * @apiUse StatusInternalServerError
  */
 export async function getCoursesHandler(req: Request, res: Response) {
-    let status = 200;
-    try {
+    handleError(res, async () => {
         let authenticatedStudent = (res.locals.user instanceof Student);
         let authenticatedPupil = (res.locals.user instanceof Pupil);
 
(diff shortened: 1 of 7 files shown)


web/controllers/courseController/index.ts
```diff
@@ -988,120 +783,62 @@ async function postCourse(student: Student, apiCourse: ApiAddCourse): Promise<Ap
...
     if (course == undefined) {
-        logger.warn(`User tried to add subcourse to non existent course (ID ${courseId})`);
-        logger.debug(student, apiSubcourse);
-        return 404;
+        throw new HTTPError(404, `User tried to add subcourse to non existent course (ID ${courseId})`);
     }
 
-    let authorized = false;
-    for (let i = 0; i < course.instructors.length; i++) {
-        if (student.id == course.instructors[i].id) {
```
Jonasdoubleyou, line 1056: Here it uses the student.id

web/controllers/courseController/index.ts
```diff
@@ -1310,144 +995,75 @@ async function postLecture(student: Student, courseId: number, subcourseId: numb
...
-            }
-        } else {
-            status = 403;
-            logger.warn("A non-student wanted to edit a course");
-            logger.debug(res.locals.user);
+        Validation.hasParams(req, "id");
+        Validation.hasBody(req, { instructors: "string[]", description: "string", allowedContact: "boolean", name: "string?", outline: "string?", correspondentID: "string?", tags: "string[]" });
+        if (
+            (req.body.outline !== undefined && typeof req.body.category !== 'string') ||
+            (req.body.outline !== undefined && typeof req.body.submit !== 'boolean')) {
```
Jonasdoubleyou, line 1005: @lsc0498 Do you know why this logic exists? Why does the outline have to not set the category? (This logic is equivalent in the pre-edit state)

Jonasdoubleyou:
I verified that
- handleError responds / logs properly
- Validation.hasBody works correctly for all types
- Couldn't validate Validation.hasParams, cause I can't find a route that matches a route without a parameter being set

web/controllers/courseController/index.ts
lsc0498: This doesn't make much sense. To me it looks much like a copy-paste error 😅

Jonasdoubleyou: Alright, removed that logic

Jonasdoubleyou:
Closing for now as I do not have time to keep up with upstream. Hopefully we might be able to move the frontend to GraphQL soon, throwing away all the Web controllers.

PR #197, opened 2021-06-13
Increased course guest limit

realmayus:
Increases course guest limit to 10 guests per course; as per https://github.com/corona-school/project-user/issues/333.


Jonasdoubleyou (approved):
Long term we might want to move that magic number into a config variable

PR #199, opened 2021-06-22
ReTool Endpoints

lsc0498:
This PR implements the endpoints required for the admin tool implemented with ReTool.


Jonasdoubleyou:
Maybe we can also directly experiment with [autogeneration from prisma schemas](https://github.com/MichalLytek/typegraphql-prisma#readme), just as an idea

Jonasdoubleyou:
@lsc0498 do we still need compability with NodeJS 13? Or can that check be removed (I upgraded the Buildpack to use NodeJS 14)

Jonasdoubleyou (dismissed):
Approving as this PR was originally opened by @lsc0498 .

Jonasdoubleyou (approved):

Jonasdoubleyou (dismissed):

lsc0498 (commented):
@Jonasdoubleyou Very good job, I really like it 👍 Before merging I would like to implement a little more state-of-the-art authentication model though. I fear that my "API key in the header" method is not quite appropriate for an API which can do literally everything 😅

prisma/schema.prisma
```diff
@@ -2,6 +2,11 @@ generator client {
   provider = "prisma-client-js"
 }
 
+// generator typegraphql {
```
lsc0498, line 5: @Jonasdoubleyou I guess this is commented out since the generated GraphQL code is committed? In order to update that code (when the data model changes) does one simply has to uncomment this in order to generate the new GraphQL endpoints?

lsc0498:
> @lsc0498 do we still need compability with NodeJS 13? Or can that check be removed (I upgraded the Buildpack to use NodeJS 14)
> By the way, v14 is LTS (and soon enters Maintenance in September) whereas v13 is not an LTS version (https://nodejs.org/en/about/releases/). Thus I guess it makes sense to upgrade to v14 and then follow the LTS schedule

No thats fine, we can (finally) switch to v14 😅

prisma/schema.prisma
Jonasdoubleyou: exactly!

Jonasdoubleyou (approved):

PR #200, opened 2021-06-23
Experiment with Prisma based migrations

Jonasdoubleyou:


Jonasdoubleyou:
As Prisma for some reasons needs `CREATE DATABASE` permissions to create a migration, we would need to run `prisma generate dev` either locally on a local database, or create [a dedicated migration database](https://www.prisma.io/docs/concepts/components/prisma-migrate/shadow-database#detecting-schema-drift). Also the Prisma migration API seems to be a "feature preview" and is not yet finalized.

Jonasdoubleyou:
I'd prefer to wait for `prisma migrate` being more mature for now ...

PR #203, opened 2021-07-11
Raised course limit to 10

realmayus:
I recognize that this should be part of an easily modifiable config at some point, but I figured that it would be better to change the (hardcoded) constant right away because of the urgency of the issue.

Closes https://github.com/corona-school/project-user/issues/339


lsc0498 (approved):
Looks good. Thank you!

PR #204, opened 2021-07-16
Allow deactivation reason and feedback

realmayus:
With this change, the frontend can provide a deactivation reason as well as additional feedback when the user decides to deactivate their account.
The additional information then gets written to the transaction log as part of the DeActivateEvent.

See https://github.com/corona-school/project-user/issues/302 for more details.


lsc0498 (approved):
Looks good. Thank you! 🙏

PR #209, opened 2021-08-05
Move config variable from NODE_ENV to ENV

Jonasdoubleyou:
This way NODE_ENV can be set to production by default (which causes Heroku to drop dev dependencies during build, reducing the slug size to about 400MB, which is 100MB away from the hard limit which we currently run into)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -20,7 +20,7 @@ MAILJET_PASSWORD=
 MAILJET_LIVE=0
 
 # Run webserver in dev mode on port 5000 (http only)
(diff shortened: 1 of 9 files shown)


lsc0498 (approved):

Jonasdoubleyou:
@lsc0498 sorry for the second review, the pruneDependencies feature of Heroku had a bug and also deleted productive dependencies (didn't see that coming 😅 )

lsc0498 (approved):

PR #212, opened 2021-08-08
Passing waiting list size in course request, added more pupils in dev.ts

realmayus:
As per https://github.com/corona-school/project-user/issues/284


lsc0498 (approved):
Looks good! 👍

PR #215, opened 2021-08-09
Add Match Creation and Dissolve Mutations

Jonasdoubleyou:
https://github.com/corona-school/project-user/issues/346


graphql/pupil/mutations.ts
```diff
@@ -58,4 +60,25 @@ export class MutatePupilResolver {
...
+
+    @Mutation(returns => Boolean)
+    @Authorized(Role.ADMIN)
+    async pupilCreateMatchRequest(@Arg("pupilId") pupilId: number, @Ctx() context: GraphQLContext): Promise<boolean> {
+        const pupil = await getPupil(pupilId);
+        const isAdmin = context.user.roles.includes(Role.ADMIN);
+
+        await createMatchRequest(pupil, isAdmin);
+
+        return true;
```
lsc0498, line 72: What happens if `createMatchRequest` doesn't go through?

lsc0498 (approved):
Looks pretty good! 👍 Sorry for the delay by the way🙈

graphql/pupil/mutations.ts
Jonasdoubleyou: Then it throws an error back to the caller (e.g. the Retool) which can show an error message to the user. Actually I wanted to return nothing here (a "void method"), but unfortunately a Void type is [not supported](https://stackoverflow.com/questions/44737043/is-it-possible-to-not-return-any-data-when-using-a-graphql-mutation). Thus I chose to return `true` which is somewhat similar. I'll add that to the documentation somewhen 😅

lsc0498: Alright, it's fine. I just found the `return true;` a little suspicious 😅

Jonasdoubleyou: Yeah, would also prefer if TypeGraphQL would support it ...

PR #216, opened 2021-08-14
GraphQL: @Authorzation enforcement & Cardinality Limitation

Jonasdoubleyou:
I made this PR just for fun because I wanted to play around with ESLint and TypeGraphQL. Anyways I guess this is useful nevertheless. This implements two features: 

**Cardinality Limitation**

As we expose very powerful query methods in our GraphQL graph, users can write queries that put a very heave load on both the database and the backend (e.g. get all participants of courses in which all pupils participate, works also recursively). 
At the moment GraphQL is only used by "trusted users" and as such we can prevent such queries being written, but in the future one might open it to others (e.g. Team QM to do data science? The frontend?). That's why I've added some basic validations were added to limit the result set (there might still be corner cases were this does not work). This enforces both the usage of proper pagination on queries (with TAKE and SKIP), and prevents associations getting to large (not more than 1000 entries). 
The later can be controlled through the `@LimitEstimated(10)` annotation on the field resolver (no limitations applied if the annotation is not there). 

**@Authorized enforcement**

When one writes a FieldResolver or Mutation and forgets the `@Authorized` annotation, the query or action can be executed by _anyone_, especially for mutations this is very problematic (a version with missing authorization nearly went to dev). To prevent this in the future I've added a compile time check through a custom ESLint Rule that ensures that inside Resolvers, all methods are annotated. In a Pull Request this looks like this:

image

And in VS Code one can even automatically fix this:

image

resolves https://github.com/corona-school/project-user/issues/366


Jonasdoubleyou:
@realmayus requested your review as I've reviewed quite some of your code, so now it's your turn ;)

realmayus (approved):
LGTM, thank you!

lsc0498 (approved):
Also looks good to me. Nice idea! Thank you 👍

PR #218, opened 2021-08-22
Checking whether *our* student is the sole instructor in the course before cancelling it

realmayus:


web/controllers/userController/index.ts
```diff
@@ -1007,7 +1011,66 @@ async function putActive(wix_id: string, active: boolean, person: Pupil | Studen
...
             }
 
-            // Step 2: Deactivate
+            // Step 2: Cancel all courses if user is student
+            if (type == Student) {
+                let courses = await getConnection()
+                    .getRepository(Course)
+                    .createQueryBuilder("course")
+                    .leftJoinAndSelect("course.instructors", "instructors")
+                    .getMany();
```
Jonasdoubleyou, line 1020: Do we really need to fetch _all courses_ from the database? The number of courses will continuously grow, and so does the runtime of this query. I guess we could already filter for `.where("instructors.id = :id", { id: person.id })`, right?

web/controllers/userController/index.ts
```diff
@@ -1007,7 +1011,66 @@ async function putActive(wix_id: string, active: boolean, person: Pupil | Studen
...
+
+                courses.forEach(async (course: Course) => {
```
Jonasdoubleyou, line 1022: This _fires and forgets_ all transactions, which I guess is not intended ... I'd rather do a `for(const course of courses)` for sequential processing or a `await Promise.all(courses.map(...))` for proper error handling ....

web/controllers/userController/index.ts
```diff
@@ -1007,7 +1011,66 @@ async function putActive(wix_id: string, active: boolean, person: Pupil | Studen
...
+                    if (course.instructors.some(i => i.id === person.id)) { // Only proceed if we're part of this course as an instructor
+                        if (course.instructors.length > 1) {
+                            // Course still has other instructors, only remove our person from those. We don't want to cancel those courses.
+                            course.instructors = course.instructors.filter(s => s.id !== person.id);
+                            entityManager.transaction(async em => {
```
Jonasdoubleyou, line 1027: This could be lifted up, so that all operations are part of one transaction (which then gets rolled back as a whole if bad stuff happens)

web/controllers/userController/index.ts
```diff
@@ -1007,7 +1011,66 @@ async function putActive(wix_id: string, active: boolean, person: Pupil | Studen
...
+            // Step 2: Cancel all courses if user is student
+            if (type == Student) {
+                let courses = await getConnection()
+                    .getRepository(Course)
+                    .createQueryBuilder("course")
+                    .leftJoinAndSelect("course.instructors", "instructors")
+                    .getMany();
+
+                courses.forEach(async (course: Course) => {
+                    if (course.instructors.some(i => i.id === person.id)) { // Only proceed if we're part of this course as an instructor
```
Jonasdoubleyou, line 1023: As a small stylistic note: Through "short circuiting" one can avoid having a lot of nested blocks which becomes quite hard to read, e.g. through inverting the condition `if(!course.instructors.some(it => it.id === person.id)) continue;`

web/controllers/userController/index.ts
```diff
@@ -1007,7 +1011,66 @@ async function putActive(wix_id: string, active: boolean, person: Pupil | Studen
...
+                                logger.info("Removed instructor " + person.firstname + " " + person.lastname + " from course " + course.name + ".");
+                            });
+                        } else {
+                        // Our person is the only instructor in the course. Cancel it.
+
+                            // We have a non-mitigated race condition here: Someone could post a new subcourse into the course, while the course gets cancelled
+                            try {
+                            // Run in transaction, so we may not have a mixed state, where some subcourses are cancelled, but others are not
+                                await entityManager.transaction(async em => {
+                                    if (course.hasOwnProperty("subcourses")) {
```
Jonasdoubleyou, line 1038: Are there cases when the subcourses property is not set at all? Shouldn't it be an empty array then?

web/controllers/userController/index.ts
```diff
@@ -1007,7 +1011,66 @@ async function putActive(wix_id: string, active: boolean, person: Pupil | Studen
...
+                                        for (let i = 0; i < course.subcourses.length; i++) {
```
Jonasdoubleyou, line 1039: I'd prefer a `for(const subcourse of  course.subcourses)` here ...

web/controllers/userController/index.ts
```diff
@@ -1007,7 +1011,66 @@ async function putActive(wix_id: string, active: boolean, person: Pupil | Studen
...
+                                                await em.save(Subcourse, course.subcourses[i]);
+                                                sendSubcourseCancelNotifications(course, course.subcourses[i]);
+                                            }
+                                        }
+                                    }
+
+                                    course.courseState = CourseState.CANCELLED;
+                                    await em.save(Course, course);
+
+                                }).catch(e => {
```
Jonasdoubleyou, line 1051: Can we really handle the error at this point? Through catching here the user gets deactivated, but some courses are in an inconsistent state (I know that we have similar patterns at other places though). I'd personally prefer to "fail and rollback" instead of "half deleting" the user.

web/controllers/userController/index.ts
```diff
@@ -1007,7 +1011,66 @@ async function putActive(wix_id: string, active: boolean, person: Pupil | Studen
...
+                                });
+
+                                transactionLog.log(new CancelCourseEvent(person as Student, course));
+                                logger.info("Successfully cancelled course");
+
+                                return 204;
+                            } catch (e) {
+                                logger.error("Can't cancel course: " + e.message);
+                                logger.debug(course, e);
+                                return 500;
```
Jonasdoubleyou, line 1063: These `return`s go into nowhere.

Jonasdoubleyou (changes requested):

web/controllers/userController/index.ts
realmayus: Oh yeah, I wasn't aware of that. Thanks

web/controllers/userController/index.ts
realmayus: Oh yeah, it's not nullable.

web/controllers/userController/index.ts
realmayus: I see what you mean. How would the rollback work? Is this handled automatically by the entityManager if it wraps the entire code? (see latest commit).
And yeah, this code block was largely copied from another endpoint, and I didn't pay much attention to those things you've commented on. I shouldn't have created a PR, since it wasn't production ready :(

realmayus: @Jonasdoubleyou

web/controllers/userController/index.ts
```diff
@@ -1007,7 +1011,58 @@ async function putActive(wix_id: string, active: boolean, person: Pupil | Studen
...
+                    .leftJoinAndSelect("course.instructors", "instructors")
+                    .where("instructors.id = :id", { id: person.id})
+                    .getMany();
+
+                console.log("CRS: ", courses);
+
+                entityManager.transaction(async em => {
+                    for (const course of courses) {
+                        if (!course.instructors.some(i => i.id === person.id)) { // Only proceed if we're part of this course as an instructor
+                            return;
```
Jonasdoubleyou, line 1028: Ah, my comment was slightly missleading here, in a `.forEach`, `return` returns control to the forEach logic and continues with the next entry, in a `for of` you'd rather need a `continue;` here ...

web/controllers/userController/index.ts
```diff
@@ -1007,7 +1011,58 @@ async function putActive(wix_id: string, active: boolean, person: Pupil | Studen
...
+            // Step 2: Cancel all courses if user is student
+            if (type == Student) {
+                let courses = await getConnection()
+                    .getRepository(Course)
+                    .createQueryBuilder("course")
+                    .leftJoinAndSelect("course.instructors", "instructors")
+                    .where("instructors.id = :id", { id: person.id})
+                    .getMany();
+
+                console.log("CRS: ", courses);
```
Jonasdoubleyou, line 1023: Before merging this should be replaced by a proper `logger.debug` ... (just marking so that we don't forgetit)

web/controllers/userController/index.ts
Jonasdoubleyou: Yes, it would if you let all errors bubble up to the entity manager callback (which catches the error, aborts the transaction, and then rethrows it). So just remove the inner `try { ... } catch(e) { .. }` and let the error bubble up.

Jonasdoubleyou: The TypeORM documentation is lacking as always, you can find the implementation behind `.transaction(...)` [here](https://github.com/typeorm/typeorm/blob/e9366b33ddff296de1254019589b85e40aa53e12/src/entity-manager/EntityManager.ts#L145)

Jonasdoubleyou:
@lsc0498 do you know why there are instructors of the Subcourse? Are all Subcourse instructors also Course instructors? (cause otherwise we would also have to look at Subcourses here)

web/controllers/userController/index.ts
realmayus: Yeah right, I have yet to test it.

lsc0498:
To my knowledge, this is pretty redundant. The reason is probably historical and the initial intention was to have the `course` as some kind of "template" and the `subcourse` as the actual course where the subcourses instructors are a subset of the courses instructors. However, in practise, the instructors of the courses are identical to the instructors of the subcourses. 
One of the many weird things in our course logic... 🙈

Jonasdoubleyou:
@lsc0498 well then why can a course be cancelled if it is just a "template"? I would strongly suggest to think of proper semantics here and document them in the entity.

lsc0498:
Yes, you're right. It doesn't make much sense anymore and was also never used like this.

Jonasdoubleyou:
@lsc0498 then I'd suggest we add the fields of the course to the subcourse, create a migration to fill these fields, drop all uses of Course and delete the entity. It bothers me that we know that there are "many weird things in our course logic..." yet there seems to be no plan to get rid of those, not even backlog issues. Anyways, for this PR this means that the current logic should be fine.

realmayus:
I just tested the code thoroughly, and found out that, for some reason, there are problems with the "Gitarre lernen für Anfänger" example course.
When I log the course, it looks like this (note the instructors array, it only has one entry)
image

Then take a look at the code that creates this course:
https://github.com/corona-school/backend/blob/b59c8ee7476009eb0ad0bc7e6f1f04c28117279b/web/dev.ts#L761
Curiously, it says that there are two instructors in the course here.

Is this a relic of the old course logic as well? Can this happen in production? 🤔 @lsc0498

realmayus:
Oh. I just realized that this happens because I added `.where("instructors.id = :id", { id: person.id})` to the query 🙈 That explains my issue, apologies!
@Jonasdoubleyou re: https://github.com/corona-school/backend/pull/218#discussion_r694695930
Is there a way to avoid fetching all courses while also retaining the full instructors list for the courses that match the criteria? When I remove the .where expression you suggested, the issue doesn't occur anymore...

Jonasdoubleyou:
I guess we can join the instructors twice then, and use them once for the WHERE clause and once to retrieve the results:

```
.leftJoinAndSelect("course.instructors", "instructors")
.leftJoin("course.instructors", "instructorSelect")
.where("instructorsSelect.id = :id", { id: person.id})
.getMany()
```

realmayus:
Neat, thanks!

Jonasdoubleyou (approved):
Looks good to me.

Please double check if we read from the subcourse.instructors somewhere, cause in that case we would also have to update those ...

Jonasdoubleyou (approved):

PR #220, opened 2021-09-03
HOTFIX: Fix match request validation to resolve issues with Modal

Jonasdoubleyou:
During the user settings update modal, the whole user account gets posted to the PUT /api/user endpoint. 
If an Admin manually matched a pupil, the endpoint will reject the user, and as such the Modal will end with an error. 
See https://app.logdna.com/3cfa259833/logs/view?t=timestamp%3A1630659184908&a=1630659184908.1386885141898444803&q=%229fee9e5a-55e3-485d-bab7-981a4ff13021%22 for a concrete occurence. 

To resolve this, the logic was adapted to only prevent users from increasing the match count. 

This should resolve https://github.com/corona-school/project-user/issues/384


lsc0498 (approved):
Looks good thank you!

PR #224, opened 2021-09-08
Add notification import and update, improve logging

Jonasdoubleyou:
Apart from improved logging, this provides means to import notifications into a system. Through that we can easily replicate the Notification configuration from QA and PROD and  vice versa. Through a "dry run", we can check whether the configuration can be applied and what it does. This way we can add such functionality to Retool and let content admins manage those (less work for the Tech Team). 

In theory we could also replicate the PROD configuration every time the QA environment is rebuilt


realmayus (approved):
Looks good to me!

PR #225, opened 2021-09-10
GraphQL Various

Jonasdoubleyou:
Added field resolver for subcourses where the pupil is on the waitinglist:

```
query { 
	pupils(take: 100) { 
          id
  	  subcoursesJoined { id }
          subcoursesWaitingList { id }
  }
}
```

This is a breaking change (renaming "subcourses" to "subcoursesJoined", I'll take care of adapting the Retool)

Also added the following mutations:
```
mutation {  pupilJoinSubcourseWaitinglist(pupilId: 1 subcourseId: 7) }
mutation {  pupilLeaveSubcourseWaitinglist(pupilId: 1 subcourseId: 7) }
mutation { matchDissolve(matchId: 1, dissolveReason: 10) }
```

Additionally adds Prisma based transaction logging to the mutations. 
To verify the transaction log, one can run (as Admin):
```
query { 
	pupils(where: { id: { equals: 1}}) { 
  	logs {  
      	  id 
          logtype
          createdAt
          user
          data
        }
    }
}
```

[Plaground](https://backend-feat-graphql-va-lpwph7.herokuapp.com/apollo)
```

### CODE REVIEW

File: graphql/pupil/mutations.ts [hunk 1]
```diff
@@ -50,6 +50,28 @@ export class MutatePupilResolver {
         return true;
     }
 
+    @Mutation(returns => Boolean)
+    @Authorized(Role.ADMIN)
```
realmayus, line 54: Is this endpoint supposed to be restricted to admins? Shouldn't a pupil be able to join the waiting list themselves? Or is this an admin-only endpoint and there's a different one for pupils?

realmayus (commented):
LGTM!

realmayus (approved):

common/transactionlog/log.ts
```diff
@@ -0,0 +1,56 @@
+import { log as Log, log_logtype_enum as LogType, pupil as Pupil, student as Student, course_attendance_log as CourseAttendanceLog, bbb_meeting as BBBMeeting } from "@prisma/client";
+import { getLogger } from "log4js";
+import { MentoringCategory } from "../mentoring/categories";
+import { prisma } from "../prisma";
+
+/* To ensure consistency with existing logs,
+  the JSON data in the log is validated against this schema */
+type LogData<Type extends LogType> = {
+    misc: never,
```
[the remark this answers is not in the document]
Jonasdoubleyou: Yeah, the log type "misc" does not make sense IMO, thus the new logging function can not be called with this type as no parameter would fullfill the never type

graphql/pupil/mutations.ts
Jonasdoubleyou: Yeah, we currently do not have user authentication for the GraphQL endpoint. If you have time you could also have a look at the "GraphQL Authentication" Pull Request which adds exactly that ...

lsc0498 (commented):
Looks good! Sorry for the delay🙈

PR #228, opened 2021-09-12
Create entity for CoC

ayushpandey8439:
Created Entity for CoC. with relation to Student and Screener entity in TypeORM and Prisma.


common/entity/CertificateOfConduct.ts
```diff
@@ -0,0 +1,40 @@
...
+    @Column({
+        nullable: false
+    })
+    criminalRecords:Boolean;
+
+    @ManyToOne((type) => Screener, (inspectingScreener) => inspectingScreener.screenings, {
+        eager: true
+    })
+    @JoinColumn()
+    inspactingScreener: Screener;
```
Jonasdoubleyou, line 32: Typo, we should avoid that in the datamodel :)

common/entity/Student.ts
```diff
@@ -29,6 +29,8 @@ import {ExpertData} from "./ExpertData";
 import { CourseGuest } from "./CourseGuest";
 import { Language } from "../daz/language";
 import * as Notification from "../notification";
+import {CertificateOfConduct} from "./CertificateOfConduct";
+import {CertificateFingerprint} from "aws-sdk/clients/cloudhsm";
```
Jonasdoubleyou, line 33: Unused import?

Jonasdoubleyou (changes requested):

common/migration/1635880160929-AddUpdatedAtInCOC.ts
```diff
@@ -0,0 +1,148 @@
...
+        await queryRunner.query(`ALTER TABLE "pupil" ALTER COLUMN "verifiedAt" DROP DEFAULT`);
+        await queryRunner.query(`COMMENT ON COLUMN "pupil"."verifiedAt" IS NULL`);
+        await queryRunner.query(`ALTER TABLE "pupil" ALTER COLUMN "verification" DROP DEFAULT`);
+        await queryRunner.query(`COMMENT ON COLUMN "pupil"."verification" IS NULL`);
+        await queryRunner.query(`ALTER TABLE "pupil_tutoring_interest_confirmation_request" ALTER COLUMN "reminderSentDate" DROP DEFAULT`);
+        await queryRunner.query(`COMMENT ON COLUMN "pupil_tutoring_interest_confirmation_request"."reminderSentDate" IS NULL`);
+        await queryRunner.query(`ALTER TABLE "project_match" ALTER COLUMN "dissolveReason" DROP DEFAULT`);
+        await queryRunner.query(`COMMENT ON COLUMN "project_match"."dissolveReason" IS NULL`);
+        await queryRunner.query(`ALTER TABLE "certificate_of_conduct" RENAME COLUMN "updatedAt" TO "inspectingScreenerId"`);
+        await queryRunner.query(`ALTER TABLE "certificate_of_conduct" ADD CONSTRAINT "FK_05729540b3678f1917321b621bf" FOREIGN KEY ("inspectingScreenerId") REFERENCES "screener"("id") ON DELETE NO ACTION ON UPDATE NO ACTION`);
```
Jonasdoubleyou, line 145: Have you upgraded TypeORM locally? These migrations look strange ...

graphql/certificate_of_conduct/fields.ts
```diff
@@ -0,0 +1,21 @@
...
+
+
+// eslint-disable-next-line camelcase
+@Resolver(of => Certificate_of_conduct)
+export class ExtendedFieldsCertificateOfConductResolver {
+    @FieldResolver(returns => [Student])
+    @Authorized(Role.ADMIN)
+    // eslint-disable-next-line camelcase
+    async student(@Root() certificate: Certificate_of_conduct) {
+        return await prisma.student.findMany({
```
Jonasdoubleyou, line 15: Shouldn't this be `findUnique`?

graphql/certificate_of_conduct/mutations.ts
```diff
@@ -0,0 +1,44 @@
...
+import {getLogger} from "log4js";
+const logger = getLogger("Certificate of Conduct");
+
+
+@Resolver(of => GraphQLModel.Certificate_of_conduct)
+export class MutateCertificateOfConductResolver {
+    @Mutation(returns => Boolean)
+    @Authorized(Role.ADMIN)
+    // eslint-disable-next-line camelcase
+    async updateCertificate(@Arg("studentId") studentId: number,
```
Jonasdoubleyou, line 18: Unfortunately mutations live in a global namespace, thus "updateCertificate" is a bit too generic as we have 3 kinds of certificates (and counting :)) ... Maybe "certificateOfConductUpdate" would be a better name (although quite long) ...

common/migration/1635880160929-AddUpdatedAtInCOC.ts
ayushpandey8439: The version both locally and remote is ~0.29 so i don't think the version is different. I don't remember changing anything else in the schema either. I agree, These do look strange when compared to other migrations. I'll dig deeper and see where all these ALTER statements are coming from. Maybe Postgres versions being different might cause this? I don't know.

graphql/certificate_of_conduct/fields.ts
ayushpandey8439: Logically, it doesnt make a different since both are unique anyways but syntactically yes. I'll update this.

prisma/schema.prisma
```diff
@@ -424,7 +425,7 @@ model screening {
 }
 
 model student {
-  id                                        Int                                    @id @default(autoincrement())
+  id                                        Int                                    @id @default(autoincrement())wix_creation_date
```
Jonasdoubleyou, line 428: I'm not sure how this change got here, however it introduces a syntax error and Prisma fails to load:
```
Error: Get config: Schema Parsing P1012

error: Error validating: This line is not a valid field or attribute definition.
  -->  schema.prisma:428
   | 
427 | model student {
428 |   id                                        Int                                    @id @default(autoincrement())wix_creation_date
429 |   createdAt                                 DateTime                               @default(now()) @db.Timestamp(6)
   | 

Validation Error Count: 1
```

Removing this again and running `npx prisma format` and `npx prisma generate` fixes this and should make the build work again

jobs/periodic/deactivate-missing-coc/index.ts
```diff
@@ -0,0 +1,43 @@
...
+                }
+            },
+            // eslint-disable-next-line camelcase
+            certificate_of_conduct: {
+                none: {}
+            }
+        }
+    });
+
+    logger.info(defaultingStudents.length+ " Defaulting students are ::: "+JSON.stringify(defaultingStudents));
```
Jonasdoubleyou, line 38: This is hard to track back from the logs, maybe Something like "5 students are going to be deactivated because of a missing CoC: "

jobs/periodic/deactivate-missing-coc/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export async function deactivateMissingCoc() {
+    const today = new Date();
+    const eightWeeksAgoDate = new Date();
+    var eightWeeksAgo = today.getDate() - 56;
+    eightWeeksAgoDate.setDate(eightWeeksAgo);
+    const defaultingStudents = await prisma.student.findMany({
+        where: {
+            createdAt: {
+                gt: new Date("2021-11-01")
```
Jonasdoubleyou, line 22: This changed to January, right @lsc0498 ?

graphql/certificate_of_conduct/mutations.ts
```diff
@@ -0,0 +1,44 @@
...
+        }
+        await updateCertificateOfConduct(student, criminalRecord, dateOfIssue, dateOfInspection);
+        return true;
+    }
+
+
+    @Mutation(returns => Boolean)
+    @Authorized(Role.ADMIN)
+    // eslint-disable-next-line camelcase
+    async createCertificate(
```
Jonasdoubleyou, line 35: `certificateOfConductUpdate` ...

common/student/index.ts
```diff
@@ -0,0 +1,19 @@
+import {student as Student} from "@prisma/client";
+import {prisma} from "../prisma";
+
+export async function updateCertificateOfConduct(student : Student, criminalRecord: boolean, dateOfIssue: Date, dateOfInspection: Date) {
+    if (!student.active) {
+        throw new Error("Student was deactivated. Tyring tyo add certificate.");
```
Jonasdoubleyou, line 6: Typo

common/student/activation.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    for (const course of studentCourses) {
+        await prisma.course_instructors_student.deleteMany({
+            where: {
+                studentId: student.id,
+                courseId: course.courseId
+            }
+        });
+        if (studentCourses.length === 1) {
+            await prisma.course.updateMany({
```
Jonasdoubleyou, line 52: Maybe this should be a `cancelCourse` function in `common/course`. As this is now the third place were we cancel courses with a different implmenentation (though your's is the best so far ;)) we might want to harmonize course cancellation at one place. Also if the student directly cancels the course (see https://github.com/corona-school/backend/blob/b59c8ee7476009eb0ad0bc7e6f1f04c28117279b/web/controllers/courseController/index.ts#L1975), we (a) send out emails and (b) insert `CancelSubcourseEvent` and `CancelCourseEvent` into the transaction log. I'm fine though if we leave this as is for the moment and harmonize this in a followup PR.

Jonasdoubleyou (commented):
Apart from the "Kleinkram" mentioned, this PR looks good overall and IMO this would then be ready for merge.

jobs/periodic/deactivate-missing-coc/index.ts
Jonasdoubleyou: As an alternative, maybe we can comment this out (after testing it), so that we can toggle it on when it really should be launched ...

ayushpandey8439: It is basically a matter of setting up the CRON Job. So this is not really being executed at the moment.

Jonasdoubleyou: Ah, true, maybe we can then set the line to `gt: new Date("9999-01-01") // TODO: Set launch date when CoC enforcement is started` then so that setting this is not forgotten ...

common/student/index.ts
```diff
@@ -3,7 +3,7 @@ import {prisma} from "../prisma";
 
 export async function updateCertificateOfConduct(student : Student, criminalRecord: boolean, dateOfIssue: Date, dateOfInspection: Date) {
     if (!student.active) {
-        throw new Error("Student was deactivated. Tyring tyo add certificate.");
+        throw new Error("Student was deactivated. Tyring to add certificate.");
```
lsc0498, line 6: There is still a misspelling ;D
`Tyring` --> `Trying`

PR #233, opened 2021-10-31
Add flag to disable browser initialization and puppeteer

realmayus:
Closes https://github.com/corona-school/project-user/issues/398

This adds support for the `--noPDF` command line argument and adds a new npm script that runs the project in dev mode, but with this new argument: `npm run web:nopdf`

### DIFF
--- a/web/index.ts
+++ b/web/index.ts
@@ -27,9 +27,10 @@ import moment from "moment-timezone";
 import { closeBrowser, setupBrowser } from "html-pppdf";
 import { performCleanupActions } from "../common/util/cleanup";
 import "reflect-metadata"; //leave it here...
-import { apolloServer } from "./../graphql";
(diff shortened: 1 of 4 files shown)


web/index.ts
```diff
@@ -69,9 +74,12 @@ createConnection().then(setupPDFGenerationEnvironment)
         addCorsMiddleware();
         addSecurityMiddleware();
 
-        configureParticipationCertificateAPI();
+
+        if (isCommandArg("--noPDF")) {
```
Jonasdoubleyou, line 78: Shouldn't this be `if (!isCommandArg(...))` ?

realmayus: Indeed, thanks! 😬

Jonasdoubleyou (approved):

PR #236, opened 2021-11-21
GraphQL: Release Student / Pupil entities via match association

Jonasdoubleyou:
Pupils can see the public data of their matched students and vice versa. 

```gql
query { 
	me { 
  	pupil { 
    	matches { 
      	id 
        pupil { 
        	id
               authToken # as this is a self reference, the pupil can also see all their data here
        }
        student { 
               # for the student only public fields can be seen
        	id
               firstname
               lastname
               languages
               subjectsFormatted { name grade { min max }}
        }
      }
    }
  }
}
```


lsc0498 (approved):
Looks good to me👍

realmayus (approved):

PR #237, opened 2021-11-26
GraphQL: Publish (public) Courses to everyone

Jonasdoubleyou:
BLOCKED BY: https://github.com/corona-school/project-user/issues/403

Everyone can now query for public courses (course was screened, not cancelled, submitted) with this query:

```
query { 
  subcoursesPublic(take: 10, onlyJoinable: true, search: "abap") { 
    id
    course { name outline }
    lectures { start duration }
  }
}
```

This also adds full text search through the `search`  parameter, filtering for joinable courses as well as pagination. 
The responses are cached (similar to the existing REST implementation, serving from cache takes about 40ms including network, a miss takes about 200ms, depending on the fields) through a `@PublicCache` annotation to cache public queries. 

Adding the caching plugins revealed several dependency missmatches in the whole GraphQL setup, 
 to resolve this all dependencies were manually pinned to specific working versions. 
Also some breaking changes occured during the upgrade from Apollo v2 to v3, the tracing functionality was simply removed (but this was not useful so far anyways). 

Additionally I finally unified the definition of `isDev` at one place and switched on debug logging in the dev landscape, 
 as debug logging is useless otherwise anyways. 

Proper usage of caching is enforced through a linter:
image


Jonasdoubleyou:
The Apollo - GraphQL - TypegrapQL dependency hell is driving me crazy:

```
UnmetGraphQLPeerDependencyError: Looks like you use an incorrect version of the 'graphql' package: "16.0.0". Please ensure that you have installed a version that meets TypeGraphQL's requirement: "^15.3.0".
  at Object.ensureInstalledCorrectGraphQLPackage (/app/node_modules/type-graphql/dist/utils/graphql-version.js:20:15)
```

web/index.ts
```diff
@@ -30,10 +30,34 @@ import "reflect-metadata"; //leave it here...
 import { apolloServer } from "./../graphql";
 import rateLimit from "express-rate-limit";
 import * as notificationController from "./controllers/notificationController";
+import { isDev } from "../common/util/environment";
+
 
 // Logger setup
 try {
-    configure("web/logconfig.json");
```
lsc0498, line 36: Is there a particular reason for this not being in a separate file anymore @Jonasdoubleyou?

lsc0498 (dismissed):
Good job! Thank you! 👍

web/index.ts
Jonasdoubleyou: Yeah, wanted to change the config dependent on `process.env.ENV`, so that it's enabled in our dev landscape but not on PROD. That's only possible with config "as code".

Jonasdoubleyou: I actually tried to move this to /common/ but then noticed that we have a different logger config for the jobs, so this would require quite some refactoring to align ...

lsc0498 (approved):

PR #239, opened 2021-11-28
AWS S3 for attachments

realmayus:
See https://github.com/corona-school/project-user/issues/402 for more details.


web/controllers/attachmentController/index.ts
```diff
@@ -0,0 +1,60 @@
+import {getLogger} from "log4js";
+import {Request, Response} from "express";
+import {getAttachmentURL} from "../../../common/attachments";
+
+const logger = getLogger();
+
+
+
+/**
+ * @api {GET} /attachments/:attachmentId/:filename getCertificateConfirmation
```
Jonasdoubleyou, line 10: getCertificateConfirmation?

web/controllers/courseController/index.ts
```diff
@@ -2805,23 +2808,67 @@ async function groupMail(student: Student, courseId: number, subcourseId: number
         return 403;
     }
 
+    const lectures = subcourse.lectures.sort(
```
Jonasdoubleyou, line 2811: By storing the return value of sort in a variable it seems as if `subcourse.lectures` is cloned, but in fact it is mutated. I think this would be clearer without the `lectures` local variable.

web/controllers/courseController/index.ts
```diff
@@ -2805,23 +2808,67 @@ async function groupMail(student: Student, courseId: number, subcourseId: number
...
+        (a, b) => a.start.getMilliseconds() - b.start.getMilliseconds()
+    );
+    const lastLecture = lectures[lectures.length - 1];
+    if (lastLecture != null) {
```
Jonasdoubleyou, line 2815: I would either prefer strict equality comparison (`!==`) to `undefined` here, or alternatively just `if (lastLecture)` ... Comparing to `null` makes it seem as if the value could possibly be null ...

web/controllers/courseController/index.ts
```diff
@@ -2903,17 +2950,47 @@ async function instructorMail(pupil: Pupil, courseId: number, subcourseId: numbe
...
+    
 
     try {
-        // send mail to correspondnet
-        await sendParticipantToInstructorMail(pupil, course.correspondent, course, mailSubject, mailBody);
+        let attachmentGroupId = uuid().toString();
+
+        let attachmentURLs = "";
+
+        if (files.length > 0) {
```
Jonasdoubleyou, line 2977: I would prefer if that logic would be moved into some reusable function, e.g. `const attachmentGroup = Notification.createAttachments(files);`

web/controllers/courseController/index.ts
```diff
@@ -2903,17 +2950,47 @@ async function instructorMail(pupil: Pupil, courseId: number, subcourseId: numbe
...
+
+
         await Notification.actionTaken(pupil, "instructor_course_participant_message", {
             participant: pupil,
             course: dropCourseRelations(course),
             subcourse: dropSubcourseRelations(subcourse),
             subject: mailSubject,
-            body: mailBody
+            body: mailBody,
+            attachmentURLs
```
Jonasdoubleyou, line 2993: I would prefer if `Notification.actionTaken` would take in the attachments not just as "another context" but as a dedicated parameter. That way, we can then also store the attachment group ids in the concrete notification, to trace back which file was shared with whom and when (e.g. when we have to delete some of the files manually for some reason, or if we add "garbage collection" somewhen).

web/controllers/courseController/index.ts
Jonasdoubleyou: Actually, as far as I can see a top-1 sort would be sufficient here: `const lastLecture = subcourse.lectures.reduce((a, b) => a.getMilliseconds() > b.getMilliseconds() ? a : b);`

common/attachments/index.ts
```diff
@@ -0,0 +1,44 @@
...
+        }
+    });
+
+    await putFile(file.buffer, `${attachmentGroupId}/${id}/${file.originalname}`, ATTACHMENT_BUCKET);
+
+    return id;
+}
+
+
+export async function getAttachmentURL(id, key, attachmentGroupId?) {
```
Jonasdoubleyou, line 31: Missing types?

common/notification/channels/mailjet.ts
```diff
@@ -23,7 +24,7 @@ export const mailjetChannel: Channel = {
             ],
             TemplateID: notification.mailjetTemplateId,
             TemplateLanguage: true,
-            Variables: context,
+            Variables: {context, ...{attachmentGroup: attachments.attachmentListHTML}},
```
Jonasdoubleyou, line 27: `{ ...context, attachmentGroup: attachments.attachmentListHTML }`

common/notification/index.ts
```diff
@@ -239,3 +247,20 @@ async function deliverNotification(concreteNotification: ConcreteNotification, n
     }
 }
 
+export async function createAttachments(files: Express.Multer.File[], uploader: Student | Pupil): Promise<AttachmentGroup> {
```
Jonasdoubleyou, line 250: Nice! What happens if no files are passed in (empty array)? Wouldit make sense to return `null` in that case? As far as I can see this currently generates a group ID with no corresponding files in the attachments table.

web/controllers/courseController/index.ts
```diff
@@ -2905,15 +2938,32 @@ async function instructorMail(pupil: Pupil, courseId: number, subcourseId: numbe
...
+        let attachmentURLs = await Notification.createAttachments(files, pupil);
+
         await Notification.actionTaken(pupil, "instructor_course_participant_message", {
             participant: pupil,
             course: dropCourseRelations(course),
             subcourse: dropSubcourseRelations(subcourse),
             subject: mailSubject,
-            body: mailBody
+            body: mailBody,
+            attachmentURLs
```
Jonasdoubleyou, line 2966: `}, attachmentGroup)` ... Currently passed into context directly?

Jonasdoubleyou (commented):
I really like the new attachment API. Apart from the minor issues (and proper testing / creation of Notifications) this should be good to go.

common/notification/index.ts
realmayus: Indeed, it would make more sense to just return `null`, especially considered that all the other notification methods take in the attachments as an optional parameter 👍

Jonasdoubleyou (dismissed):
Looks good!
Don't forget to create a Notification and transport that to PROD too ;)

realmayus:
@Jonasdoubleyou https://github.com/corona-school/backend/blob/cfcfe2de46f452f105d8367679839e073d9a0109/common/notification/index.ts#L129
I assume that attachments for reminders should also be implemented, right? If so, I'd also have to find some way of obtaining the `attachmentGroup` object for passing it here, as we only have access to the `attachmentGroupId` in the `concrete_notification` rows of our database.

common/attachments/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+attachmentGroupId: Unique per group of attachments (per message)
+returns: attachmentId: Unique per individual attachment
+ */
+export async function createAttachment(file: Express.Multer.File, uploader: Student | Pupil, attachmentGroupId: string) {
+    let attachmentId = uuid().toString();
+    await prisma.attachment.create({
+        data: {
+            id: attachmentId,
+            uploaderID: uploader.wix_id,
```
Jonasdoubleyou, line 24: Ah, one last thing: Do we want to continue to use the `wix_id`? Do we use that as a user id in other places too?

Jonasdoubleyou:
Yes, it would be good if the Reminder also supports the attachment. And querying the attachment table sounds like a good idea.

common/attachments/index.ts
realmayus: Oh, I was under the impression that we used wix_id, but that's probably because I worked on some older code a few months ago. Changed it to `getUserId(uploader)` 👍

realmayus: (which is what the notification system also uses)

Jonasdoubleyou: Hm, the more I think about this, `getUserId` is not quite optimal as we can't JOIN it on database level (well, we can with some `SUBSTRING(...)` magic. The good thing with the wix_id (apart from it's name) is that it is actually unique across all users, the bad thing is that for retrieving the related user record, one has to query all user tables (student, pupil, screener, mentor). Maybe we should rethink this before shipping this to production (the notification system is currently not used anywhere yet)

Jonasdoubleyou:
@realmayus this has conflicts with master?

common/attachments/index.ts
```diff
@@ -50,4 +61,60 @@ export async function getAttachmentURL(attachmentId: string, key: string, attach
...
+ * @param attachmentGroupId
+ * @return              HTML of the list.
+ */
+export async function getAttachmentListHTML(attachments: { attachmentId: string, filename: string, size: number }[], attachmentGroupId: string) {
+    let attachmentListHTML = "<h3>Anhänge</h3>";
+
+    for (let {attachmentId, filename, size} of attachments) {
+        let key;
+        if(!filename || !size) {  // When sending reminders, for example, we don't have immediate access to data like size and the filename, we only have the attachmentGroupId
+            logger.warn("Fetching data for attachment with ID " + attachmentId);
```
Jonasdoubleyou, line 79: Who is supposed to be warned? Also how can the filename be undefined if the functions signature suggests that it is present?

common/attachments/index.ts
```diff
@@ -50,4 +61,60 @@ export async function getAttachmentURL(attachmentId: string, key: string, attach
...
+                select: {
+                    size: true,
+                    filename: true
+                }
+            })
+            filename = fetched_attachment.filename;
+            key = `${attachmentId}/${filename}`;
+            size = fetched_attachment.size;
+        } else {
+            key = `${attachmentId}/${filename}`;
```
Jonasdoubleyou, line 93: `const key = ...` after the `if(...) { ... }`

common/notification/index.ts
```diff
@@ -39,6 +43,8 @@ export async function actionTaken(user: Person, actionId: string, notificationCo
             if (!relevantNotifications) {
                 logger.debug(`Notification.actionTaken found no notifications for action '${actionId}'`);
                 return;
+            } else {
```
Jonasdoubleyou, line 46: unnecessary `else`

common/notification/index.ts
```diff
@@ -129,14 +134,18 @@ export async function checkReminders() {
                 throw new Error(`Notification(${reminder.id}) was supposed to contain a context when sending out reminders`);
             }
 
+            const attachmentGroup = await getAttachmentGroupByAttachmentGroupId(reminder.attachmentGroupId);
```
Jonasdoubleyou, line 137: What happens here if no `reminder.attachmentsGroupId` is present

common/notification/index.ts
```diff
@@ -135,8 +135,10 @@ export async function checkReminders() {
                 throw new Error(`Notification(${reminder.id}) was supposed to contain a context when sending out reminders`);
             }
 
-            const attachmentGroup = await getAttachmentGroupByAttachmentGroupId(reminder.attachmentGroupId);
-
+            let attachmentGroup = undefined;
+            if (reminder.attachmentGroupId !== undefined) {
```
Jonasdoubleyou, line 138: Does Prisma really return `undefined` and not `null` for NULLable columns? Have you tested this?

realmayus: Oh true, I changed it last minute before pushing 🤦 sorry..

Jonasdoubleyou (dismissed):
Don't forget to add the new ENV variable ;)

Jonasdoubleyou (dismissed):

realmayus:
Just deployed this on [backend-feat-402-w9ycxxoabxz8j/](https://dashboard.heroku.com/apps/backend-feat-402-w9ycxxoabxz8j/) and seems to work fine with our heroku setup. I wonder why the linter is acting up all of a sudden, though, since I didn't touch those files 🤔

web/controllers/courseController/index.ts
```diff
@@ -2773,7 +2773,7 @@ export async function groupMailHandler(req: Request, res: Response) {
             && req.params.subid != undefined
             && typeof req.body.subject == "string"
             && typeof req.body.body == "string") {
-            status = await groupMail(res.locals.user, Number.parseInt(req.params.id, 10), Number.parseInt(req.params.subid, 10), req.body.subject, req.body.body);
+            status = await groupMail(res.locals.user, Number.parseInt(req.params.id, 10), Number.parseInt(req.params.subid, 10), req.body.subject, req.body.body, JSON.parse(req.body.addressees), req.files as Express.Multer.File[]);
```
Jonasdoubleyou, line 2776: Why are the mails passed in here? What happens to the current web user app which does not pass in `req.body.addressees`?

realmayus: `req.body.addressees` aren't mails, they are user IDs. They're in there because this feature allows you to specify the participants you want to send the message to. As for the current web app, it would probably error out, but that's why there's a [frontend PR](https://github.com/corona-school/web-user-app/pull/207) which adds this feature :) We'd have to merge them into master at the same time then.

Jonasdoubleyou (dismissed):

Jonasdoubleyou (dismissed):

lsc0498 (dismissed):
Looks good👍

Jonasdoubleyou (changes requested):

common/notification/channels/mailjet.ts
```diff
@@ -78,7 +80,7 @@ export const mailjetChannel: Channel = {
                 "Content-Type": "application/json"
             },
             method: "POST"
-        }).then(res => res.json());
+        }).then(res => res.json()).catch(ex => logger.error("Couldn't send email", ex));
```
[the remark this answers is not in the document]
realmayus: Oops, I didn't see the `await` and thought the error would go unnoticed. Should be fixed now

Jonasdoubleyou (approved):

common/notification/channels/mailjet.ts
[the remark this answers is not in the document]
Jonasdoubleyou: Erroring here should set the concrete notification state to 'error' and should write the error message into the error column. We should monitor the concrete notifications after changes to the notification system ...

PR #240, opened 2021-12-01
removed node version 13

tillhoffmann1411:
Remove node version 13 from the Docker setup, as the code does not compile under this version. This is because of optional chaining `object.?param` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) is not supported under node version 13. However, this feature is used in some parts of the code.


Jonasdoubleyou:
Yeah, seems I missed the dev container when I've upgraded our build pipelines to NodeJS 14. I guess we still want to keep devcontainer and pipeline on the same version, so maybe this line should stay just with Node 14?

lsc0498 (approved):

Jonasdoubleyou:
abandoned

PR #261, opened 2022-01-02
Updated packages

tillhoffmann1411:
The packages were updated as far as possible and breaking changes were fixed. Furthermore, the package-lock.json was regenerated so that references to other package registries were removed.

I have also formulated a short guide for the Docker setup.


Jonasdoubleyou:
abandoned

PR #266, opened 2022-01-15
Bulk Actions for newly introduced Notifications

Jonasdoubleyou:
[SQL Script for testing](https://github.com/corona-school/dev-db/blob/main/dev/SimulateMatche.pgsql)

Seems to work quite fine, I'll bulk test on DEV with some thousand matches.

### DIFF
--- a/common/notification/index.ts
+++ b/common/notification/index.ts
@@ -260,6 +260,8 @@ async function deliverNotification(concreteNotification: ConcreteNotification, n
 /* When introducing new notifications, it might sometimes make sense to schedule them "in retrospective" once for all existing users,
    as if the user took that action at actionDate */
 export async function actionTakenAt(actionDate: Date, user: Person, actionId: string, notificationContext: NotificationContext, apply: boolean) {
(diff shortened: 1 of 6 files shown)


common/notification/index.ts
```diff
@@ -241,3 +257,56 @@ async function deliverNotification(concreteNotification: ConcreteNotification, n
     }
 }
 
+/* When introducing new notifications, it might sometimes make sense to schedule them "in retrospective" once for all existing users,
+   as if the user took that action at actionDate */
+export async function actionTakenAt(actionDate: Date, user: Person, actionId: string, notificationContext: NotificationContext, apply: boolean) {
```
lsc0498, line 262: Are we sure that we will never send mails to deactivated users?

Jonasdoubleyou: Very true, that check was missing. Good catch!

Jonasdoubleyou: Though at least in theory all matches of deactivated users should be dissolved (just noticed when trying to test this)

lsc0498 (approved):
Looks good to me. Thank you!

PR #273, opened 2022-02-07
Fix ID of created notifications

Jonasdoubleyou:


common/notification/notification.ts
```diff
@@ -115,8 +115,14 @@ export async function create(notification: Prisma.notificationCreateInput) {
...
     }
 
+    // To keep DEV and PROD parity, notifications are inserted with their id (see "importNotifications")
+    // Unfortunately this creates inconsistency between the ids present and the sequence used to generate new ids
+    // Thus the id is manually calculated and does not rely on the DEFAULT value
     const result = await prisma.notification.create({
-        data: { ...notification }
+        data: {
+            ...notification,
+            id: await prisma.notification.count() + 1
```
lsc0498, line 124: Can't the `id` be set to `undefined` or something like that to use the autoincrement from the database?🤔

Jonasdoubleyou: Yeah that's the point: The autoincrement from the database is wrong. For sure there are other solutions to this (e.g. use a UUID instead of an incrementing ID) but this was the most minimal fix.

lsc0498: I see... I thought the reason for the problem was that `...notification` includes an `id` field which overrides the autoincrement.

lsc0498 (approved):

PR #279, opened 2022-02-25
Add: createCourse endpoint

Leo310:


graphql/course/mutations.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  @TypeGraphQL.Field(_type => course_category_enum, {
+      nullable: false
+  })
+  category!: "revision" | "club" | "coaching";
+
+  @TypeGraphQL.Field(_type => Boolean, {
+      nullable: true
+  })
+  allowContact?: boolean | undefined;
```
Jonasdoubleyou, line 41: this could be non-nullable. Also the other `nullable: false` are obsolete, they're the default value ... Maybe this could be adressed in a followup change

Jonasdoubleyou (approved):

PR #282, opened 2022-03-20
Update and pin Prisma versions

Jonasdoubleyou:
This should fix version missmatches when running `npx prisma generate`. 
Also prisma versions are pinned, this should prevent missmatches in the future. 

typegraphql-prisma 0.14.6 -> 0.19.2 ([changelog](https://github.com/MichalLytek/typegraphql-prisma/releases)):
 - breaking changes: none according to changelog? (kind of, a new property is added to a Required<..> type in authorizations.ts, so not breaking, but requires changes on our end)
 - useful features: "omit" -> maybe we can stop generating all mutations (as we don't use many of them)

prisma: 2.30.3 -> 3.11.0 ([changelog](https://github.com/prisma/prisma/releases)):
- breaking changes:
  - JSON null handling (not used by us?)
  - aggregations are now prefixed with underscore (`count` -> `_count`)
  - [naming constraints](https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-3/named-constraints)
- useful features:
  - CLI improvements, e.g. `prisma migrate diff` to diff schemas
  - [full text search and order by relevance](https://www.prisma.io/docs/concepts/components/prisma-client/filtering-and-sorting#sort-by-relevance-postgresql)
  - referential actions (restrict deletion of associated records on DB level)
  - [Seeding](https://www.prisma.io/docs/guides/database/seed-database), basically what we do in dev.ts, just nicer (with diffs between versions)
  - order by aggregate
  - Node-API (the Prisma engine now runs as part of the Node process which should be way more efficient)

Have fun reviewing the 1.876 changed files ;) (only package.json and authorizations.ts were manually changed, the rest is autogenerated ...)


realmayus (approved):

lsc0498 (approved):
Looks good! Sorry for the delay.

PR #287, opened 2022-04-02
Cancel Notifications via GraphQL

Jonasdoubleyou:
- With `concreteNotificationCancel(id: 1)` Admins can cancel pending reminders (to clean up faulty backend logic / users complaining about reminders)
- Fixed CoC cancellation to exclude users who revoked their tutoring state after screening
- Added resolver from concrete_notifications to notifications to easier query user notifications


graphql/concrete_notification/fields.ts
```diff
@@ -0,0 +1,13 @@
+import { Concrete_notification as ConcreteNotification, Notification } from "../generated";
+import { Authorized, FieldResolver, Resolver, Root } from "type-graphql";
+import { prisma } from "../../common/prisma";
+import { Role } from "../authorizations";
+
+@Resolver(of => ConcreteNotification)
+export class ExtendedFieldsConcreteNotificationResolver {
+    @FieldResolver(returns => Notification)
+    @Authorized(Role.UNAUTHENTICATED)
```
lsc0498, line 9: Which privileges does this `UNAUTHENTICATED` role have? Why isn't this constrained to admins?

lsc0498 (approved):
Looks good!

graphql/concrete_notification/fields.ts
Jonasdoubleyou: Everyone has the Unauthenticated role, it basically disables authentication (explicitly). Cause everyone who can access a certain concrete_notification should be able to follow the edge to the corresponding notification. Also notifications are "public configuration" (description, interval, etc.) so there's no need to enforce strict authentication there.

PR #292, opened 2022-04-16
Don't send match followup mails for LF Plus

Jonasdoubleyou:


Jonasdoubleyou:
I'll verify whether this works as intended on DEV, as the review apps are currently broken

lsc0498 (approved):
Thanks a lot!

PR #296, opened 2022-04-23
Match Pool toggling

Jonasdoubleyou:
Provides way to influence the behavior for Matching, with a flexible interface for different match pools


Jonasdoubleyou:
(force merged as we have some urgency here, feel free to review post-merge)

PR #297, opened 2022-04-24
Update: Nullable types and return Course on edit instead of boolean

Leo310:
The follow up change you (@Jonasdoubleyou ) proposed. And I thought it would be nice to also return the course on the edit mutation (for caching?).

### DIFF
--- a/graphql/course/mutations.ts
+++ b/graphql/course/mutations.ts
@@ -37,13 +37,19 @@ class PublicCourseCreateInput {
 }
 
 class PublicCourseEditInput {
-  @TypeGraphQL.Field(_type => String)
+  @TypeGraphQL.Field(_type => String, {
+      nullable: true
+  })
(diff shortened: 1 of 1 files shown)


graphql/course/mutations.ts
```diff
@@ -10,34 +10,24 @@ import { GraphQLContext } from "../context";
...
-      nullable: false
-  })
+  @TypeGraphQL.Field(_type => course_category_enum)
   category!: "revision" | "club" | "coaching";
 
-  @TypeGraphQL.Field(_type => Boolean, {
-      nullable: true
-  })
+  @TypeGraphQL.Field(_type => Boolean)
   allowContact?: boolean | undefined;
```
Jonasdoubleyou, line 41: allowContact: boolean

Jonasdoubleyou (approved):

PR #308, opened 2022-05-26
Add lecture create and delete enpoints

Leo310:


graphql/subcourse/mutations.ts
```diff
@@ -84,11 +85,41 @@ export class MutateSubcourseResolver {
...
+        }
+        const result = await prisma.lecture.create({ data: { ...lecture, subcourseId }});
+        return result;
+    }
+
+    @Mutation((returns) => Boolean)
+    @AuthorizedDeferred(Role.ADMIN, Role.OWNER)
+    async lectureDelete(@Ctx() context: GraphQLContext, @Arg("lectureId") lectureId: number): Promise<Boolean> {
+        const lecture = await getLecture(lectureId);
+        const subcourse = await getSubcourse(lecture.subcourseId);
```
Jonasdoubleyou, line 119: I think we should also prevent the deletion of lectures if (a) they are already in the past and (b) if the subcourse is already published (in that case we might have to inform the participants that the lecture was rescheduled, but I guess we can simply forbid this for now and implement once this becomes an issue)

Jonasdoubleyou (approved):

PR #309, opened 2022-05-26
[Cleanup] Remove old matching

Jonasdoubleyou:


common/match/pool.ts
```diff
@@ -144,7 +144,7 @@ export const pools: MatchPool[] = [
             subjects: { not: "[]" },
             screening: { success: true },
             isCodu: false,
```
lsc0498, line 146: The `isCodu` flag should also removed here. In fact we never used the the CoDu-`registrationSource` for CoDu-Students since they could register for the research project even after registration (as distinguished from pupils).

lsc0498 (changes requested):

lsc0498 (approved):
Great!🎉

Jonasdoubleyou:
that was fast :sweat_smile: @lsc0498

PR #311, opened 2022-05-26
[Draft] GraphQL Integration Tests

Jonasdoubleyou:
IMO this adds way more coverage to our codebase as our implementations are rather "along workflows", thus creating unit tests is usually hard and adds little coverage. The integration tests cover less logic, however they at least ensure that nothing breaks in serious ways. 

I'm not quite sure how to integrate this into CI/CD as running those tests without a database / uplink is difficult. 

resolves https://github.com/corona-school/project-user/issues/447

### DIFF
--- a/common/secret/token.ts
+++ b/common/secret/token.ts
@@ -28,7 +28,7 @@ export async function createToken(user: User): Promise<string> {
             userId: user.userID,
             secret: hash,
             expiresAt: null,
-            lastUsed: new Date()
+            lastUsed: null
         }
     });
 
--- a/integration-tests/auth.ts
+++ b/integration-tests/auth.ts
@@ -1,0 +1,64 @@
+import { test } from "./base";
+import { pupilOne } from "./user";
+import * as assert from "assert";
+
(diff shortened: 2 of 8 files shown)


Jonasdoubleyou:
@lsc0498 @Leo310 what do you think of this (and having integration tests in general)?

Leo310:
I think integration tests are a good idea even if it adds more work. The testing framework you made looks quite nice. Maybe a way to inject results of previous queries (e.g. pupilId) into the next one, could be via a context variable that gets passed to each runner (I need to look up what comment binding variables are).

And why without a database? Cant we just deploy one in the CI/CD process? Or use sqlite or something that can be deployed in the application itself?

lsc0498 (approved):
Looks good! I think the best would be to integrate this to our GitHub commit checks.

lsc0498:
I think we already have some kind of testing database 🤔

Jonasdoubleyou:
Ah, true, actually I overlooked that. Indeed we do start a Postgres with every CI run

PR #315, opened 2022-05-31
Added prettier

Leo310:
Added small default .prettierrc config. I have run my formatter on subcourse/mutations and course/mutation so that you can see what effect prettier has.


Leo310:
If there is nothing more to discuss, would be cool if we could merge this. Would make formatting much easier.

Jonasdoubleyou (approved):

Leo310:
What should I do with a conflicting package-lock.json?

Jonasdoubleyou:
We could also add `npm run prettier` to a precommit hook

Jonasdoubleyou (approved):

PR #318, opened 2022-07-01
Reduce Dependencies

realmayus:
This PR aims to clean up a few dependencies listed in https://github.com/corona-school/project-user/issues/449.

* Bumped node version to 18.x to support native `fetch`
* Implemented custom mailjet solution to avoid having to ship the SDK
* Migrated the S3 SDK to v3 which is modular, resulting in a smaller dependency size

### DIFF
--- a/common/mails/mailjet.ts
+++ b/common/mails/mailjet.ts
@@ -3,7 +3,7 @@ import { getLogger } from "log4js";
 import * as mailjet from "./mailjetTypes";
 
 const logger = getLogger();
-const auth = btoa(`${mailjetSmtp.auth.user}:${mailjetSmtp.auth.pass}`);
+const auth = Buffer.from(`${mailjetSmtp.auth.user}:${mailjetSmtp.auth.pass}`).toString("base64");
 
 async function sendMessage(message: mailjet.SendParamsMessage, sandbox: boolean = false) {
     //determine whether we have sandbox mode or not...
--- a/common/mails/mailjet.ts
+++ b/common/mails/mailjet.ts
@@ -1,10 +1,11 @@
 import { mailjetSmtp } from "./config";
-import * as mailjetAPI from "node-mailjet";
 import { getLogger } from "log4js";
+import * as mailjet from "./mailjetTypes";
 
 const logger = getLogger();
+const auth = btoa(`${mailjetSmtp.auth.user}:${mailjetSmtp.auth.pass}`);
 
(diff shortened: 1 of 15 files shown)


common/mails/mailjet.ts
```diff
@@ -1,10 +1,11 @@
 import { mailjetSmtp } from "./config";
-import * as mailjetAPI from "node-mailjet";
 import { getLogger } from "log4js";
+import * as mailjet from "./mailjetTypes";
 
 const logger = getLogger();
+const auth = btoa(`${mailjetSmtp.auth.user}:${mailjetSmtp.auth.pass}`);
```
Jonasdoubleyou, line 6: `Buffer.from(...).toString('base64')` ...

Jonasdoubleyou:
Um why should I review this when this doesn't build? There's another `btoa` ....

realmayus:
Oops, forgot to run the code 😅

Jonasdoubleyou (dismissed):
Have you tested the mail sending? We currently send out a few hundred mails each day, so if that fails it gets us in some trouble ...

realmayus:
Just tested the email sending again, and it works both on my machine and on Heroku. I tested both directly sent emails and notifications, so it *should* work.

web/controllers/expertController/index.ts
```diff
@@ -265,7 +265,7 @@ async function putExpert(wixId: string, student: Student, info: ApiPutExpert): P
 async function GetExpertiseTagEntities(tagNames: string[]): Promise<ExpertiseTag[]> {
     const entityManager = getManager();
 
-    const tags: ExpertiseTag[] = await entityManager.find(ExpertiseTag, { where: tagNames.map(t => ({ name: t }))});
+    const tags: ExpertiseTag[] = await entityManager.find(ExpertiseTag, { where: { name: In(tagNames)}});
```
realmayus, line 268: This suddenly wouldn't compile anymore.... Maybe it has to do with the updated Node version? 
The In() is equivalent to the tagNames.map() function in this case, right?

realmayus: Concrete error:
```
Error:(268, 11) TS2322: Type '{ name: string; }[]' is not assignable to type 'ExpertiseTag[]'.
  Type '{ name: string; }' is missing the following properties from type 'ExpertiseTag': id, expertData
 ```

Jonasdoubleyou: no repro. Probably something wrong with your build setup

Jonasdoubleyou (approved):

PR #336, opened 2022-09-06
Redact inactive accounts

realmayus:
Closes https://github.com/corona-school/project-user/issues/466.

Due to statistical reasons, accounts aren't deleted; they're redacted.


Jonasdoubleyou:
I think we would also need to 
- drop the concrete notification context of all notifications of the user (they should be cancelled anyways with account deactivation)
- delete the transaction log as that might also contain personal data
- What about the other fields such as university? Do they fulfill kAnonymity?
- attachments?
- signatures in certificates?

jobs/periodic/drop-old-notification-contexts/index.ts
```diff
@@ -0,0 +1,24 @@
...
+import moment from "moment";
+import {getLogger} from "log4js";
+
+const DELETE_AFTER = 30 * 3; // in days
+const logger = getLogger();
+
+export default async function execute() {
+
+    const res = await prisma.concrete_notification.updateMany({
+        where: {
```
Jonasdoubleyou, line 11: status: NotificationState.SENT, should be the case for all notifications but just want to be sure that no pending notification gets redacted accidentially before its sent out (which will then probably crash Mailjets template processing)

jobs/periodic/drop-old-notification-contexts/index.ts
```diff
@@ -0,0 +1,24 @@
...
+            sentAt: {
+                lte: moment().startOf("day")
+                    .subtract(DELETE_AFTER, "days")
+                    .toDate()
+            }
+        },
+        data: {
+            context: {}
```
Jonasdoubleyou, line 19: If we want to show all notifications to the user in some UI, it might be a good idea to somehow be able to distinguish redacted notifications, maybe we could introduce a new notification state ARCHIVED? In theory an empty notification context could also be valid for a sent notification (though it usually always includes the user)

Jonasdoubleyou (approved):

Jonasdoubleyou:
Can we bring this in (potentially deactivating the job if it is not ready yet)? The changes to the notification system block me. @realmayus

Jonasdoubleyou (approved):

PR #337, opened 2022-09-07
Add: pupil isParticipant and isOnWaitingList resolver

Leo310:
Implemented these 2 resolvers. The isOnWaitingList resolver is not working atm, because there is no such field in the schema. Where and how should I add this field?


graphql/pupil/fields.ts
```diff
@@ -110,4 +110,18 @@ export class ExtendFieldsPupilResolver {
     async canJoinSubcourses(@Root() pupil: Required<Pupil>) {
         return canJoinSubcourses(pupil);
     }
-}
\ No newline at end of file
+
+    @FieldResolver(returns => [Boolean])
+    @Authorized(Role.ADMIN)
+    async isParticipant(@Root() pupil: Required<Pupil>, @Arg("subcourseId") subcourseId: number) {
```
Jonasdoubleyou, line 116: I'd prefer if the resolver would be on the Subcourse, and would by default take the session pupil (and support the pupilId as an optional parameter)

graphql/pupil/fields.ts
```diff
@@ -110,4 +110,18 @@ export class ExtendFieldsPupilResolver {
...
+        return await prisma.subcourse_participants_pupil.count({ where: { pupilId: pupil.id, subcourseId }}) > 0 ? true : false;
```
Jonasdoubleyou, line 117: ? true : false is useless ;)

graphql/pupil/fields.ts
Leo310: ahh ok that was kind of my first approach but then I saw the isParticipant field in pupil and thought that a fieldResolver should resolve that field haha, because in subcourse is no such field

graphql/pupil/fields.ts
Leo310: you sure? I think i tested it and it didnt work when returning the count. I think something with "iterable" was the error. But I think I will try again

Jonasdoubleyou: Yeah but the result of the comparison with zero (>0) is already a boolean :)

graphql/pupil/fields.ts
Jonasdoubleyou: Yeah pupil.isParticipant means the user can generally participate in courses and pupil.isPupil means the pupil can take one-to-one tutoring ... Maybe we should somewhen rename those ...

graphql/pupil/fields.ts
Leo310: lol I am stupid hahaha

graphql/pupil/fields.ts
Leo310: Ok I see, but what field in subcourses should be resolved then? Or should it be a query

Jonasdoubleyou: No, a field resolver, subcourse.isParticipant(subcourseId?)

Leo310: Ok then maybe I am misunderstanding something because I thought a field resolver would resolve a field in the schema so making a field resolver in subcourse would require a field "isParticipant" in the subcourse entity. But there is no such field. I will take a look at it again tomorrow

Leo310: Ok, but there also is no field in the Graphql entity and I get the following error when trying to resolve that field with this resolver:
image
Error:
image

So should I create a new field or am I still getting something wrong.

Jonasdoubleyou: @Return(_type => Boolean) ...

Leo310: ahhh damn I didnt see that the entire time. Dont know why I added brackets there. Maybe something left over from copy pasting

Jonasdoubleyou (commented):
Looks good but crashes on start?

Leo310:
hmm weird. I think i messed something up when rebasing master into this branch to get the prettier changes. But I am not sure what the problem is atm

graphql/subcourse/fields.ts
```diff
@@ -1,11 +1,14 @@
...
+import { Arg, Authorized, Ctx, FieldResolver, Query, Resolver, Root } from "type-graphql";
+import { Prisma } from "@prisma/client";
+import { prisma } from "../../common/prisma";
+import { Role } from "../authorizations";
+import { LimitedQuery, LimitEstimated } from "../complexity";
+import { CourseState } from "../../common/entity/Course";
+import { PublicCache } from "../cache";
+import { getSessionPupil } from "../authentication";
+import { GraphQLContext } from "../context";
+import { BBBMeeting } from "../../common/entity/BBBMeeting";
```
Jonasdoubleyou, line 11: Yeah the import here was changed, and is wrong

Jonasdoubleyou (approved):

graphql/subcourse/fields.ts
Leo310: Ah ok I see ty

PR #345, opened 2022-09-15
Add: isInstructor fields and update subcourse.participants

Leo310:
Added FieldResolvers for Instructors: `course.isInstructor`, `subcourse.isInstructor`, `lecture.isInstructor`

### DIFF
--- a/graphql/lecture/fields.ts
+++ b/graphql/lecture/fields.ts
@@ -1,0 +1,16 @@
+import { Role } from '../authorizations';
+import { Arg, Authorized, Ctx, FieldResolver, Query, Resolver, Root } from 'type-graphql';
+import { Course, Lecture, Subcourse, Pupil, Bbb_meeting as BBBMeeting } from '../generated';
+import { GraphQLContext } from '../context';
+import { getSessionStudent } from '../authentication';
+import { prisma } from '../../common/prisma';
+
+@Resolver((of) => Lecture)
+export class ExtendedFieldsLectureResolver {
+    @FieldResolver((returns) => Boolean)
+    @Authorized(Role.ADMIN, Role.STUDENT)
+    async isInstructor(@Ctx() context: GraphQLContext, @Root() subcourse: Subcourse, @Arg('studentId', { nullable: true }) studentId: number) {
+        const student = await getSessionStudent(context, studentId);
+        return (await prisma.lecture.count({ where: { instructorId: student.id } })) > 0;
+    }
+}
--- a/graphql/lecture/fields.ts
+++ b/graphql/lecture/fields.ts
@@ -1,16 +1,15 @@
 import { Role } from '../authorizations';
-import { Arg, Authorized, Ctx, FieldResolver, Query, Resolver, Root } from 'type-graphql';
-import { Course, Lecture, Subcourse, Pupil, Bbb_meeting as BBBMeeting } from '../generated';
+import { Arg, Authorized, Ctx, FieldResolver, Resolver, Root } from 'type-graphql';
+import { Lecture } from '../generated';
 import { GraphQLContext } from '../context';
 import { getSessionStudent } from '../authentication';
-import { prisma } from '../../common/prisma';
 
 @Resolver((of) => Lecture)
 export class ExtendedFieldsLectureResolver {
     @FieldResolver((returns) => Boolean)
     @Authorized(Role.ADMIN, Role.STUDENT)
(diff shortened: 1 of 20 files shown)


graphql/lecture/fields.ts
```diff
@@ -0,0 +1,16 @@
...
+import { getSessionStudent } from '../authentication';
+import { prisma } from '../../common/prisma';
+
+@Resolver((of) => Lecture)
+export class ExtendedFieldsLectureResolver {
+    @FieldResolver((returns) => Boolean)
+    @Authorized(Role.ADMIN, Role.STUDENT)
+    async isInstructor(@Ctx() context: GraphQLContext, @Root() subcourse: Subcourse, @Arg('studentId', { nullable: true }) studentId: number) {
+        const student = await getSessionStudent(context, studentId);
+        return (await prisma.lecture.count({ where: { instructorId: student.id } })) > 0;
```
Jonasdoubleyou, line 14: Missing lecture id, checking lecture.instructorId = student.id should be enough (and the resolver is slightly useless then, although maybe good for consistency, haven't thought this through...)

Jonasdoubleyou: Also Root is not a Subcourse?

Leo310:
is this what you meant with (double check that authorizations on the pupil entity are correct and the instructor can only see the name and grade of the participant, no other personal information!)? @Jonasdoubleyou

graphql/subcourse/fields.ts
```diff
@@ -79,9 +79,21 @@ export class ExtendedFieldsSubcourseResolver {
     }
 
     @FieldResolver((returns) => [Pupil])
-    @Authorized(Role.ADMIN)
+    @Authorized(Role.ADMIN, Role.INSTRUCTOR)
```
Jonasdoubleyou, line 82: Role.OWNER (only for own courses instructors should see the participants)

graphql/subcourse/fields.ts
```diff
@@ -79,9 +79,21 @@ export class ExtendedFieldsSubcourseResolver {
...
     @LimitEstimated(100)
-    async participants(@Root() subcourse: Subcourse) {
+    async participants(@Ctx() context: GraphQLContext, @Root() subcourse: Subcourse) {
+        if (!isElevated(context)) {
+            return await prisma.pupil.findMany({
+                select: { firstname: true, lastname: true, grade: true },
```
Jonasdoubleyou, line 87: I initially thought about something like Role.IS_INSTRUCTOR_OF analogous to Role.OWNER, but the more I think about it that would be quite complex to implement (securely). I actually like the idea of only returning the fields here, as it can't go wrong. I would however like to distinguish the resolvers then, as `courses { participants { email }}` might now suddenly return null values (and further field resolvers might fail in weird ways). I would prefer if we have two resolvers instead, `participantsAsPupil` which is only accessible to Role.ADMIN and which returns regular pupils, and `participants` which is accessible to Role.OWNER, and which only returns the (firstname, lastname, grade). For the `participants` resolver I guess we should then also introduce a new type (`@ObjectType() class Participant { ... }`) so that the participants resolver only has the fields that are actually available.

Jonasdoubleyou: And I guess the participants resolver should also return the pupil id, so that the instructor can then call mutations for specific participants (i.e. send them a message or the participation certificate) ... Sorry for coming up with all this now, just didn't think this through when writing the issue ...

Leo310: No problemo, have implemented it now. Couldnt test the participant resolver for now, because I am not sure how to preserve the login token from legacyLogin in Apollo explorer. But is this what you imagined?

Jonasdoubleyou: Yes, very nice. Though the diff looks weird, maybe merge master into the branch before squash merging (not sure what would happen when squash merging in the current state)...

Jonasdoubleyou (approved):

graphql/subcourse/fields.ts
Leo310: Ok I already merged (rebased) master into the branch, but maybe not entirely? I will take a look at it again later

Jonasdoubleyou: Ah, seems this is by design (https://stackoverflow.com/questions/63603247/github-showing-an-incorrect-diff-between-main-branch-and-pr-branch), so merging (git merge master) works better than rebasing with Github PRs. Anyways, merging fixed it and the diff is now correct :)

Leo310: Ah ok I see :+1:

PR #356, opened 2022-10-03
Added subject and schooltype fields to courses

realmayus:
See https://github.com/corona-school/project-user/issues/467.


lsc0498 (approved):
Looks good to me. It would be nice to have subjects as an enum in the future though 😅

PR #357, opened 2022-10-04
Add Enum for subjects

realmayus:


graphql/course/mutations.ts
```diff
@@ -26,8 +28,9 @@ class PublicCourseCreateInput {
     category!: 'revision' | 'club' | 'coaching';
     @TypeGraphQL.Field((_type) => Boolean)
     allowContact?: boolean;
-    @TypeGraphQL.Field((_type) => String)
-    subject?: string;
+    @TypeGraphQL.Field((_type) => course_subject_enum)
+    subject?: course_subject_enum;
```
Jonasdoubleyou, line 32: Should probably also be editable (PublicCourseEditInput), right?

Jonasdoubleyou (approved):

graphql/course/mutations.ts
Jonasdoubleyou: Oh, and maybe we can also provide filtering options in the publicCourses Query ... (both things which can be done in a followup PR)

PR #359, opened 2022-10-08
Invalidate Interest Confirmations instead of deleting

Jonasdoubleyou:
Also partially switch to the new interest confirmation job.
As we currently have a rather large backlog of pupils, we can disable interest confirmations for now and reenable the job after the backlog was matched, then we can verify that the new interest confirmation logic works as intended.

NOTE: We also need new templates for the new interest confirmations.


lsc0498 (approved):
Looks good to me!👍

PR #397, opened 2022-11-14
 #LF-6 add dummy notifications resolver

jeffersondarcy:
adds a new resolver for user's (concrete) notifications
for testing locally run "npm run build" first, than you can try new query in graphql playground at http://localhost:5000/apollo
an example query would be 
```
#mutation{loginLegacy(authToken: "authtokenP1")} #uncomment and run this line first for auth
{me{concreteNotifications{id,headline,body,createdAt,notification{messageType}}}}
```

### DIFF
--- a/graphql/concrete_notification/dummy_data.ts
+++ b/graphql/concrete_notification/dummy_data.ts
@@ -1,0 +1,20 @@
+import { ConcreteNotification, ConcreteNotificationState } from '../../common/entity/ConcreteNotification';
+
+const ids = Array.from({ length: 10 }, (_, i) => i + 1);
+const dummyMinutes = [0, 5, 10, 15, 20, 30, 60, 120, 150, 60 * 24, 60 * 48];
+
+// Dummy dates indexed by dummy ids
+export const getDummyCreatedAt = (id: number): Date => new Date(Date.now() - dummyMinutes[id] * 1000 * 60);
+
+export const getDummyConcreteNotifications = (userId): ConcreteNotification[] =>
+    ids.map(
+        (id): ConcreteNotification => ({
+            id,
+            attachmentGroupId: '',
+            context: {},
+            notificationID: id,
+            sentAt: new Date(),
+            userId,
+            state: ConcreteNotificationState.SENT,
+        })
+    );
--- a/graphql/concrete_notification/fields.ts
+++ b/graphql/concrete_notification/fields.ts
@@ -4,6 +4,7 @@ import { prisma } from '../../common/prisma';
 import { Role } from '../authorizations';
 import { JSONResolver } from 'graphql-scalars';
 import { ConcreteNotificationState } from '../../common/entity/ConcreteNotification';
+import { getDummyCreatedAt } from './dummy_data';
 
 @ObjectType()
 class Campaign {
@@ -34,6 +35,25 @@ export class ExtendedFieldsConcreteNotificationResolver {
     @Authorized(Role.UNAUTHENTICATED)
     async notification(@Root() concreteNotification: ConcreteNotification) {
         return await prisma.notification.findUnique({ where: { id: concreteNotification.notificationID } });
+    }
+
+    @FieldResolver((returns) => String)
+    @Authorized(Role.OWNER, Role.ADMIN)
+    async headline(@Root() concreteNotification: ConcreteNotification) {
+        return `Mock Headline ${concreteNotification.id}`;
+    }
+
+    @FieldResolver((returns) => String)
+    @Authorized(Role.OWNER, Role.ADMIN)
+    async body() {
+        return `Mock Body Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut`;
+    }
+
+    // @TODO: this should be a field of ConcreteNotification
+    @FieldResolver((returns) => String)
+    @Authorized(Role.OWNER, Role.ADMIN)
+    async createdAt(@Root() concreteNotification: ConcreteNotification) {
+        return getDummyCreatedAt(concreteNotification.id);
     }
 
     @Query((returns) => [Campaign])
--- a/graphql/concrete_notification/fields.ts
+++ b/graphql/concrete_notification/fields.ts
@@ -50,7 +50,7 @@ export class ExtendedFieldsConcreteNotificationResolver {
     }
 
     // @TODO: this should be a field of ConcreteNotification
-    @FieldResolver((returns) => String)
+    @FieldResolver((returns) => Date)
     @Authorized(Role.OWNER, Role.ADMIN)
     async createdAt(@Root() concreteNotification: ConcreteNotification) {
         return getDummyCreatedAt(concreteNotification.id);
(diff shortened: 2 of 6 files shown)


graphql/concrete_notification/fields.ts
```diff
@@ -99,3 +101,53 @@ export class ExtendedFieldsConcreteNotificationResolver {
...
+class ConcreteNotificationFormatted {
+    @Field((type) => ID)
+    id: number;
+    @Field()
+    headline: string;
+    @Field()
+    body: string;
+    @Field()
+    notificationClass: string;
+    @Field() // is there a better solution than "string" as type?
```
Jonasdoubleyou, line 115: Date - which is transported as an ISO string and then automatically serialized / deserialized by Apollo Client / Prisma

graphql/concrete_notification/fields.ts
```diff
@@ -99,3 +101,53 @@ export class ExtendedFieldsConcreteNotificationResolver {
         return Object.values(byCampaign);
     }
 }
+
+@ObjectType()
+class ConcreteNotificationFormatted {
```
Jonasdoubleyou, line 106: `extends ConcreteNotification`, if the type is only used as an "output" (as the result of a Resolver) then adding FieldResolvers for the respective types would be enough (i.e. `FieldResolver(returns => String) headline(@Root() concreteNotification: ConcreteNotification) { ... }`. `@ObjectType`s make sense for inputs (types of arguments), or when returning an entity which is always accessed as a whole (thus there is no need for dedicated field resolvers). For the concrete notification, this is not really the case as not everyone wants to "render" concrete notifications (especially the body, it might also make sense to just query the headline in some cases)

graphql/concrete_notification/fields.ts
```diff
@@ -99,3 +101,53 @@ export class ExtendedFieldsConcreteNotificationResolver {
...
+    @Field((type) => ID)
+    id: number;
+    @Field()
+    headline: string;
+    @Field()
+    body: string;
+    @Field()
+    notificationClass: string;
```
Jonasdoubleyou, line 114: What is a notification class? And why is it on the concrete notification and not the notification?

graphql/concrete_notification/fields.ts
```diff
@@ -99,3 +101,53 @@ export class ExtendedFieldsConcreteNotificationResolver {
...
+@ObjectType()
+class NotificationsList {
+    @Field((type) => [ConcreteNotificationFormatted])
+    notifications: ConcreteNotificationFormatted[];
+    @Field()
+    user: string; // for debugging, @TODO: remove later
+}
+
+const dummyMinutes = [0, 5, 10, 15, 20, 30, 60, 120, 150, 60 * 24, 60 * 48];
+const getDummyNotifications = () =>
```
Jonasdoubleyou, line 128: If test data is needed, this should be directly filled into the database (table concrete_notification / notification) in web/dev.ts.

graphql/concrete_notification/fields.ts
```diff
@@ -99,3 +101,53 @@ export class ExtendedFieldsConcreteNotificationResolver {
...
+            notificationClass: `NotificationClass${key}`,
+            createdAt: `${Date.now() - dummyMinutes[key] * 1000 * 60}`,
+        })
+    );
+
+@Resolver()
+export class UserNotificationsListResolver {
+    @Query((returns) => NotificationsList)
+    // @Authorized(Role.USER)
+    @Authorized(Role.UNAUTHENTICATED) // for debugging, // @TODO: replace with Role.USER
```
Jonasdoubleyou, line 143: Role.ADMIN, Role.OWNER on the field resolver of the user entity

Jonasdoubleyou (changes requested):

graphql/concrete_notification/fields.ts
jeffersondarcy: this should reflect notification type, like "reminder", "match" ... which would be mapped to different icons in frontend. I thought "notificationClass" might be good name, but I am open to all suggestions

graphql/concrete_notification/fields.ts
jeffersondarcy: This ticket [LF-6](https://typedigital.atlassian.net/browse/LF-6?atlOrigin=eyJpIjoiOWI2MGNiYTcxMDIyNDExZjk1NzI4MTJmOWMwN2FkN2UiLCJwIjoiaiJ9) is an intermediate step with the goal of display something in frontend and establishing the connection to graphql. Next ticket will be more refined and we would have more clarity on what data is needed. I would suggest to move the test data to the database with the next ticket (probably [LF-7](https://typedigital.atlassian.net/browse/LF-7?atlOrigin=eyJpIjoiZmIzZjBjNGNlZGJjNDY3MzhkN2QzNzRlMmQ2MjllZGIiLCJwIjoiaiJ9) or a future ticket introducing templates)

graphql/concrete_notification/fields.ts
jeffersondarcy: I think you are right and this information should come from Notification. Not sure if it should change the query to get this over notification or would it be an option to "transfer" this field directly to concrete_notification

graphql/concrete_notification/dummy_data.ts
```diff
@@ -0,0 +1,20 @@
+import { ConcreteNotification, ConcreteNotificationState } from '../../common/entity/ConcreteNotification';
+
+const ids = Array.from({ length: 10 }, (_, i) => i + 1);
+const dummyMinutes = [0, 5, 10, 15, 20, 30, 60, 120, 150, 60 * 24, 60 * 48];
+
+// Dummy dates indexed by dummy ids
+export const getDummyCreatedAt = (id: number): Date => new Date(Date.now() - dummyMinutes[id] * 1000 * 60);
```
Jonasdoubleyou, line 7: What is the purpose of "createdAt"? Concrete Notifications have a "sentAt" field, which for SENT notifications is the date when the notification was sent out

Jonasdoubleyou (dismissed):

graphql/concrete_notification/dummy_data.ts
jeffersondarcy: I was not sure about sentAt - is this the timestamp, when notification was created or when it was actually sent by mailJet? if sentAt is only set once when created and not connected to mailJet, I would use it instead

Jonasdoubleyou: c.f. https://github.com/corona-school/backend/blob/master/common/entity/ConcreteNotification.ts#L35

Jonasdoubleyou: As the name suggests, it's when the notification was sent ...

jeffersondarcy: okay, that means we need a new field createdAt to know when the notification was created, right? it seems prisma could populate it automatically for us on creation

Jonasdoubleyou: No? Why would the creation date of the database entry be of relevance?

PR #404, opened 2022-11-18
Websocket service

timidak:
- Added websocket service with client management
- Added functionalities to be later used by channel

Can be tested via Postman or Insomnia with a `ws` request to the web server with a url param for the connection id.


common/websocket/index.ts
```diff
@@ -0,0 +1,115 @@
...
+        if (!this.instance && !server) {
+            throw new Error('No instance running and no instanciation possible without a server.');
+        }
+        if (!this.instance) {
+            this.instance = new WebSocketService(server);
+        }
+        return this.instance;
+    }
+
+    getCients(): ConnectedClients {
```
jeffersondarcy, line 40: would it make sense to have an array of ids as optional parameter to only receive connected Clients which will get a notification? than we might not need isClientOnline method

common/websocket/index.ts
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * initializes the events and handlers for the websocket server
+     */
+    configure(): void {
+        this.websocketServer.on('connection', (ws: WebSocket & { id: string }, request) => {
+            try {
+                // We should do some kind of auth check here, so that the token in the header matches the param
+                // otherwise some user could get the notifications of otherones just by sending the request with another userIds
+                const id = getUserIdFromConnectionRequest(request.url);
+                ws['id'] = id;
```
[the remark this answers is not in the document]
LomyW: maybe using the `sessionId` instead of?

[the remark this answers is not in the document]
timidak: I added an array of websocket per userId and I also modified some of the methods to support multiple websockets per user. 🙂

common/websocket/index.ts
timidak: I think I'm not getting it, you should only get connected clients with this method. 😄

LomyW (commented):
well done! 😊

common/websocket/index.ts
```diff
@@ -41,52 +42,79 @@ class WebSocketService {
...
+        Object.keys(this.connectedClients).forEach((key) => {
+            const idx = this.connectedClients[key].findIndex((ws: WebSocket & { clientId: string }) => ws.clientId === id);
+            if (idx !== -1) {
+                this.connectedClients[key].splice(idx, 1);
+            }
+            if (this.connectedClients[key].length === 0) {
+                delete this.connectedClients[key];
+            }
+        });
         delete this.connectedClients[id];
```
jeffersondarcy, line 59: not sure about this line - shouldn't the array be only indexed by userId?

timidak: I just forgot to delete this line. Thank's for your attention.
I will remove that line.

common/websocket/index.ts
```diff
@@ -41,52 +42,79 @@ class WebSocketService {
         return this.connectedClients;
     }
 
-    private addClient(id: string, connection: WebSocket) {
-        this.connectedClients[id] = connection;
+    private addClient(userId: string, connection: WebSocket) {
+        (this.connectedClients[userId] = this.connectedClients[userId] || []).push(connection);
     }
 
     private removeClient(id: string) {
```
jeffersondarcy, line 49: you can simplify the method when you store the userId on ws object

common/websocket/index.ts
```diff
@@ -41,52 +42,79 @@ class WebSocketService {
...
             try {
-                // We should do some kind of auth check here, so that the token in the header matches the param
-                // otherwise some user could get the notifications of otherones just by sending the request with another userIds
-                const id = getUserIdFromConnectionRequest(request.url);
-                ws['id'] = id;
-                this.addClient(id, ws);
-                log.info(`Connected websocket client with id: ${id}`);
+                let sessionUser: GraphQLUser;
+
+                if (request.headers['authorization'] && request.headers['authorization'].startsWith('Bearer ')) {
```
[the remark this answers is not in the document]
Jonasdoubleyou: Yes, Basic Authentication is only used for "technical users" like ADMIN, they do not need push notifications or other real time updates :)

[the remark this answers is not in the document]
timidak: Resolving this for now 😄

common/websocket/index.ts
```diff
@@ -0,0 +1,144 @@
...
+import { getUserForSession, GraphQLUser } from '../../graphql/authentication';
+
+type UserId = string;
+
+type ConnectedClients = {
+    [k: UserId]: WebSocket[];
+};
+
+type Message = {
+    notificationId: string;
```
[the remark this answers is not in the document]
timidak: I chose to extend the type later and leave it only with the `concreteNotificationId` for now. But using a discriminated union as you proposed, would be fine for me.

common/websocket/index.ts
timidak: Done 🙂

common/websocket/index.ts
```diff
@@ -0,0 +1,153 @@
...
+
+type UserId = string;
+type ClientId = string;
+
+interface ExtendedWebsocket extends WebSocket {
+    clientId: ClientId;
+    userId: UserId;
+}
+
+type ConnectedClients = Map<UserId, Map<ClientId, ExtendedWebsocket>>;
```
jeffersondarcy, line 15: would ConnectedUsers maybe be a better name? see comment bellow

common/websocket/index.ts
```diff
@@ -0,0 +1,153 @@
...
+        return this.connectedClients;
+    }
+
+    private addClient(userId: string, connection: ExtendedWebsocket) {
+        let userConnections = this.connectedClients.get(userId);
+        if (!userConnections) {
+            userConnections = new Map();
+        }
+        userConnections.set(connection.clientId, connection);
+        this.connectedClients.set(userId, userConnections);
```
jeffersondarcy, line 52: just a minor thing - this line could go inside the if block above, because otherwise just adding a new connection to an existing user is enough, right?
the naming `this.connectedClients` is misleading as on this level we are dealing with users and not their clients

jeffersondarcy (commented):
Suggestion: eliminate the term "Client" and replace it with "Connection" or "User" in names accordingly to make more clear what is meant where

common/websocket/index.ts
timidak: The above if block just creates a new Map storing the user connections if none is present and then adds everything to the connectedUsers :smile:.
I adjusted the naming.

common/websocket/index.ts
```diff
@@ -0,0 +1,153 @@
...
+    userId: UserId;
+}
+
+type ConnectedUsers = Map<UserId, Map<ClientId, ExtendedWebsocket>>;
+
+type Message = {
+    concreteNotificationId: number;
+};
+
+const log = getLogger();
```
Jonasdoubleyou, line 21: Here one can also pass a topic `getLogger("Websockets")` to structure the logs

Jonasdoubleyou (approved):

LomyW (approved):
good job! 😄

jeffersondarcy (approved):

PR #408, opened 2022-11-21
Added participationCertificateUpdate endpoint, Extracted issueCertifi…

realmayus:
- Added participationCertificateUpdate endpoint
- Extracted issueCertificateRequest to its own endpoint
- Fieldresolvers for Students and Pupils

Open questions:
1. Are the mutable fields sufficient/are there fields that shouldn't be mutable (other than state)?
2. I assumed that updating the certificate is only possible as a student and only when the state is `awaiting-approval`. Is this correct?


Jonasdoubleyou (approved):

Jonasdoubleyou:
Both assumptions are IMO correct.

PR #413, opened 2022-11-23
In app channel

LomyW:
added a new channel for `inapp` messages.
_fails for now by `npm run web`_  (@Jonasdoubleyou, @timidak )

### DIFF
--- a/common/notification/channels/inapp.ts
+++ b/common/notification/channels/inapp.ts
@@ -1,0 +1,19 @@
+import { Message, WebSocketService } from '../../websocket';
+import { AttachmentGroup } from '../../attachments';
+import { Channel, Context, Person, Notification } from '../types';
+import { User } from '../../user';
+
+export const inAppChannel: Channel = {
+    type: 'inapp',
+    async send(notification: Notification, to: Person, context: Context, concreteID: number, attachments?: AttachmentGroup, user?: User): Promise<void> {
+        const ws = WebSocketService.getInstance();
+        const message: Message = {
+            concreteNotificationId: concreteID,
+        };
+        await ws.sendMessageToUser(user.userID, message);
+    },
+    canSend: (_notification: Notification, user?: User) => {
+        const ws = WebSocketService.getInstance();
+        return ws.isUserOnline(user.userID);
+    },
+};
--- a/common/notification/channels/inapp.ts
+++ b/common/notification/channels/inapp.ts
@@ -5,12 +5,12 @@ import { User } from '../../user';
 
 export const inAppChannel: Channel = {
     type: 'inapp',
-    async send(notification: Notification, user: User, context: Context, concreteID: number, attachments?: AttachmentGroup): Promise<void> {
+    async send(notification: Notification, to: User, context: Context, concreteID: number, attachments?: AttachmentGroup): Promise<void> {
         const ws = WebSocketService.getInstance();
         const message: Message = {
             concreteNotificationId: concreteID,
         };
-        await ws.sendMessageToUser(user.userID, message);
+        await ws.sendMessageToUser(to.userID, message);
     },
     canSend: (_notification: Notification, user?: User) => {
         const ws = WebSocketService.getInstance();
--- a/common/notification/channels/inapp.ts
+++ b/common/notification/channels/inapp.ts
@@ -5,7 +5,7 @@ import { User } from '../../user';
 
 export const inAppChannel: Channel = {
     type: 'inapp',
-    async send(notification: Notification, to: Person, context: Context, concreteID: number, attachments?: AttachmentGroup, user?: User): Promise<void> {
+    async send(notification: Notification, user: User, context: Context, concreteID: number, attachments?: AttachmentGroup): Promise<void> {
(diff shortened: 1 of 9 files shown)


common/notification/index.ts
```diff
@@ -258,19 +259,24 @@ async function deliverNotification(
...
 
         if (notification.hookID) {
-            const newUser = getUserForTypeORM(user);
             await triggerHook(notification.hookID, newUser);
         }
 
         // TODO: Check if user silenced this notification
 
-        await channel.send(notification, user, context, concreteNotification.id, attachments);
+        Promise.all(
```
Jonasdoubleyou, line 274: Missing await

common/notification/index.ts
```diff
@@ -258,19 +259,24 @@ async function deliverNotification(
     };
 
     try {
-        const channel = channels.find((it) => it.canSend(notification));
-        if (!channel) {
+        const newUser = getUserForTypeORM(user);
```
jeffersondarcy, line 262: would it be better to rename `newUser` to something else?

common/notification/index.ts
```diff
@@ -258,19 +259,24 @@ async function deliverNotification(
...
-            const newUser = getUserForTypeORM(user);
             await triggerHook(notification.hookID, newUser);
         }
 
         // TODO: Check if user silenced this notification
 
-        await channel.send(notification, user, context, concreteNotification.id, attachments);
+        await Promise.all(
+            channelsToSendTo.map(async (channel) => {
+                await channel.send(notification, user, context, concreteNotification.id, attachments, newUser);
```
jeffersondarcy, line 276: is the id from user different from newUser? do we need a second user parameter?

common/notification/index.ts
Jonasdoubleyou: Maybe newUser -> user, user -> legacyUser? ;)

common/notification/index.ts
Jonasdoubleyou: I think we can fully move to User here and drop Person from the Channel interface, as the only field read in the Mailjet channel is the email, which is present in both anyways.

common/notification/channels/mailjet.ts
```diff
@@ -18,13 +19,13 @@ const senderEmails: { [sender in NotificationSender]: string } = {
 
 export const mailjetChannel: Channel = {
     type: 'mailjet',
-    async send(notification: Notification, to: Person, context: Context, concreteID: number, attachments?: AttachmentGroup) {
+    async send(notification: Notification, user: User, context: Context, concreteID: number, attachments?: AttachmentGroup) {
```
timidak, line 22: imho, I would call it the variable receiver or to of type user as it is the role of the user in this context.

common/notification/index.ts
```diff
@@ -258,19 +259,24 @@ async function deliverNotification(
     };
 
     try {
-        const channel = channels.find((it) => it.canSend(notification));
-        if (!channel) {
+        const legacyUser = getUserForTypeORM(user);
```
timidak, line 262: iirc, this might not be the legacy user but the newUser, just calling it user and the other one in the function signature legacyUser should be fine.
Please correct me if I'm wrong @Jonasdoubleyou.

common/websocket/index.ts
```diff
@@ -73,13 +73,24 @@ class WebSocketService {
...
-        if (!connections) {
-            throw new Error('No connections found.');
-        }
-        for (let [_, connection] of connections) {
-            connection.send(message, (err) => log.error(`Error while sending websocket message: ${err.message}`));
+    async sendMessageToUser(userId: string, message: Message): Promise<void> {
+        try {
+            const connections = this.getConnectionsByUserId(userId);
+            if (!connections) {
+                throw new Error('No connections found.');
```
timidak, line 80:
```suggestion
                throw new Error(`No connections found for user with userId: ${userId}`);
```

timidak (changes requested):
A few changes left. Everything else looks pretty good to me 👍🏼

common/notification/index.ts
timidak: @LomyW I guess I was right with my suggestion as you can see here 😄

common/notification/channels/inapp.ts
```diff
@@ -0,0 +1,19 @@
...
+export const inAppChannel: Channel = {
+    type: 'inapp',
+    async send(notification: Notification, to: User, context: Context, concreteID: number, attachments?: AttachmentGroup): Promise<void> {
+        const ws = WebSocketService.getInstance();
+        const message: Message = {
+            concreteNotificationId: concreteID,
+        };
+        await ws.sendMessageToUser(to.userID, message);
+    },
+    canSend: (_notification: Notification, user?: User) => {
```
timidak, line 15: `user` here isn't really optional and we're anyways passing it to canSend in `notification/index.ts`. So maybe this and additionally the `Channel` interface needs to be adjusted.

timidak (approved):
The last change is nice to have, but everything else looks pretty good to me 👍🏼

jeffersondarcy (approved):

PR #414, opened 2022-11-23
Fix cyclic dependencies by moving Session to common

Jonasdoubleyou:


timidak (approved):
Works on my machine and the changes are fine for me. 
Should we enable an eslint-rule to warn us about dependency cycles?

Jonasdoubleyou:
Sounds good @timidak

PR #417, opened 2022-11-25
Add: anonymise-attendance-log job

Leo310:
Resolving Issue: corona-school/project-user#505

Not tested and not sure how I should reset the pupilID.
And anonymsing the transaction log seems quite cumbersome, because as far as I could understand it, the courseAttendanceLog data is stringified and then stored in the transaction log.

### DIFF
--- a/jobs/list.ts
+++ b/jobs/list.ts
@@ -10,6 +10,7 @@ import { cleanupSecrets } from '../common/secret';
 import redactInactiveAccounts from './periodic/redact-inactive-accounts';
 import dropOldNotificationContexts from './periodic/drop-old-notification-contexts';
 import { runInterestConfirmations } from '../common/match/pool';
+import anonymiseAttendanceLog from './periodic/anonymise-attendance-log';
 
 // A list of all jobs that should be scheduled at the moment
 export const allJobs: CSCronJob[] = [
@@ -34,6 +35,7 @@ export const allJobs: CSCronJob[] = [
     // every 10 minutes during the day (to distribute load and send out notifications faster)
     { cronTime: '00 */10 * * * *', jobFunction: Notification.checkReminders },
     // each night - database cleanups
+    { cronTime: '00 00 05 * * *', jobFunction: anonymiseAttendanceLog },
     { cronTime: '00 00 04 * * *', jobFunction: cleanupSecrets },
     { cronTime: '00 00 02 * * *', jobFunction: redactInactiveAccounts },
     { cronTime: '00 00 01 * * *', jobFunction: dropOldNotificationContexts },
--- a/jobs/periodic/anonymise-attendance-log/index.ts
+++ b/jobs/periodic/anonymise-attendance-log/index.ts
@@ -5,6 +5,6 @@ const DELETE_AFTER = 7; // in days
 const logger = getLogger();
 
 export default async function execute() {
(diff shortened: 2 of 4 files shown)


jobs/periodic/anonymise-attendance-log/index.ts
```diff
@@ -0,0 +1,18 @@
+import { getLogger } from "log4js";
+import { getManager } from "typeorm";
+import { CourseAttendanceLog } from "../../../common/entity/CourseAttendanceLog";
+
+const DELETE_AFTER = 7; // in days
+const logger = getLogger();
+
+export default async function execute() {
+    const entityManager = getManager();
+    await entityManager
```
Jonasdoubleyou, line 10: Prisma please :) 
And please also check that this does not also delete the pupil, the schema has a 'cascading' setting and I'm not sure what that does.

jobs/list.ts
```diff
@@ -34,6 +35,7 @@ export const allJobs: CSCronJob[] = [
     // every 10 minutes during the day (to distribute load and send out notifications faster)
     { cronTime: '00 */10 * * * *', jobFunction: Notification.checkReminders },
     // each night - database cleanups
+    { cronTime: '00 00 05 * * *', jobFunction: anonymiseAttendanceLog },
```
Jonasdoubleyou, line 38: If you alao add this to manualExecution.ts, you can run it via GraphQL with _executeJob, thats easier to test ...

jobs/periodic/anonymise-attendance-log/index.ts
```diff
@@ -0,0 +1,18 @@
...
+        .createQueryBuilder(CourseAttendanceLog, 'courseAttendanceLog')
+        .update()
+        .where('courseAttendanceLog.createdAt < :date', { date: Date.now() - DELETE_AFTER })
+        .set({ ip: "", pupil: { id: 0 } })
```
Jonasdoubleyou, line 14: So here the problem is that the pupilId is non nullable?

Leo310: So what pupilId should I then reference?
image

Should it point to an existing empty Mustermann pupil?

lsc0498: Yes, that's fine. There is no need to anonymize the (internal) pupilId. As long as no IP address, names or email addresses are left if a user deactivates their account it should be okay 👍

jobs/periodic/anonymise-attendance-log/index.ts
```diff
@@ -0,0 +1,10 @@
+import { prisma } from "../../../common/prisma";
+import { getLogger } from "log4js";
+
+const DELETE_AFTER = 7; // in days
+const logger = getLogger();
+
+export default async function execute() {
+    await prisma.course_attendance_log.updateMany({ where: { createdAt: { lt: new Date(Date.now() - DELETE_AFTER) } }, data: { ip: "" } });
```
Jonasdoubleyou, line 8: Date.now() is in milliseconds, - 7 is 7 milliseconds before

Jonasdoubleyou (changes requested):

Leo310:
"And anonymsing the transaction log seems quite cumbersome, because as far as I could understand it, the courseAttendanceLog data is stringified and then stored in the transaction log."

Any thoughts on this point?

Jonasdoubleyou (dismissed):

Leo310:
Bisherige logs müssten trotzdem nochmal mit scripts gecleaned und anonymisiert werden.

Jonasdoubleyou (approved):

PR #425, opened 2022-11-29
Added lastTimeCheckNotification 

LomyW:
Added `lastTimeCheckedNotifications`  to `Person`
- **Query**: example `pupil`
`query {
      me {
        pupil {
          lastTimeCheckedNotifications
        }
      }
    }`
- **Mutation**: Update example `pupil`
`mutation updateLastTimeCheckedNotifications( $lastTimeCheckedNotifications: DateTime) 
{ meUpdate ( update: { pupil: { lastTimeCheckedNotifications: $lastTimeCheckedNotifications }})}`

### DIFF
--- a/graphql/student/mutations.ts
+++ b/graphql/student/mutations.ts
@@ -80,7 +80,6 @@ export class StudentUpdateInput {
     aboutMe?: string;
 
     @Field((type) => Date, { nullable: true })
-    @MaxLength(100)
     lastTimeCheckedNotifications: Date;
 }
 
--- a/graphql/student/mutations.ts
+++ b/graphql/student/mutations.ts
@@ -78,11 +78,15 @@ export class StudentUpdateInput {
     @Field((type) => String, { nullable: true })
     @MaxLength(500)
     aboutMe?: string;
+
+    @Field((type) => Date, { nullable: true })
+    @MaxLength(100)
+    lastTimeCheckedNotifications: Date;
 }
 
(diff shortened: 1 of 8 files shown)


graphql/student/mutations.ts
```diff
@@ -109,6 +113,7 @@ export async function updateStudent(context: GraphQLContext, student: Student, u
             registrationSource: ensureNoNull(registrationSource),
             state: ensureNoNull(state),
             aboutMe: ensureNoNull(aboutMe),
+            lastTimeCheckedNotifications: ensureNoNull(lastTimeCheckedNotifications),
```
jeffersondarcy, line 116: is this correct? field can be null

jeffersondarcy: is this the right mutation? we are usually updating user fields and last checked date separately, right?

jeffersondarcy: same question with other roles

Jonasdoubleyou: ensureNoNull replaces null with undefined (which is desired, in Prisma undefined means 'ignore' whether null means 'set to NULL'), so the use here is correct. 

meUpdate is supposed to update the user record with simple write operations, so as long as the time is only written I think it's fine, if you plan to add more logic here then I guess it would deserve a separate mutation.

jeffersondarcy (approved):

LomyW:
lastTimeCheckedNotifications is added in another PR #429

PR #429, opened 2022-12-02
User Notification Preferences

LomyW:
some changes to save the `notification preferences` and mutate them.


common/entity/Person.ts
```diff
@@ -66,6 +66,16 @@ export abstract class Person {
...
+
+    @Column({
+        type: 'timestamp',
+        nullable: true,
+        default: new Date(),
+    })
+    lastTimeCheckedNotifications: Date;
+
+    @Column({ nullable: true, type: 'json' })
+    notificationPreferences: any;
```
jeffersondarcy, line 78: why not string?

common/migration/1669676905990-AddLastCheckedNotifications.ts
```diff
@@ -0,0 +1,20 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+
+export class AddLastCheckedNotifications1669676905990 implements MigrationInterface {
+    name = 'AddLastCheckedNotifications1669676905990'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "pupil" ADD "lastTimeCheckedNotifications" TIMESTAMP DEFAULT '"2022-11-28T23:08:26.067Z"'`);
```
jeffersondarcy, line 7: why not default NULL?

common/entity/Person.ts
```diff
@@ -66,6 +66,16 @@ export abstract class Person {
         default: false,
     })
     isRedacted: boolean;
+
+    @Column({
+        type: 'timestamp',
+        nullable: true,
+        default: new Date(),
```
jeffersondarcy, line 73: I am not sure about default. When the column is created everything should be NULL, but on writing current timestamp is right

common/user/session.ts
```diff
@@ -15,6 +15,8 @@ export const UNAUTHENTICATED_USER = {
     firstname: '',
     lastname: '',
     userID: '-/-',
+    lastTimeCheckedNotifications: new Date(),
```
jeffersondarcy, line 18: I think it's wrong - all present notifications would be older than the current timestamp

common/migration/1669676905990-AddLastCheckedNotifications.ts
jeffersondarcy, line 7: I think you need to alter every table in two steps, first using TIMESTAMP DEFAULT NULL to fill the column with null values, than alter DEFAULT to current timestamp for future mutating queries https://www.postgresqltutorial.com/postgresql-date-functions/postgresql-current_timestamp/

common/util/preferences-utils.ts
```diff
@@ -0,0 +1,9 @@
+type ChannelPreference = {
```
jeffersondarcy, line 1: I would call it `Channels`

common/util/preferences-utils.ts
```diff
@@ -0,0 +1,9 @@
...
+    email: boolean;
+};
+
+type PreferenceCategories = 'chat' | 'match' | 'course' | 'appointment' | 'survey' | 'news';
```
jeffersondarcy, line 5: I would call it `MessageTypes` or `NotificationTypes` or maybe `MessageEvents`

jeffersondarcy (changes requested):
please check my comments on timestamp defaults

common/entity/Person.ts
LomyW: Idk, I think this code might be from @Jonasdoubleyou. Maybe the type `NotificationPrefereces | null` could be used here, but I'm not sure.

common/migration/1669676905990-AddLastCheckedNotifications.ts
LomyW: This was automatically generated, I wouldn't touch this.

common/user/session.ts
LomyW: This is an unauthorized user, I just set a value to satisfy TS an unauthorized user shouldn't have any notifications.

common/entity/Person.ts
Jonasdoubleyou: Ah yeah, a proper type could also be used here, was too lazy 😅

common/entity/Person.ts
Jonasdoubleyou: Ah, maybe we should set it to the timestamp that was generated in the migration, otherwise TypeORM might generate a new migration each time ... (haven't checked this though)

common/entity/Person.ts
```diff
@@ -66,6 +67,16 @@ export abstract class Person {
...
+
+    @Column({
+        type: 'timestamp',
+        nullable: true,
+        default: new Date(0),
+    })
+    lastTimeCheckedNotifications: Date;
+
+    @Column({ nullable: true, type: 'json' })
+    notificationPreferences: Prisma.JsonValue;
```
jeffersondarcy, line 79: @Jonasdoubleyou is Prisma.JsonValue the correct type?

Jonasdoubleyou: No, NotificationPreferences was better.

Jonasdoubleyou: The entity/* Files are from TypeORM, maybe better not to mix in Prisma Types here

common/notification/index.ts
```diff
@@ -261,16 +262,39 @@ async function deliverNotification(
...
+                if (enabledChannels[channel] === true) {
+                    activeChannelForNotificationType.push(channel);
+                }
+            }
+        }
+
+        const channelsToSendTo = channels
+            .filter((it) => it.canSend(notification, user))
+            .filter((c) => {
+                const compareValue = c.type === 'mailjet' ? 'email' : c.type;
```
Jonasdoubleyou, line 291: Uh, parity please between preferences and the channel names. Feel free to rename the channel (I guess "email" is a better name anyways)

common/notification/index.ts
```diff
@@ -261,16 +262,39 @@ async function deliverNotification(
...
+
         // TODO: Check if user silenced this notification
+        const { notificationPreferences } = await getStudentOrPupil(user);
+
+        if (notificationPreferences && typeof notificationPreferences === 'string') {
+            const enabledChannels: { [channel: string]: boolean } = JSON.parse(notificationPreferences)[messageType];
+
+            for (const channel in enabledChannels) {
+                if (enabledChannels[channel] === true) {
+                    activeChannelForNotificationType.push(channel);
```
Jonasdoubleyou, line 283: I don't see the point of `activeChannelForNotificationType`, what benefit does this array have over the `enabledChannels` dictionary?

common/notification/index.ts
```diff
@@ -261,16 +262,39 @@ async function deliverNotification(
...
+                }
+            }
+        }
+
+        const channelsToSendTo = channels
+            .filter((it) => it.canSend(notification, user))
+            .filter((c) => {
+                const compareValue = c.type === 'mailjet' ? 'email' : c.type;
+                return activeChannelForNotificationType.includes(compareValue);
```
Jonasdoubleyou, line 292: `enabledChannels[ channel.type ]` ...

common/notification/index.ts
```diff
@@ -261,16 +262,39 @@ async function deliverNotification(
...
+        const { messageType } = getMessage(concreteNotification);
+
+        // default channel is inapp
+        const activeChannelForNotificationType: string[] = ['inapp'];
+
         // TODO: Check if user silenced this notification
+        const { notificationPreferences } = await getStudentOrPupil(user);
+
+        if (notificationPreferences && typeof notificationPreferences === 'string') {
+            const enabledChannels: { [channel: string]: boolean } = JSON.parse(notificationPreferences)[messageType];
```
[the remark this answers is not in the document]
Jonasdoubleyou: Also this should gracefully fall back when `messageType` is not present in the preferences

common/notification/index.ts
```diff
@@ -261,16 +262,39 @@ async function deliverNotification(
...
+            for (const channel in enabledChannels) {
+                if (enabledChannels[channel] === true) {
+                    activeChannelForNotificationType.push(channel);
+                }
+            }
+        }
+
+        const channelsToSendTo = channels
+            .filter((it) => it.canSend(notification, user))
+            .filter((c) => {
```
Jonasdoubleyou, line 290: I'm not really a fan of one character variable names, how about `channel`?

common/notification/index.ts
```diff
@@ -261,16 +262,39 @@ async function deliverNotification(
...
             await triggerHook(notification.hookID, user);
         }
 
+        const { messageType } = getMessage(concreteNotification);
+
+        // default channel is inapp
+        const activeChannelForNotificationType: string[] = ['inapp'];
+
         // TODO: Check if user silenced this notification
+        const { notificationPreferences } = await getStudentOrPupil(user);
```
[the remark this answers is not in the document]
Jonasdoubleyou: Here `queryUser({ notificationPreferences: true })` would also work

common/notification/index.ts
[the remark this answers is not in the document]
Jonasdoubleyou: Also I'd prefer typing whatever `JSON.parse` returns before then accessing it's members.

graphql/authorizations.ts
```diff
@@ -240,7 +240,8 @@ export const authorizationModelEnhanceMap: ModelsEnhanceMap = {
             state: adminOrOwner,
             teacherEmailAddress: adminOrOwner,
             coduToken: adminOrOwner,
-
+            lastTimeCheckedNotifications: adminOrOwner,
```
timidak, line 243: Why is this here as well? 
Something off with the history?

timidak (commented):
Just a few questions 😄

common/notification/index.ts
```diff
@@ -261,21 +282,26 @@ async function deliverNotification(
...
 
         if (notification.hookID) {
             await triggerHook(notification.hookID, user);
         }
 
-        // TODO: Check if user silenced this notification
+        // default channel is webApp is always enabled
+        const enabledChannels: Array<Channel> = [channels['webApp']];
+
+        const messageTypeChannelPreferences = await getNotificationChannelPreferences(user, concreteNotification);
```
timidak, line 293: channelPreferencesForMessageType?

graphql/authorizations.ts
jeffersondarcy: one is for Pupil and another for Student :)

common/notification/index.ts
```diff
@@ -243,6 +246,21 @@ async function createConcreteNotification(
     return concreteNotification;
 }
 
+const getNotificationChannelPreferences = async (user: User, concreteNotification: ConcreteNotification): Promise<Channels> => {
+    const { messageType } = getMessage(concreteNotification);
```
Jonasdoubleyou, line 250: So the message type now determines the notification channel? I don't like that. This should be notification.type.

jeffersondarcy: currently we only map types in templates. as we are moving templates to db anyway we could add the type field to notification table. would that be a good solution

Jonasdoubleyou: Yes, can also be done as part of LF-60 if you want to bring this PR in

Jonasdoubleyou (approved):

common/notification/index.ts
jeffersondarcy: so I will keep this solution till change it in LF-60, okay?

PR #433, opened 2022-12-06
LF 15 - real notifications

jeffersondarcy:
Templating for Concrete Notifications.  
Removed individual field resolvers for Message Parts (body, header). 
Removed old dummy data

### DIFF
--- a/notifications/templates.ts
+++ b/notifications/templates.ts
@@ -19,12 +19,30 @@ export const getMessage = (concreteNotification: ConcreteNotification, user: Gra
             headline: `bla bla ${firstname}`,
             body: `bla bla ${firstname} bla bla ${lastname}`,
             navigateTo: 'welcome',
-            messageType: MessageType.APPOINTMENT,
+            messageType: MessageType.COURSE,
         },
         3: {
             headline: `bla bla ${firstname}`,
             body: `bla bla ${firstname} bla bla ${lastname}`,
-            messageType: MessageType.MATCH,
+            messageType: MessageType.MESSAGE,
+        },
+        4: {
+            headline: `bla bla ${firstname}`,
+            body: `bla bla ${firstname} bla bla ${lastname}`,
+            navigateTo: `http://www.somewhere`,
+            isUrlExternal: true,
+            messageType: MessageType.CHAT,
+        },
+        5: {
+            headline: `bla bla ${firstname}`,
+            body: `bla bla ${firstname} bla bla ${lastname}`,
+            navigateTo: 'welcome',
+            messageType: MessageType.NEWS,
+        },
+        6: {
+            headline: `bla bla ${firstname}`,
+            body: `bla bla ${firstname} bla bla ${lastname}`,
+            messageType: MessageType.SURVEY,
         },
     };
 


notifications/templates.ts
```diff
@@ -0,0 +1,45 @@
...
+        },
+        // 0 used as fallback
+        0: {
+            header: `Message not found`,
+            body: `Error: message details not found. Message for ${firstname} ${lastname}`,
+            messageType: MessageType.MATCH,
+            navigateTo: 'welcome',
+        },
+    };
+    const index = templates.hasOwnProperty(notificationID) ? notificationID : 0;
```
Jonasdoubleyou, line 43: Uh, instead of a magic number "0" I'd prefer a separate constant for the error message, then `templates[notificationID] ?? MESSAGE_NOT_FOUND`.

notifications/templates.ts
```diff
@@ -0,0 +1,45 @@
...
+    messageType: MessageType;
+} & NotificationAction;
+
+export const getMessage = (concreteNotification: ConcreteNotification, user: GraphQLUser): MessageTemplate => {
+    const { firstname, lastname } = user;
+    const { notificationID, context } = concreteNotification;
+
+    const templates: { [key: number]: MessageTemplate } = {
+        1: {
+            header: `bla bla ${firstname}`,
```
Jonasdoubleyou, line 19: So as a next step you would then move the MessageTemplates to the DB correct? Or how are you going to proceed here? (Good intermediate step though)

notifications/templates.ts
```diff
@@ -0,0 +1,45 @@
+import { MessageType } from './messageTypes';
+import { GraphQLUser } from '../common/user/session';
+import { Concrete_notification as ConcreteNotification } from '../graphql/generated';
+
+// only actionPath or url is optionally allowed but not both
+type NotificationAction = { navigateTo?: string; url?: undefined } | { url: string; navigateTo?: undefined };
+export type MessageTemplate = {
```
Jonasdoubleyou, line 7: Maybe it makes sense to already distinguish "MessageTemplate" and "Message" here?

graphql/user/fields.ts
```diff
@@ -77,7 +76,7 @@ export class UserFieldsResolver {
     @FieldResolver((returns) => [ConcreteNotification])
     @Authorized(Role.OWNER, Role.ADMIN)
     async concreteNotifications(@Root() user: User): Promise<ConcreteNotification[]> {
-        return getDummyConcreteNotifications(user.userID);
+        return await prisma.concrete_notification.findMany({ where: { userId: user.userID } });
```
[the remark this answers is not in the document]
Jonasdoubleyou: Also have a look at the @LimitedQuery functionality, findMany can potentially create large result sets (especially on our largest table concrete_notifications)

notifications/templates.ts
jeffersondarcy: the default message was more for testing purposes, because I was not sure which notifications I would get

notifications/templates.ts
jeffersondarcy: we wanted to have them in a file, so they will stay here until we decide for a different solution (if we decide to change it)

graphql/user/fields.ts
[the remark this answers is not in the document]
jeffersondarcy: changed

graphql/concrete_notification/fields.ts
```diff
@@ -39,26 +39,12 @@ export class ExtendedFieldsConcreteNotificationResolver {
         return await prisma.notification.findUnique({ where: { id: concreteNotification.notificationID } });
     }
 
-    @FieldResolver((returns) => String)
+    @FieldResolver((returns) => ConcreteNotification)
```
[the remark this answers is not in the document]
jeffersondarcy: changed to Object because MessageTemplate is a type and not a class and the linter complains in this case. tell me if there is a better way

notifications/templates.ts
jeffersondarcy: changed error handling

graphql/concrete_notification/fields.ts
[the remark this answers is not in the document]
Jonasdoubleyou: You can create a class for the type so that GraphQL has proper reflection and can generate the type definition from it. See https://github.com/corona-school/backend/blob/91031aa8762ca38eff877ecb0ca9bfa1c2436a2b/graphql/types/user.ts#L6 for an example

graphql/user/fields.ts
```diff
@@ -76,8 +76,9 @@ export class UserFieldsResolver {
 
     @FieldResolver((returns) => [ConcreteNotification])
     @Authorized(Role.OWNER, Role.ADMIN)
+    @LimitedQuery()
```
Jonasdoubleyou, line 79: For this you also need `@Arg('take')` (and `@Arg('skip')`) (c.f. https://github.com/corona-school/backend/blob/master/graphql/subcourse/fields.ts#L65 for an analogous case)

notifications/messageTypes.ts
```diff
@@ -0,0 +1,9 @@
+export enum MessageType {
+    MESSAGE = 'MessageType.MESSAGE',
+    MATCH = 'MessageType.MATCH',
+    COURSE = 'MessageType.COURSE',
+    APPOINTMENT = 'MessageType.APPOINTMENT',
+    SURVEY = 'MessageType.SURVEY',
+    NEWS = 'MessageType.NEWS',
+    CHAT = 'MessageType.CHAT',
+}
```
LomyW, line 9: why `MessageType.MESSAGE` and not `message` as a string?
I need the following MessageTypes: **chat, match, course, appointment, survey, news**

graphql/concrete_notification/fields.ts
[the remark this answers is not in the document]
jeffersondarcy: I have replaced type with an abstract class for MessageTemplate

notifications/templates.ts
jeffersondarcy: I have renamed it to NotificationMessage now, I don't see a place for a second type

notifications/messageTypes.ts
jeffersondarcy: fixed

graphql/user/fields.ts
jeffersondarcy: fixed.  
schoolForTeacherEmail in school/fields.ts was not using this args, so I didn't add them either first - I think the query won't make use of them

Jonasdoubleyou: @jeffersondarcy oh no, actually this is a bug and the usage of LimitedQuery there is wrong. Please pass on take and skip to Prisma  `findMany({ take, skip, where: { ... }, orderBy: { sentAt: 'desc' } })`. I just have a bad feeling exposing unlimited queries to users.

graphql/concrete_notification/fields.ts
```diff
@@ -39,26 +39,12 @@ export class ExtendedFieldsConcreteNotificationResolver {
         return await prisma.notification.findUnique({ where: { id: concreteNotification.notificationID } });
     }
 
-    @FieldResolver((returns) => String)
+    @FieldResolver((returns) => NotificationMessage)
```
Jonasdoubleyou, line 42: Have you tried this? I don't think this will work without the @Field() and @ObjectType() annotations.

jeffersondarcy: it works, but I have to query the fields

graphql/user/fields.ts
jeffersondarcy: fixed

Jonasdoubleyou (approved):

PR #446, opened 2022-12-20
Prevent students from changing their name by themselves

realmayus:
See https://github.com/corona-school/project-user/issues/538

### DIFF
--- a/graphql/me/mutation.ts
+++ b/graphql/me/mutation.ts
@@ -311,8 +311,10 @@ export class MutateMeResolver {
             if (pupil) {
                 throw new PrerequisiteError(`Tried to update pupil data on student`);
             }
+            student.firstname = undefined; // don't allow students to change their names by themselves as their identity is verified during screening
+            student.lastname = undefined;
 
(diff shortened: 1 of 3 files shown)


Jonasdoubleyou:
What about studentUpdate?

graphql/me/mutation.ts
```diff
@@ -311,8 +311,10 @@ export class MutateMeResolver {
             if (pupil) {
                 throw new PrerequisiteError(`Tried to update pupil data on student`);
             }
+            student.firstname = undefined; // don't allow students to change their names by themselves as their identity is verified during screening
+            student.lastname = undefined;
 
-            await updateStudent(context, prevStudent, { firstname, lastname, ...student });
+            await updateStudent(context, prevStudent, student);
```
Jonasdoubleyou, line 317: I'd prefer if we do this check in the updateStudent function instead, then we also cover the updateStudent query. Just as we do in there with email

Jonasdoubleyou (approved):

PR #450, opened 2022-12-29
New message categories

LomyW:
**Attention**: templates will be deleted in a upcoming story, also `messageCategories` will be removed.


notifications/defaultPreferences.ts
```diff
@@ -1,18 +1,26 @@
-import { NotificationPreferences } from '../graphql/types/preferences';
+type EmailChannel = { email: boolean };
+type Channel = { [channel: string]: boolean };
+export type Preference = { [category: string]: EmailChannel };
 
-export const DEFAULT_PREFERENCES: NotificationPreferences = {
+export const DEFAULT_PREFERENCES: Preference = {
```
timidak, line 5: I guess the default prefs for marketing notifications should be "off" so `false`.

notifications/messageCategories.ts
```diff
@@ -1,31 +1,29 @@
...
-type MessageCategoriesMapping = { [key in MessageCategory]: MessageCategory };
-const MessageCategories: MessageCategoriesMapping = (() => {
+const MessageCategories = (() => {
     const result = {};
     for (const messageCategory of categories) {
         result[messageCategory] = messageCategory;
     }
-    return result as MessageCategoriesMapping;
+    return result;
 })();
```
timidak, line 27: Do we even need this?

notifications/messageCategories.ts
```diff
@@ -1,31 +1,29 @@
...
-    'training',
-    'events',
-    'newsoffer',
+    'event',
     'request',
-    'learnoffer',
-    'alternativeoffer',
-    'feedback',
+    'alternative',
 ];
```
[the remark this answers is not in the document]
timidak: With this union type you can type everything that is a category. 😄

timidak (changes requested):
Some changes to be made - the rest looks pretty good!

notifications/messageCategories.ts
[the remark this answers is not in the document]
LomyW: as commented above, `messageCategories` are removed as templates are moved to the DB. This file was only used to create templates.

notifications/messageCategories.ts
LomyW: in the comment above :)

notifications/defaultPreferences.ts
```diff
@@ -1,18 +1,26 @@
...
+    suggestion: { email: false },
+    announcement: { email: false },
+    call: { email: false },
+    news: { email: false },
+    event: { email: false },
+    request: { email: false },
+    alternative: { email: false },
+};
+
+export const FIX_PREFERENCES: Preference = {
```
jeffersondarcy, line 19: Maybe we should not call it ENABLED_NOTIFICATIONS or FORCED_TYPES

jeffersondarcy: would it be better to have an export with all preferences already merged so we don't need to merge them in common/notification/index.ts?

notifications/defaultPreferences.ts
```diff
@@ -1,18 +1,28 @@
-import { NotificationPreferences } from '../graphql/types/preferences';
+type EmailChannel = { email: boolean };
+type Channel = { [channel: string]: boolean };
+export type Preferences = { [category: string]: EmailChannel };
```
timidak, line 3: I would move the categories here:
const categories = [
    'chat',
    'match',
    'course',
    'appointment',
    'survey',
    'news',
    'newsletter',
    'training',
    'events',
    'newsoffer',
    'request',
    'learnoffer',
    'alternativeoffer',
    'feedback',
];
and create a union out of it (`type Categories = typeof categories[number];`)  to use it instead of the `[category: string]`

timidak (approved):
Just a minor change but not mandatory! 
Looks good to me!

PR #454, opened 2023-01-04
Fix validation promotion of course

LomyW:
A validation of the course before notify pupil about this course is only necessary when the course gets promoted, not when it gets published!


timidak (approved):
:+1: nice work

PR #464, opened 2023-01-12
Implement studentRegisterPlusMany and pupilRegisterPlusMany

realmayus:
In light of the failed automated import of LF Plus students last time, I added mutations for registering students and pupils in batches server-side.
The code uses Prisma's [Interactive Transactions](https://www.prisma.io/docs/concepts/components/prisma-client/transactions#interactive-transactions) which allows it to automatically roll back changes when a step in the registration process fails such that we don't have any incomplete accounts in the DB.

Here's the registration process for students, for example:
Untitled Diagram drawio(1)

Basically, you pass data like this to the mutation:
```json
{ "entries": [
  {
    "email": "test+dev+p20@lern-fair.de",
    "register": {...}
    "activate": {...}
    "screen": {...}
  }
]}
```
where:
* `email` is mandatory
* `register`, `activate` and `screen` are optional
If you only want to register a new account, only pass `register` data. If you only want to `activate` and `screen` an account, only pass the relevant data.
This enables you to only activate/screen existing users if you don't want to overwrite existing data.

The output looks something like this:
image

As you can see, if a single entry in the list fails to be registered/activated/screened, only it fails and the function continues processing the other entries. The reason and success state are returned for display in the Retool.

In the entire process, not a single notification is sent. We'll have to check if this is desired or if the notifications should be collected and then sent out at the end for those entries that succeeded.


package.json
```diff
@@ -13,6 +13,7 @@
     "build:clean": "(rm -rf ./built/ 2> /dev/null || find ./built/ -mindepth 1 -delete) && npm run build",
     "heroku-postbuild": "npm run build && ./setup.sh",
     "jobs": "node ./built/jobs/index.js",
+    "jobsDEV": "node ./built/jobs/manualExecution.js dropOldNotificationContexts",
```
Jonasdoubleyou, line 16: rename / remove?

Jonasdoubleyou (dismissed):
Very nice!

graphql/student/mutations.ts
```diff
@@ -160,7 +202,73 @@ export async function updateStudent(context: GraphQLContext, student: Student, u
...
+    return res;
+}
+
+async function studentRegisterPlus(data: StudentRegisterPlusInput, ctx: GraphQLContext): Promise<{ success: boolean; reason: string }> {
+    const log = logInContext('Student', ctx);
+    const { email, register, activate, screen } = data;
+    const screener = await getSessionScreener(ctx);
+
+    try {
+        if (register && register.email !== email) {
```
dhenkel92, line 214: /nit
We should try to be consistent in the way we are doing checks. Here we are checking `if (register)` whereas later it's `if(!!register)`.

dhenkel92 (commented):
Amazing PR description! It helped a lot to understand the purpose of the new endpoints and how they are supposed to work. 🙂

graphql/student/mutations.ts
```diff
@@ -160,7 +202,73 @@ export async function updateStudent(context: GraphQLContext, student: Student, u
...
+            let student;
+            if (!!register) {
+                //registration data was provided
+                if (!!existingAccount) {
+                    log.info(`Account with email ${email} already exists, updating account with registration data instead...`);
+                    // updating existing account with new registration data:
+                    student = await updateStudent(ctx, existingAccount, { ...register, projectFields: undefined, languages: undefined }, tx); // languages are added in next step (becomeTutor)
+                } else {
+                    student = await registerStudent(register, true, tx);
+                    log.info(`Registered account with email ${email}`);
```
Jonasdoubleyou, line 234: I have the fear that sooner or later something goes wrong during a mass import, and then we have to search for broken accounts. For that case we should have good logging, so I'd suggest to:
- Log the start and end of the mass import (so that one can filter for the range)
- Add Pupil({id}) and Student({id}) to every log line so that one can filter by those too

Apart from that, nice PR!

dhenkel92: I thought the same while reading the code, but Marius already took care of most cases with using transactions. 🎉 

Do you have any thoughts about what still can go wrong?

Jonasdoubleyou: Not really, but I guess the whole thing might get more complex in the future.

Maybe alternatively we could just write the result into the log? So in case someone runs the thing in the Retool but forgets to store the output, we still know what they saw.

graphql/student/mutations.ts
```diff
@@ -160,7 +202,73 @@ export async function updateStudent(context: GraphQLContext, student: Student, u
     await updateSessionUser(context, userForStudent(res));
 
     log.info(`Student(${student.id}) updated their account with ${JSON.stringify(update)}`);
+    return res;
 }
+
+async function studentRegisterPlus(data: StudentRegisterPlusInput, ctx: GraphQLContext): Promise<{ success: boolean; reason: string }> {
+    const log = logInContext('Student', ctx);
+    const { email, register, activate, screen } = data;
```
Jonasdoubleyou, line 210: Uh, we should validate and normalize the email with

email = validateEmail(email);

Also this might be a good candidate for an integration test :)

Jonasdoubleyou (approved):
Very nice!

PR #465, opened 2023-01-13
start server without resetting db

jeffersondarcy:
in index.ts the only change is in line 376, everything else is formatting

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "jobs:debug": "npm run build && node --inspect=0 ./built/jobs/index.js",
     "web": "node ./built/web/index.js",
     "web:dev": "node -r dotenv/config ./built/web/index.js",
-    "web:nosync": "node -r dotenv/config ./built/web/index.js --noSync",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -17,6 +17,7 @@
     "jobs:debug": "npm run build && node --inspect=0 ./built/jobs/index.js",
     "web": "node ./built/web/index.js",
     "web:dev": "node -r dotenv/config ./built/web/index.js",
+    "web:nosync": "node -r dotenv/config ./built/web/index.js --noSync",
```
Jonasdoubleyou, line 20: I would prefer 'keep-db' over 'no-sync'

jeffersondarcy: okay, for some reason isCommandArg is not working with ('--noSync'), do you have an idea what is different wit '--noPDF'?

jeffersondarcy: I think isCommandArg is buggy, but it has worked, because --noPDF was the only possible param

Jonasdoubleyou (approved):

PR #467, opened 2023-01-23
Hotfixes to the notification system

Jonasdoubleyou:
- HOT: Readd hooks, they were lost during merge conflict resolution
- HOT: Disable InApp Channel inside of Jobs Dyno (as no Websocket Server is present)
- Add debug logging for rollout
- Improve Tooling around Message Templates

### DIFF
--- a/common/notification/channels/inapp.ts
+++ b/common/notification/channels/inapp.ts
@@ -2,6 +2,9 @@ import { Message, WebSocketService } from '../../websocket';
 import { AttachmentGroup } from '../../attachments';
 import { Channel, Context, Notification } from '../types';
 import { User } from '../../user';
+import { getLogger } from 'log4js';
+
+const logger = getLogger('Notification InApp');
 
 export const inAppChannel: Channel = {
     type: 'inapp',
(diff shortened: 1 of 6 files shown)


common/notification/channels/inapp.ts
```diff
@@ -13,6 +16,14 @@ export const inAppChannel: Channel = {
         await ws.sendMessageToUser(to.userID, message);
     },
     canSend: (_notification: Notification, user: User) => {
+        if (!WebSocketService.hasInstance()) {
+            // In the Jobs Dyno (when sending Reminders) we do not have a connection to the Web Dyno which runs the Websocket Server
+            // thus we cannot deliver an In App Notification in that case
+            // This is not great, but hard to fix from an architectural point of view
+            logger.debug(`Did not sent In App Notification as no Websocket Server is available`);
```
Jonasdoubleyou, line 23: FYI @timidak ... I totally overlooked this. We run our backend in two different Dynos (aka 'NodeJS processes'), Notifications are sent from both, but only one has the websocket server running. Thus In App Notifications will not work for reminders for the moment. I have no idea currently how to fix that, and for the time being I guess we can keep it this way until we have reminders were In App Notifications are actually relevant.

graphql/message_translation/fields.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Message_translation as MessageTranslation } from '../generated';
+import { getNotification, getSampleContext } from '../../common/notification/notification';
+import { renderTemplate } from '../../utils/helpers';
+import { Role } from '../roles';
+
+@Resolver((of) => MessageTranslation)
+export class ExtendedFieldsMessageTranslationResolver {
+    @FieldResolver((returns) => NotificationMessageType)
+    @Authorized(Role.ADMIN)
+    async sampleMessage(@Root() messageTranslation: MessageTranslation): Promise<NotificationMessageType> {
```
Jonasdoubleyou, line 12: FYI @jeffersondarcy: I've added some utilities here to have better tooling for maintaining messages in the retool

common/notification/channels/inapp.ts
timidak: Thanks for the heads-up @Jonasdoubleyou! Let's have a quick chat about this, maybe on wednesday?

Jonasdoubleyou: Hm, I'm busy the rest of the week. So either asynchronous via Chat or next week monday?

PR #470, opened 2023-01-26
refactor: ws server closes connection on error

timidak:
- refactored terminating to closing the connection
- added initial websocket CloseCodes based on the rfc

### DIFF
--- a/common/websocket/index.ts
+++ b/common/websocket/index.ts
@@ -7,6 +7,11 @@ import { getUserForSession, GraphQLUser } from '../user/session';
 type UserId = string;
 type ConnectionId = string;
 
+enum CloseCodes {
+    NORMAL = 1000,
+    SERVER_ERROR = 1011,
+}
+
 interface ExtendedWebsocket extends WebSocket {
     connectionId: ConnectionId;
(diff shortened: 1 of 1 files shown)


common/websocket/index.ts
```diff
@@ -146,7 +151,7 @@ class WebSocketService {
                 } else {
                     log.error(`Error in websocket service.`);
                 }
-                ws.terminate();
+                ws.close(CloseCodes.SERVER_ERROR, err?.message ?? 'Internal server error while operating');
```
jeffersondarcy, line 154: what happens without explicit error code, do we get 1006?

jeffersondarcy (approved):

LomyW (approved):
nice! 👌

common/websocket/index.ts
timidak: No, it would be a normal closure so `1000` (cf. https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close#parameters). But when terminating it, as it was made before, it would result as a `1006` in the client.

Jonasdoubleyou (approved):

PR #493, opened 2023-02-25
Add: configurable ImportantInformation entity

Leo310:
Resolves part of corona-school/project-user#609.

Implementation of Important Information entity.
Not completely sure if exposing the Important_informationCrudResolver is desired in graphql/index.ts since no other resolver is implemented that way.


Jonasdoubleyou (approved):
Looks good 👍 The Crud Resolver is the right one

PR #508, opened 2023-03-07
Allow instructors to manage the waiting list

Jonasdoubleyou:
- Instructors can query the pupils on the waitinglist:
```gql
query { subcourse(...) { pupilsOnWaitinglist { id firstname lastname } } }
```

- Pupils can check whether they can join the waitinglist:
```gql
query { subcourse(...) { canJoinWaitinglist } } }
```

- Instructors can join single pupils from the waitinglist:
```gql
mutation { subcourseJoinFromWaitinglist(subcourseId: 1, pupilId: 1) }
```

### DIFF
--- a/graphql/subcourse/fields.ts
+++ b/graphql/subcourse/fields.ts
@@ -13,7 +13,7 @@ import { Bbb_meeting as BBBMeeting, Course, Lecture, Pupil, pupil_schooltype_enu
 import { Decision } from '../types/reason';
 import { Instructor } from '../types/instructor';
 import { canContactInstructors } from '../../common/courses/contact';
-import { getCourse } from '../util';
+import { Deprecated, getCourse } from '../util';
 
 @ObjectType()
 class Participant {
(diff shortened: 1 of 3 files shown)


graphql/subcourse/fields.ts
```diff
@@ -318,6 +318,23 @@ export class ExtendedFieldsSubcourseResolver {
         });
     }
 
+    @FieldResolver((returns) => [Participant])
+    @Authorized(Role.ADMIN, Role.OWNER)
+    @LimitEstimated(100)
+    async pupilsOnWaitinglist(@Root() subcourse: Subcourse): Promise<Participant[]> {
+        return await prisma.pupil.findMany({
+            select: { id: true, firstname: true, lastname: true, grade: true, schooltype: true, aboutMe: true },
```
LomyW, line 326: which prop can I use to sort the waitling list according to the longest waiting students? Or will I get a sorted list from BE?

LomyW (approved):
Nice!! 🎉

graphql/subcourse/fields.ts
Jonasdoubleyou: Unfortunately this is not possible at the moment. We would have to add a timestamp to the `subcourse_participants_pupil`, but as that is a generated table from TypeORM this is a bit more complicated ... Might address this in a followup change.

PR #514, opened 2023-03-10
Add pupil screenings upon running interest confirmation job

realmayus:
Closes https://github.com/corona-school/project-user/issues/631

### DIFF
--- a/common/match/pool.ts
+++ b/common/match/pool.ts
@@ -486,7 +486,7 @@ export async function confirmationRequestsToSend(pool: MatchPool) {
     const openOffers = Math.max(0, offers + OVERPROVISION_DEMAND - requests);
 
     // If the interest confirmation rate is 10%, we need to ask 100 pupils to get 10 confirmations
-    const confirmationsNeeded = Math.floor(openOffers / (await getInterestConfirmationRate()));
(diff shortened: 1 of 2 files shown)


common/match/pool.ts
```diff
@@ -540,7 +541,22 @@ export async function sendConfirmationRequests(pool: MatchPool) {
     }
 }
 
+// Add a pending pupil screening to every pupil that needs one
+async function addPupilScreenings() {
+    // find all pupils that do not have a valid screening
+    const pupils = await prisma.pupil.findMany({ where: { pupil_screening: { none: { invalidated: false } } } });
```
Jonasdoubleyou, line 547: We don't want to request a pupil screening from everyone?

Jonasdoubleyou: I would propose to 
- add two new toggles analogous to "confirmation-pending", "confirmation-unknown", i.e. "screening-pending" and "screening-unknown"
- Adapt `getPupilsToConfirmInterest` to `getPupilsToContactNext(toggle: "confirmation-unknown" | "screening-unknown")`
- Add a toggle to the Match Pool, analogous to `pool.confirmInterest`, i.e. `pool.needsScreening`. Then we can toggle of confirmInterest, manually request pupil screenings, and once the process switch is done we can toggle on needsScreening and the automatic job takes over

PR #512, opened 2023-03-10
Invalidate pupil screening

realmayus:
Closes https://github.com/corona-school/project-user/issues/626

PR #516, opened 2023-03-10
Invalidate screenings of pupil upon matching

realmayus:
Closes https://github.com/corona-school/project-user/issues/633
This contains some commits from #512 which should be invisible when #512 has been merged.

see https://github.com/corona-school/project-user/issues/633#issuecomment-1464578745

### DIFF
--- a/common/pupil/screening.ts
+++ b/common/pupil/screening.ts
@@ -30,9 +30,11 @@ export async function addPupilScreening(pupil: Pupil, screening: PupilScreeningI
 
 export async function invalidatePupilScreening(screeningId: number) {
     const screening = await prisma.pupil_screening.findUniqueOrThrow({ where: { id: screeningId } });
+
     if (screening.invalidated) {
         throw new RedundantError(`Pupil screening ${screeningId} is already invalidated`);
     }
+
     await prisma.pupil_screening.update({ where: { id: screeningId }, data: { invalidated: true } });
     logger.info(`Invalidated pupil screening ${screeningId} for pupil ${screening.pupilId}`);
 }
--- a/common/pupil/screening.ts
+++ b/common/pupil/screening.ts
@@ -7,6 +7,8 @@ const logger = getLogger('Pupil Screening');
 export async function invalidatePupilScreening(screeningId: number) {
     const screening = await prisma.pupil_screening.findUniqueOrThrow({ where: { id: screeningId } });
     if (screening.invalidated) {
-        throw new RedundantError(`Pupil screening ${screeningId} already is invalidated`);
+        throw new RedundantError(`Pupil screening ${screeningId} is already invalidated`);
     }
+    await prisma.pupil_screening.update({ where: { id: screeningId }, data: { invalidated: true } });
+    logger.info(`Invalidated pupil screening ${screeningId} for pupil ${screening.pupilId}`);
 }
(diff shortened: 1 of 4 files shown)


Jonasdoubleyou (dismissed):

Jonasdoubleyou:
I guess we should also invalidate screenings when the open match request is revoked

graphql/pupil/mutations.ts
```diff
@@ -196,4 +197,11 @@ export class MutatePupilResolver {
 
         return true;
     }
+
+    @Mutation(() => Boolean)
+    @Authorized(Role.ADMIN)
+    async pupilInvalidateScreening(@Arg('pupilScreeningId') pupilScreeningId?: number): Promise<boolean> {
+        await invalidatePupilScreening(pupilScreeningId);
```
dhenkel92, line 204: Would it make sense to always run `invalidateAllScreeningsOfPupil`? Looks like it doesn't hurt and it would make sure that we never have multiple valid entries.

common/pupil/screening.ts
```diff
@@ -27,3 +27,20 @@ export async function addPupilScreening(pupil: Pupil, screening: PupilScreeningI
...
+        throw new RedundantError(`Pupil screening ${screeningId} is already invalidated`);
+    }
+
+    await prisma.pupil_screening.update({ where: { id: screeningId }, data: { invalidated: true } });
+    logger.info(`Invalidated pupil screening ${screeningId} for pupil ${screening.pupilId}`);
+}
+
+// Use this whenever you want to invalidate a pupil's screenings. A pupil should only have one valid screening at most; but we go sure that all are invalidated.
+export async function invalidateAllScreeningsOfPupil(pupilId: number) {
+    const { count } = await prisma.pupil_screening.updateMany({ where: { pupilId }, data: { invalidated: true } });
```
dhenkel92, line 44: This will make sure that we only update the entries that are valid. It doesn't change anything to the logic, but might release some pressure from the db.

```suggestion
    const { count } = await prisma.pupil_screening.updateMany({ where: { pupilId, invalidated: false }, data: { invalidated: true } });
```

graphql/pupil/mutations.ts
realmayus: Yeah, I think that would be a better option; I implemented this function for the GraphQL endpoint because the issue required a way to invalidate a given pupil screening.
@lsc0498 What do you think?

common/pupil/screening.ts
```diff
@@ -70,3 +70,9 @@ export async function invalidatePupilScreening(screeningId: number) {
     await prisma.pupil_screening.update({ where: { id: screeningId }, data: { invalidated: true } });
     logger.info(`Invalidated PupilScreening (${screeningId}) for Pupil (${screening.pupilId})`);
 }
+
+// Use this whenever you want to invalidate a pupil's screenings. A pupil should only have one valid screening at most; but we go sure that all are invalidated.
+export async function invalidateAllScreeningsOfPupil(pupilId: number) {
+    const { count } = await prisma.pupil_screening.updateMany({ where: { pupilId, invalidated: false }, data: { invalidated: true } });
+    logger.info(`Invalidated ${count} pupil screening(s) for pupil ${pupilId}`);
```
Jonasdoubleyou, line 77: Pupil(id)

Jonasdoubleyou (approved):

Jonasdoubleyou (approved):

PR #517, opened 2023-03-11
create update pupil screening mutation

dhenkel92:
closes https://github.com/corona-school/project-user/issues/625

### DIFF
--- a/common/notification/actions.ts
+++ b/common/notification/actions.ts
@@ -38,8 +38,11 @@ const _notificationActions = {
     pupil_screening_add: {
         description: 'Pupil / Screening was added',
     },
-    pupil_screening_update: {
-        description: 'Pupil / Screening was updated',
+    pupil_screening_rejected: {
+        description: 'Pupil / Screening was rejected',
+    },
+    pupil_screening_succeeded: {
+        description: 'Pupil / Screening was successful',
     },
     pupil_registration_finished: {
         description: 'Pupil / Registration finished',
--- a/common/notification/actions.ts
+++ b/common/notification/actions.ts
@@ -38,6 +38,9 @@ const _notificationActions = {
     pupil_screening_add: {
         description: 'Pupil / Screening was added',
     },
+    pupil_screening_update: {
+        description: 'Pupil / Screening was updated',
+    },
     pupil_registration_finished: {
         description: 'Pupil / Registration finished',
     },
--- a/common/pupil/screening.ts
+++ b/common/pupil/screening.ts
@@ -30,8 +30,8 @@ export async function addPupilScreening(pupil: Pupil, screening: PupilScreeningI
 }
 
 interface PupilScreeningUpdate {
-    status: PupilScreeningStatus;
-    comment: string;
+    status?: PupilScreeningStatus;
+    comment?: string;
 }
 
 export async function updatePupilScreening(pupilScreeningId: number, screeningUpdate: PupilScreeningUpdate) {
--- a/common/pupil/screening.ts
+++ b/common/pupil/screening.ts
@@ -35,7 +35,7 @@ interface PupilScreeningUpdate {
 }
 
 export async function updatePupilScreening(pupilScreeningId: number, screeningUpdate: PupilScreeningUpdate) {
-    const screening = await prisma.pupil_screening.findFirst({ where: { id: pupilScreeningId } });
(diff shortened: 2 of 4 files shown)


common/pupil/screening.ts
```diff
@@ -27,3 +28,19 @@ export async function addPupilScreening(pupil: Pupil, screening: PupilScreeningI
...
+
+interface PupilScreeningUpdate {
+    status: PupilScreeningStatus;
+    comment: string;
+}
+
+export async function updatePupilScreening(pupilScreeningId: number, screeningUpdate: PupilScreeningUpdate) {
+    const screening = await prisma.pupil_screening.findFirst({ where: { id: pupilScreeningId } });
+    if (screening === null) {
+        logger.error('cannot find pupil with', { id: pupilScreeningId });
```
Jonasdoubleyou, line 40: "Cannot find PupilScreening(${pupilScreeningId}) for update", we generally use "Entity(id)" to make it simpler to find all logs related to one entity.

common/pupil/screening.ts
```diff
@@ -27,3 +28,19 @@ export async function addPupilScreening(pupil: Pupil, screening: PupilScreeningI
...
+        throw new NotFoundError('pupil screening not found');
+    }
+
+    await prisma.pupil_screening.update({ where: { id: pupilScreeningId }, data: { ...screeningUpdate, updatedAt: new Date() } });
+    logger.debug('successfully updated pupil', { pupilScreeningId, screeningUpdate });
```
Jonasdoubleyou, line 45: Same here "Successfully updated PupilScreening(id)", I guess this should be INFO

graphql/pupil/mutations.ts
```diff
@@ -72,6 +73,15 @@ export class PupilUpdateInput {
     matchReason?: string;
 }
 
+@InputType()
+export class PupilScreeningUpdateInput {
+    @Field(() => PupilScreeningStatus)
```
Jonasdoubleyou, line 78: Do we always want to set both status AND comment? Maybe we should make both optional here (`, { nullable: true }` in the field) so that one can also just set the comment or the status

common/pupil/screening.ts
```diff
@@ -27,3 +28,19 @@ export async function addPupilScreening(pupil: Pupil, screening: PupilScreeningI
...
+}
```
Jonasdoubleyou, line 46: If the status changes to success or rejected we want to send an email to the user, so best already integrate `Notification.actionTaken` here.

Jonasdoubleyou (commented):
Very nice first PR! Some minor improvements to be made

common/pupil/screening.ts
dhenkel92: I'll update it 👍 
But we should revisit the setup when refactoring the logging.

common/notification/actions.ts
```diff
@@ -38,6 +38,9 @@ const _notificationActions = {
     pupil_screening_add: {
         description: 'Pupil / Screening was added',
     },
+    pupil_screening_update: {
```
Jonasdoubleyou, line 41: Maybe two actions instead, one for success and one for rejection? Then one can also add two different notifications ...

dhenkel92: Okay 👍 
I have no idea how that system works

common/pupil/screening.ts
```diff
@@ -27,3 +28,23 @@ export async function addPupilScreening(pupil: Pupil, screening: PupilScreeningI
...
+    const screening = await prisma.pupil_screening.findFirst({ where: { id: pupilScreeningId }, include: { pupil: {} } });
+    if (screening === null) {
+        logger.error(`cannot find PupilScreening(${pupilScreeningId})`);
+        throw new NotFoundError('pupil screening not found');
+    }
+
+    await prisma.pupil_screening.update({ where: { id: pupilScreeningId }, data: { ...screeningUpdate, updatedAt: new Date() } });
+    logger.debug(`successfully updated PupilScreening(${pupilScreeningId})`, { pupilScreeningId, screeningUpdate });
+
+    if (screeningUpdate.status === PupilScreeningStatus.success || screeningUpdate.status === PupilScreeningStatus.rejection) {
```
Jonasdoubleyou, line 47: And I guess this should also check that the previous status was pending, otherwise editing multiple times might accidentally trigger multiple notifications.

dhenkel92: Good catch

Jonasdoubleyou (approved):

PR #518, opened 2023-03-11
fix web logger

dhenkel92:
I noticed the logger was not working during the pupil screening update mutation PR.
The problem is that log4js needs to be configured before running `getLogger`, because otherwise, a new logger with `level: 'OFF'` is created.

This PR will ensure that the first thing in the web/index is setting up the logger so we can use `getLogger` in the root of all the following imports.

We should refactor the usage in the future to have a similar setup to the job/index.

### DIFF
--- a/web/setupLogger.ts
+++ b/web/setupLogger.ts
@@ -1,19 +1,14 @@
 import { isCommandArg } from '../common/util/basic';
 import { configure } from 'log4js';
 
-try {
-    configure({
-        appenders: {
-            out: { type: 'stdout', layout: { type: 'coloured' } },
+configure({
+    appenders: {
+        out: { type: 'stdout', layout: { type: 'coloured' } },
+    },
+    categories: {
+        default: {
+            appenders: ['out'],
+            level: isCommandArg('--debug') ? 'debug' : 'info',
         },
-        categories: {
-            default: {
-                appenders: ['out'],
-                level: isCommandArg('--debug') ? 'debug' : 'info',
-            },
-        },
-    });
-} catch (e) {
-    // TODO: should this be a breaking condition? Loosing all the logs, might be a big problem.
-    console.warn("Couldn't setup logger", e);
-}
+    },
+});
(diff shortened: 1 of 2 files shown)


web/setupLogger.ts
```diff
@@ -0,0 +1,19 @@
+import { isCommandArg } from '../common/util/basic';
+import { configure } from 'log4js';
+
+try {
+    configure({
+        appenders: {
+            out: { type: 'stdout', layout: { type: 'coloured' } },
```
dhenkel92, line 7: I've changed this from `stderr` to `stdout`. Please let me know if this is a problem

Jonasdoubleyou: What are the advantages?

web/setupLogger.ts
```diff
@@ -0,0 +1,19 @@
...
+        },
+        categories: {
+            default: {
+                appenders: ['out'],
+                level: isCommandArg('--debug') ? 'debug' : 'info',
+            },
+        },
+    });
+} catch (e) {
+    // TODO: should this be a breaking condition? Loosing all the logs, might be a big problem.
```
Jonasdoubleyou, line 17: Yes, I think so too, feel free to remove the catch

Jonasdoubleyou (approved):
Nice!

Jonasdoubleyou:
> We should refactor the usage in the future to have a similar setup to the job/index.

Agreed, already started that some time ago but didn't have the time to finish it so far (https://github.com/corona-school/backend/pull/258)

web/setupLogger.ts
dhenkel92: stdout is the default log stream and you could theoretically split error logs into it's own stream, so that you can separate them more easily in the command line.
But in such applications it doesn't make sense to use stderr, because all of them will be shipped to a central place and include a level tag.

web/setupLogger.ts
dhenkel92: Removed 🙂

Jonasdoubleyou (approved):

PR #520, opened 2023-03-14
Make subcourseEdit inputs optional

LomyW:


Jonasdoubleyou:
In https://github.com/corona-school/backend/blob/master/graphql/subcourse/mutations.ts#L160 we should check if maxParticipants is actually changed by the update (I would prefer to not rely on the null to NaN (NaN < something -> false)).

Also while at it, we might want to only call fillSubcourse in https://github.com/corona-school/backend/blob/2d6fe4c189f4cc7fe7d1de417e36e46176d3fb2b/graphql/subcourse/mutations.ts#L164 when the maxParticpants was actually increased

dhenkel92 (approved):

PR #530, opened 2023-03-24
feat: replicate message translations from prod

timidak:
- added new function to replicate message translations
- inspired by `importNotifications`
- tested locally

### DIFF
--- a/common/notification/notification.ts
+++ b/common/notification/notification.ts
@@ -333,7 +333,7 @@ export async function importMessageTranslations(messageTranslations: MessageTran
 
             log += `${untouched} existing MessageTranslations will not be modified\n`;
 
-            for await (const translation of messageTranslations) {
+            for (const translation of messageTranslations) {
                 const templateExists = await prisma.message_translation.findFirst({ where: { id: translation.id } });
 
                 if (templateExists) {
--- a/common/notification/notification.ts
+++ b/common/notification/notification.ts
@@ -2,12 +2,23 @@
    New notifications can be created / modified at runtime. This module contains various utilities to do that */
 
 import { prisma } from '../prisma';
-import { Context, Notification, NotificationID } from './types';
+import { Context, Notification, NotificationID, NotificationMessage } from './types';
 import { NotificationRecipient } from '../entity/Notification';
 import { Prisma } from '@prisma/client';
 import { getLogger } from 'log4js';
 import { hookExists } from './hook';
 import { getNotificationActions } from './actions';
+import { MessageTemplateType } from '../../graphql/types/notificationMessage';
+
+type MessageTranslationFromDb = {
+    template: {
+        body: string;
+        headline: string;
+    };
+    id: number;
+    notificationId: number;
+    navigateTo: string;
+};
 
 type NotificationsPerAction = Map<String, { toSend: Notification[]; toCancel: Notification[] }>;
 let _notificationsPerAction: Promise<NotificationsPerAction>;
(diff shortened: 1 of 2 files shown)


web/dev.ts
```diff
@@ -1405,3 +1406,39 @@ async function importNotificationsFromProd() {
...
+        })
+    ).json();
+
+    let messageTranslations = [];
+
+    prodMessageTranslations.data.notifications.forEach((t) => {
+        if (t.messageTranslations.length > 0) {
+            messageTranslations = [...messageTranslations, ...t.messageTranslations];
+        }
+    });
```
dhenkel92, line 1441: I think this is a perfect use case for the reduce function.

```suggestion
    const messageTranslations = prodMessageTranslations.data.notifications.reduce((acc: any[], cur: any) => [...acc, ...cur.messageTranslations], []);
```

web/dev.ts
```diff
@@ -1405,3 +1406,39 @@ async function importNotificationsFromProd() {
...
+
+    importMessageTranslations(messageTranslations);
```
dhenkel92, line 1443: Should we await this function?

common/notification/notification.ts
```diff
@@ -303,4 +314,57 @@ function diff(prev: any, curr: any, depth = 0) {
     return result;
 }
 
+export async function importMessageTranslations(messageTranslations: MessageTranslationFromDb[], overwrite = false) {
```
dhenkel92, line 317: For me, `overwrite` is a little bit misleading, because it sounds like we are only opting in to overwrite existing entries with new data. But in this case we are erasing all the entries, which is more like a force sync. I like Prisma's naming

```
    synchronize(dropBeforeSync?: boolean): Promise<void>;
```

common/notification/notification.ts
```diff
@@ -303,4 +314,57 @@ function diff(prev: any, curr: any, depth = 0) {
...
+                const removed = await prisma.message_translation.deleteMany({});
+                log += `Through Overwrite ${removed.count} existing MessageTranslations were removed\n`;
+            }
+            const untouched = await prisma.message_translation.count({
+                where: { id: { notIn: messageTranslations.map((it) => it.id) } },
+            });
+
+            log += `${untouched} existing MessageTranslations will not be modified\n`;
+
+            for await (const translation of messageTranslations) {
```
dhenkel92, line 336: I never saw a `for await (...` before 😅 
What is it doing in this case?

common/notification/notification.ts
```diff
@@ -303,4 +314,57 @@ function diff(prev: any, curr: any, depth = 0) {
...
+                    const { id, ...update } = translation;
+                    await prisma.message_translation.update({
+                        data: update,
+                        where: { id: id },
+                    });
+
+                    log += `Updated MessageTranslation(${translation.id})\n`;
+                    log += diff(templateExists, translation, 1);
+                } else {
+                    const result = await prisma.message_translation.create({
```
dhenkel92, line 349: The result is not used.

common/notification/notification.ts
```diff
@@ -303,4 +314,57 @@ function diff(prev: any, curr: any, depth = 0) {
...
+            });
+
+            log += `${untouched} existing MessageTranslations will not be modified\n`;
+
+            for await (const translation of messageTranslations) {
+                const templateExists = await prisma.message_translation.findFirst({ where: { id: translation.id } });
+
+                if (templateExists) {
+                    const { id, ...update } = translation;
+                    await prisma.message_translation.update({
```
dhenkel92, line 341: What about using `prisma.message_translation.upsert` instead of separate create / update branches?

web/dev.ts
timidak: You're totally right! Thanks!

common/notification/notification.ts
timidak: It's basically awaiting the loop and all the async operations in it before beginning the next loop. 
You can think of it as a `Promise.all([].map())` without the parallelization.

common/notification/notification.ts
timidak: Good point 👍🏻 Haven't thought of this. But this would lead to bigger changes for logging and stuff. We couldn't easily distinguish if it was an update or a creation and this might be helpful for debugging. I also want to keep the to functions `importNotifications` and `importMessageTranslations` relatively in sync with each other, so that they are processing everything in "almost" the same way. 

[In this discussion](https://github.com/prisma/prisma/discussions/3432) I found a way how to distinguish, but I think that we should keep the right id's used in production. So, this approach doesn't work for us.

web/dev.ts
Jonasdoubleyou: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat ;)

common/notification/notification.ts
timidak: Thanks for clarifying 💯

Jonasdoubleyou (approved):

Jonasdoubleyou (approved):

PR #522, opened 2023-03-16
[Part 1] Implement webflow sync job

dhenkel92:
This PR is providing a construct to solve the problem described in this slack thread:
https://lern-fair.slack.com/archives/C01548FCGBY/p1678732194485839

The idea is to have a straightforward diff mechanism that triggers a series of API updates and automatically publishes the new data.

I still need some information about the data we want to sync. 
But till then, we should use the time and discuss the overall idea.

PR #531, opened 2023-03-24
[Part 2] Implement webflow sync job

dhenkel92:
I've split the webflow sync into two PRs so that it's easier to review. [Here](https://github.com/corona-school/backend/pull/522) is part 1, which is responsible to sync courses.

Part 2 will focus on the lecture sync as well as the relation of courses to lectures.


dhenkel92:
I've migrated them to graphite, because it's easier to handle them. New PR:

https://github.com/corona-school/backend/pull/535

PR #537, opened 2023-03-26
Turn off interest confirmations for the Lern-Fair Now Pool

Jonasdoubleyou:
I think it makes sense to already deactivate the interest confirmations now so that we have a smaller backlog of pending interest confirmations when pupil screening starts.


Maascha (approved):
LGTM

PR #539, opened 2023-03-26
Sync courses / lectures to Webflow

dhenkel92:


jobs/periodic/sync-to-webflow/sync-courses.ts
```diff
@@ -0,0 +1,163 @@
+import { createNewItem, deleteItems, emptyMetadata, getCollectionItems, publishItems, WebflowMetadata } from './webflow-adapter';
+import { diff, hash, mapDBIdToId, DBIdMap } from './diff';
+import { Logger } from 'log4js';
+import moment, { Moment } from 'moment';
+import { accessURLForKey } from '../../../common/file-bucket';
+import { WebflowSubcourse, getWebflowSubcourses } from './queries';
+import { lectureDTOFactory } from './sync-lectures';
+
+// This is needed so that the weekday will be translated properly.
+moment.locale('de');
```
Jonasdoubleyou, line 10: Does this impact all uses of moment? If yes, can we move this into some higher level file?

jobs/periodic/sync-to-webflow/sync-courses.ts
```diff
@@ -0,0 +1,163 @@
...
+
+const collectionId = process.env.WEBFLOW_COURSE_COLLECTION_ID;
+const lectureCollectionId = process.env.WEBFLOW_LECTURE_COLLECTION_ID;
```
Jonasdoubleyou, line 13: What if these are not set? I would expect an error to be logged, or the job to do nothing (i.e. in the dev environments)

jobs/periodic/sync-to-webflow/index.ts
```diff
@@ -0,0 +1,16 @@
+import syncCourses from './sync-courses';
+import { getLogger } from '../../utils/logging';
+import syncLectures from './sync-lectures';
+
+const logger = getLogger();
```
Jonasdoubleyou, line 5: Please set a topic, `getLogger('SyncToWebflow')`

jobs/periodic/sync-to-webflow/sync-courses.ts
dhenkel92: Yes and good point. We should not set a global value in any file.
I'll move the setting to locals.

jobs/periodic/sync-to-webflow/index.ts
dhenkel92: I'm using the `getLogger` function from the `jobs/utils/logging` file, which doesn't support a topic. Should I rather use the one from log4js?

Jonasdoubleyou (approved):

PR #540, opened 2023-03-26
Match toggle improvements

Jonasdoubleyou:
- Use the same toggles for all pools (which we do anyways)
- By default, show only screened pupils for Lern-Fair Plus matching
- By default, show pupils that were either screened or confirmed their interest for Lern-Fair Now

### DIFF
--- a/common/match/pool.ts
+++ b/common/match/pool.ts
@@ -152,6 +152,70 @@ function formattedSubjectToSubjectWithGradeRestriction(subject: Subject): Subjec
     };
 }
 
+const INTEREST_CONFIRMATION_TOGGLES = ['confirmation-success', 'confirmation-pending', 'confirmation-unknown'] as const;
+type InterestConfirmationToggle = typeof INTEREST_CONFIRMATION_TOGGLES[number];
+
+function addInterestConfirmationFilter(query: Prisma.pupilWhereInput, toggles: string[] | InterestConfirmationToggle[]) {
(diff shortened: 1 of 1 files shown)


common/match/pool.ts
```diff
@@ -533,9 +537,11 @@ export async function predictPupilMatchTime(pool: MatchPool, averageMatchesPerMo
     // as they were not yet asked for an interest confirmation
     // From those we lose about a third of pupils as they do not confirm their interest
     // This needs to be factored in, as it reduces the actual waiting time
-    if (pool.toggles.includes('skip-interest-confirmation')) {
+    if (pool.toggles.includes('confirmation-pending')) {
         backlog +=
-            ((await getPupilCount(pool, ['confirmation-pending'])) + (await getPupilCount(pool, ['confirmation-unknown']))) *
+            ((await getPupilCount(pool, ['confirmation-pending'])) +
+                (await getPupilCount(pool, ['confirmation-pending'])) +
```
realmayus, line 543: Shouldn't this be `confirmation-success`?

realmayus (changes requested):
Nice!

common/match/pool.ts
Jonasdoubleyou: No, the pupils with confirmation-success are already in the backlog, but yes, this line is wrong and I accidentally changed it, good catch!. Adapted it to also calculate pending pupil screenings

realmayus (approved):

PR #548, opened 2023-03-30
User change email

LomyW:


common/entity/Secret.ts
```diff
@@ -36,6 +36,7 @@ export class Secret {
     @Column({ nullable: true })
     lastUsed?: Date;
 
+    // for EMAIL_TOKEN the description is used for new email
```
Jonasdoubleyou, line 39: That comment could be a bit more verbose:

```
// The description depends on the type of the Secret:
// - PASSWORD: currently unused
// - TOKEN: Describes the reason the token was issued, i.e. if it was issued by Support or on which device the token was issued
// - EMAIL_TOKEN the email to which the token was sent
//   - if not set, the user logs in via the email in their account and was not yet authenticated,
//       when they use the token we verify the email and log them in
//   - if set, the user requests to be logged in with a new email. When they log in using the token, we update the email in their profile and also verify it
//      => this MUST only be used by authenticated users, otherwise accounts could be taken over!
```

graphql/secret/mutation.ts
```diff
@@ -39,6 +39,27 @@ export class MutateSecretResolver {
...
+    @Authorized(Role.USER)
+    async meEmailChange(@Arg('userId') userId: string, @Arg('email') email: string) {
+        const user = await getUser(userId);
+        requestToken(user, 'user-email-change', '/start', email);
+        return true;
+    }
+
+    @Mutation((returns) => Boolean)
+    @Authorized(Role.USER)
+    async meChangeEmailVerify(@Ctx() context: GraphQLContext, @Arg('token') token: string) {
```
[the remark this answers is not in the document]
LomyW: where do I call the loginToken in the user-app?

common/secret/token.ts
```diff
@@ -137,6 +150,22 @@ export async function loginToken(token: string): Promise<User | never> {
             await prisma.secret.update({ data: { lastUsed: new Date() }, where: { id: secret.id } });
             logger.info(`User(${user.userID}) logged in with email token Secret(${secret.id}) it will expire at ${secret.expiresAt.toISOString()}`);
         }
+        if (secret.description) {
+            const user = await getUser(secret.userId, /* active */ true);
+            const email = secret.description;
+            if (user.studentId) {
```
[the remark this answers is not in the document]
Jonasdoubleyou: Cause then we can use the function to update any property ({ email }, { firstname, lastname }, ...)

integration-tests/auth.ts
```diff
@@ -81,19 +85,56 @@ test("Token Request", async () => {
...
+test('Change Email', async () => {
+    const {
+        client,
+        pupil: {
+            pupil: { id },
+        },
+    } = await pupilOne;
+    const changedEmailNotification = await createMockNotification('user-email-change', 'EmailChangeNotification');
+
+    await client.requestShallFail(`mutation MeEmailChange { meEmailChange(email: "test+wrong@lern-fair.de")}`);
```
Jonasdoubleyou, line 116: Why should this fail?

integration-tests/auth.ts
```diff
@@ -81,19 +85,56 @@ test("Token Request", async () => {
...
+    const {
+        me: {
+            pupil: { id: id1 },
+        },
+    } = await otherDeviceClient.request(`
+        query CheckLoggedIn {
+            me {
+                pupil {
+                    id
+                    email
```
Jonasdoubleyou, line 135: Here we could assert the changed email

common/user/index.ts
```diff
@@ -266,3 +267,19 @@ export async function queryUser<Select extends UserSelect>(user: User, select: S
 
     throw new Error(`Unknown User(${user.userID})`);
 }
+
+export async function updateUser(userId: string, email?: string, update?: Partial<User>) {
```
Jonasdoubleyou, line 271: { email: string } as second parameter and remove unused update ... We might want to extend this for other properties in the future

Jonasdoubleyou (approved):

PR #562, opened 2023-04-04
Add Decisions canEdit, canCancel, canContactParticipants

Jonasdoubleyou:
That way we can remove the logic that replicates this in the frontend.

Also move implementations to common for future reuse.

### DIFF
--- a/common/courses/states.ts
+++ b/common/courses/states.ts
@@ -6,6 +6,7 @@ import { getCourse } from '../../graphql/util';
 import { sendPupilCourseSuggestion, sendSubcourseCancelNotifications } from '../mails/courses';
 import { fillSubcourse } from './participants';
 import { PrerequisiteError } from '../util/error';
+import { getLastLecture } from './lectures';
 
 const logger = getLogger('Course States');
 
@@ -23,10 +24,7 @@ export async function hasStarted(subcourse: Subcourse) {
 
 
 export async function subcourseOverGracePeriod(subcourse: Subcourse) {
-    const lastLecture = await prisma.lecture.findFirst({
-        where: { subcourseId: subcourse.id },
-        orderBy: { start: 'desc' },
-    });
+    const lastLecture = await getLastLecture(subcourse);
 
     if (!lastLecture) {
         return false;
@@ -39,10 +37,7 @@ export async function subcourseOverGracePeriod(subcourse: Subcourse) {
 }
 
 export async function subcourseOver(subcourse: Subcourse) {
-    const lastLecture = await prisma.lecture.findFirst({
-        where: { subcourseId: subcourse.id },
-        orderBy: { start: 'desc' },
-    });
+    const lastLecture = await getLastLecture(subcourse);
 
     if (!lastLecture) {
         return false;
--- a/common/courses/states.ts
+++ b/common/courses/states.ts
@@ -108,7 +108,7 @@ export async function cancelSubcourse(subcourse: Subcourse) {
 
     await prisma.subcourse.update({ data: { cancelled: true }, where: { id: subcourse.id } });
     const course = await getCourse(subcourse.courseId);
-    sendSubcourseCancelNotifications(course, subcourse);
+    await sendSubcourseCancelNotifications(course, subcourse);
     logger.info(`Subcourse (${subcourse.id}) was cancelled`);
 }
 
--- a/common/courses/states.ts
+++ b/common/courses/states.ts
@@ -35,7 +35,7 @@ export async function subcourseOverGracePeriod(subcourse: Subcourse) {
     const thirtyDaysAgo = new Date();
     thirtyDaysAgo.setDate(thirtyDaysAgo.getDate() - 30);
 
-    return lastLecture.start + lastLecture.duration * 6000 < thirtyDaysAgo;
+    return +lastLecture.start + lastLecture.duration * 6000 < +thirtyDaysAgo;
 }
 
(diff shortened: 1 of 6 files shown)


common/courses/states.ts
```diff
@@ -12,6 +21,39 @@ export async function hasStarted(subcourse: Subcourse) {
...
+    thirtyDaysAgo.setDate(thirtyDaysAgo.getDate() - 30);
+
+    return lastLecture.start < thirtyDaysAgo;
+}
+
+export async function subcourseOver(subcourse: Subcourse) {
+    const lastLecture = await prisma.lecture.findFirst({
+        where: { subcourseId: subcourse.id },
+        orderBy: { start: 'desc' },
+    });
```
dhenkel92, line 45: Maybe this could be moved to common/courses/lectures ?
There is already a `getFirstLecture`.

common/courses/states.ts
```diff
@@ -12,6 +21,39 @@ export async function hasStarted(subcourse: Subcourse) {
     return startedLectures === 0;
 }
 
+
+export async function subcourseOverGracePeriod(subcourse: Subcourse) {
```
dhenkel92, line 25: What is this grace period for?

common/courses/states.ts
```diff
@@ -32,18 +74,80 @@ export async function canPublish(subcourse: Subcourse): Promise<Decision> {
...
+        throw new Error(`Cannot Publish Subcourse(${subcourse.id}), reason: ${can.reason}`);
+    }
 
-    if (!lastLecture) {
-        return false;
+    await prisma.subcourse.update({ data: { published: true, publishedAt: new Date() }, where: { id: subcourse.id } });
+    logger.info(`Subcourse (${subcourse.id}) was published`);
+
+    const course = await getCourse(subcourse.courseId);
+    await sendPupilCourseSuggestion(course, subcourse, 'instructor_subcourse_published');
```
dhenkel92, line 87: Maybe we should add logs to show that the suggestion is executed.

common/courses/states.ts
```diff
@@ -32,18 +74,80 @@ export async function canPublish(subcourse: Subcourse): Promise<Decision> {
...
+
+export async function cancelSubcourse(subcourse: Subcourse) {
+    const can = await canCancel(subcourse);
+    if (!can.allowed) {
+        throw new Error(`Cannot cancel Subcourse(${subcourse.id}), reason: ${can.reason}`);
+    }
+
+    await prisma.subcourse.update({ data: { cancelled: true }, where: { id: subcourse.id } });
+    const course = await getCourse(subcourse.courseId);
+    sendSubcourseCancelNotifications(course, subcourse);
```
dhenkel92, line 116:
```suggestion
    await sendSubcourseCancelNotifications(course, subcourse);
```

graphql/subcourse/fields.ts
```diff
@@ -438,13 +438,31 @@ export class ExtendedFieldsSubcourseResolver {
...
     @FieldResolver((returns) => Decision)
     @Authorized(Role.PARTICIPANT, Role.INSTRUCTOR)
     async canContactInstructor(@Root() subcourse: Required<Subcourse>) {
         const course = await getCourse(subcourse.courseId);
         return await canContactInstructors(course, subcourse);
     }
 
+    @FieldResolver((returns) => Decision)
+    @Authorized(Role.PARTICIPANT, Role.INSTRUCTOR)
+    async canContactParticipants(@Root() subcourse: Required<Subcourse>) {
```
dhenkel92, line 462: Can a participant contact another participant?

common/courses/states.ts
Jonasdoubleyou: After the course ended, participants might want to send feedback to the instructor or ask followup questions, instructors might want to send some final results or promote a next course. So we allow certain course actions for a 'grace period' (terrible name, couldn't come up with something better) Should I add a comment?

common/courses/states.ts
Jonasdoubleyou: True, I have the feeling that we will have to rework the course promotion feature once that spams our users anyways, so :)
Did some minor cleanup and added logging

graphql/subcourse/fields.ts
Jonasdoubleyou: No, but they might want to know whether course instructors could contact them (hypothetically). I'm usually very lose with access restrictions on the can... Resolvers as on some pages it might make sense to do the same query for different types of users without having to do different queries for different roles. 
I guess canInstructorContactParticipants would be a better name, but we already have canContactInstructors so that is kind of symmetrical

common/courses/states.ts
dhenkel92: Thank you for the explanation

> Should I add a comment?

That would be amazing.

graphql/subcourse/fields.ts
dhenkel92: Okay, I see 👍

dhenkel92 (approved):

common/courses/states.ts
```diff
@@ -12,6 +21,39 @@ export async function hasStarted(subcourse: Subcourse) {
...
+        where: { subcourseId: subcourse.id },
+        orderBy: { start: 'desc' },
+    });
+
+    if (!lastLecture) {
+        return false;
+    }
+
+    const now = new Date();
+    return lastLecture.start < now;
```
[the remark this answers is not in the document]
Jonasdoubleyou: Uh, good catch!

PR #563, opened 2023-04-05
add open slots to webflow courses

dhenkel92:
We want to show conditional elements based on the open slots like "ausgebucht" or "weniger als 5 Plätze verfügbar".

Unfortunately, Webflow doesn't support calculations like max-current in these cases, so we have to do it during the sync.


Jonasdoubleyou (approved):

dhenkel92:
By now, I see the Webflow CMS more like a stupid cache to temporarily store the data 😅

PR #564, opened 2023-04-06
Require CoCs for all students

realmayus:
Closes https://github.com/corona-school/project-user/issues/658


common/student/screening.ts
```diff
@@ -65,3 +71,29 @@ export async function ScheduleCoCReminders(student: Student) {
     await createRemissionRequest(student);
     await Notification.actionTaken(student, 'coc_reminder', {});
 }
+
+export async function cancelCoCReminders(student: Student) {
+    const entityManager = getManager();
+    const transactionLog = getTransactionLog();
+
+    const notificationIDs = (
```
Jonasdoubleyou, line 79: I don't think we should replicate the notification system here. I guess it would be sufficient to just trigger another Notification.actionTaken here like 'coc_cancelled', that can then be used in the cancelOnAction on all notifications

graphql/student/mutations.ts
```diff
@@ -246,4 +244,23 @@ export class MutateStudentResolver {
         });
         return getFileURL(file);
     }
+
+    @Mutation(() => Boolean)
+    @Authorized(Role.ADMIN)
+    async studentScheduleCoCReminder(@Arg('studentId') studentId: number) {
```
Jonasdoubleyou, line 250: Hmm maybe 'studentRequestCoC' reflects the external meaning better than 'CoC Reminder'?

Jonasdoubleyou (approved):

realmayus:
@Jonasdoubleyou Would it be a good idea to cancel all CoC requests before sending a new request? That way, we would prevent multiple reminders from existing at the same time, with one possibly deactivating an account before the others have taken place. Or does the notification system already prevent duplicate concrete notifications?

PR #565, opened 2023-04-06
write json formatted logs in non dev environments

dhenkel92:
Right now, most of the logs are multi-line, which decreases the visibility in our monitoring services. Therefore, we should format them as json, so that every log will be represented in a single line.

With this PR a log line will look like that:

```
{"startTime":"2023-04-06T19:00:03.381Z","categoryName":"default","data":[],"level":{"level":20000,"levelStr":"INFO","colour":"green"},"context":{},"pid":88324,"message":"Server listening on port 4000"}
```

### DIFF
--- a/web/setupLogger.ts
+++ b/web/setupLogger.ts
@@ -8,9 +8,9 @@ addLayout('json', function () {
     };
 });
 
-let appenders = ['outJson'];
-if (process.env.ENV === 'dev') {
-    appenders = ['out'];
+let appenders = ['out'];
+if (process.env.LOG_FORMAT === 'json') {
+    appenders = ['outJson'];
 }
 
 configure({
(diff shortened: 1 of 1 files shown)


web/setupLogger.ts
```diff
@@ -1,13 +1,26 @@
...
+
+addLayout('json', function () {
+    return function (logEvent) {
+        const message = logEvent.data.shift();
+        return JSON.stringify({ ...logEvent, message });
+    };
+});
+
+let appenders = ['outJson'];
+if (process.env.ENV === 'dev') {
```
Jonasdoubleyou, line 12: We use ENV = 'dev' for review apps & staging. Maybe we should do something like ENV != local or add a different env to turn this on?

dhenkel92: Ah, good to know. I'll add a different env.

Jonasdoubleyou (approved):

PR #568, opened 2023-04-07
Streamline logging

dhenkel92:
The idea is that we move the variables to either the context or to the additional data section. Otherwise, it's hard to filter for specific messages.


graphql/logging.ts
```diff
@@ -59,5 +59,8 @@ export function logInContext(name: string, context: GraphQLContext) {
         sessionID = 'ADMIN';
     }
 
-    return getLogger(`${name} [${sessionID}]`);
+    const logger = getLogger(name);
+    logger.addContext('session', sessionID);
```
dhenkel92, line 63: This will make it easier to search for the category name & you are still able to filter by session ID

jobs/logconfig.json
dhenkel92, line 1: I moved the contents of this file to `jobs/utils/logging/index.ts`, so that we can react on env variables.

jobs/utils/logging/index.ts
```diff
@@ -1,8 +1,54 @@
...
+if (process.env.LOG_FORMAT === 'json') {
+    stdoutLayout = { type: 'json' };
+}
 
 export function setup() {
     try {
-        configure('jobs/logconfig.json');
+        configure({
+            appenders: {
+                file: {
```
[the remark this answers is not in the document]
dhenkel92: Then I'll use the same configuration as for the web container.

Jonasdoubleyou (approved):

PR #570, opened 2023-04-08
create dedicated logger class and replace log4js in all files

dhenkel92:


common/logger/logger.ts
```diff
@@ -0,0 +1,74 @@
...
+        default: {
+            appenders,
+            level: isCommandArg('--debug') ? 'debug' : 'info',
+        },
+    },
+});
+
+export type LogData = { [key: string]: any };
+
+export class Logger {
```
dhenkel92, line 32: This is a wrapper class to abstract the logging library used. It enables us to use other components in the future.

dhenkel92:
One example is the [graphql query duration](https://app.datadoghq.eu/logs?query=host%3Abackend1%20service%3Aweb.1%20&cols=service&context_event=AYddvFDNAABxZCBSne5KUQAD&event=AgAAAYddvE9y7VpNfAAAAAAAAAAYAAAAAEFZZGR2RkROQUFCeFpDQlNuZTVLVVFBRAAAACQAAAAAMDE4NzVkZDUtZWE2OC00YmExLWFjZWEtZDkyMTQ4Mjc4MzU5&index=%2A&messageDisplay=inline&stream_sort=time%2Cdesc&to_event=AgAAAYddvM3NwZJ9EQAAAAAAAAAYAAAAAEFZZGR2TTdfQUFEV3lnMDc5WWRVaXdBRAAAACQAAAAAMDE4NzVkZDUtZWE2OC00YmExLWFjZWEtZDkyMTQ4Mjc4MzU5&viz=&from_ts=1680904534930&to_ts=1680904894930&live=false):

The path is `.data[0].duration`, which can't be used to create reports based on the information.

jobs/periodic/fetch/utils/verification/index.ts
```diff
@@ -13,7 +13,7 @@ export function generateToken(): string {
         .toString("base64")
         .replace(/\//g, "-")
         .replace(/\+/g, "_");
-    logger.debug("Generated token: ", token);
```
Jonasdoubleyou, line 16: Uh nice, logging credentials 😅  ... This function will hopefully be removed soon, so no need to fix ...

common/logger/logger.ts
```diff
@@ -0,0 +1,74 @@
...
+    constructor(private logger: Log4jsLogger) {}
+
+    getLoggerImpl(): any {
+        return this.logger;
+    }
+
+    trace(message: any, args: LogData = {}): void {
```
Jonasdoubleyou, line 39: Don't we want message to always be a string?

Jonasdoubleyou (commented):
nice!

common/logger/logger.ts
dhenkel92: Good catch!

jobs/periodic/fetch/utils/verification/index.ts
dhenkel92: 😅  I haven't even noticed

common/logger/logger.ts
dhenkel92: I'll have to fix quite a lot of logs now, but it needs to be done 😆

Jonasdoubleyou: We could also do multiple PRs

dhenkel92: Good Idea.
But in this case I would keep it in the same PR even tho it's going to be even more changes, because it's easier to rollback.

common/logger/logger.ts
```diff
@@ -0,0 +1,78 @@
...
+        this.logger.trace(message, args);
+    }
+
+    // TODO switch to string
+    debug(message: any, args: LogData = {}): void {
+        if (typeof message !== 'string') {
+            message = JSON.stringify(message);
+        }
+        this.logger.debug(message, args);
+    }
```
dhenkel92, line 49: There are a looot of debug statements that just provide data without any message.
This is a shortcut for now and we have to check every single one in the future.

Jonasdoubleyou (approved):

PR #582, opened 2023-04-13
add tags attribute to all logs

dhenkel92:
This will add more metadata information in Datadog


common/logger/logger.ts
```diff
@@ -10,7 +10,10 @@ addLayout('json', function () {
         const error = logEvent.context['error'] || {};
         delete logEvent.context['error'];
 
-        return JSON.stringify({ ...logEvent, message, data, error });
+        // These tags will be used to identify the logs in datadog later on
+        const tags = { env: process.env.ENV, version: process.env.VERSION || 'latest' };
+
+        return JSON.stringify({ ...logEvent, message, data, error, tags, service: process.env.SERVICE_NAME });
```
dhenkel92, line 16: @Jonasdoubleyou could you please set a `SERVICE_NAME` environment variable for the api/worker dynos?

Maybe something like
```
SERVICE_NAME=backend-api
SERVICE_NAME=backend-worker
```

common/logger/logger.ts
```diff
@@ -10,7 +10,10 @@ addLayout('json', function () {
         const error = logEvent.context['error'] || {};
         delete logEvent.context['error'];
 
-        return JSON.stringify({ ...logEvent, message, data, error });
+        // These tags will be used to identify the logs in datadog later on
+        const tags = { env: process.env.ENV, version: process.env.VERSION || 'latest' };
```
dhenkel92, line 14: @Jonasdoubleyou Is it possible to set a VERSION environment variable in Heroku (or maybe there is one already)? 
Maybe something around the commit hash or similar, so that we can distinguish logs of different deployments.

common/logger/logger.ts
Jonasdoubleyou: I turned on [Dyno Metadata](https://devcenter.heroku.com/articles/dyno-metadata) so you could use HEROKU_DYNO_ID to distinguish them. Unfortunately we cannot set the ENV per Dyno otherwise. 

Alternatively we could set the service name in package.json, i.e.
```
"web": "SERVICE_NAME=backend-api node ...",
```

common/logger/logger.ts
Jonasdoubleyou: Yes, there is now HEROKU_SLUG_COMMIT (c.f. [Dyno Metadata](https://devcenter.heroku.com/articles/dyno-metadata)). Only turned on in dev for now.

Jonasdoubleyou (approved):
let's try!

PR #586, opened 2023-04-16
Deprovision API for Screening Backend

Jonasdoubleyou:
As we deprovisioned the Screening Backend, all the API exclusively used by it are not needed anymore


web/index.ts
```diff
@@ -276,24 +274,6 @@ createConnection()
             app.use('/api/register', registrationRouter);
         }
 
-        function configureScreenerAPI() {
-            const screenerApiRouter = express.Router();
-            screenerApiRouter.use(screenerAuthCheck);
```
dhenkel92, line 281: do we still need the `screenerAuthCheck`?

Jonasdoubleyou: Good catch, probably not no.

PR #587, opened 2023-04-16
Deprovision old Course API

Jonasdoubleyou:
Replaced by GraphQL that is used by both the Retool and the new User-App


dhenkel92 (approved):
Amazing 😍 
Deleting code is always great

PR #589, opened 2023-04-16
Fix wrong logging of expected errors

Jonasdoubleyou:


Jonasdoubleyou:
I'm pretty confident that this one character change works, so I'll already bring this in. Then we can see tomorrow if it worked and all the errors are silent

PR #595, opened 2023-04-20
send userReportIssue reports to Datadog

dhenkel92:
This PR will send the `userReportIssue` reports to Datadog.

closes https://github.com/corona-school/project-user/issues/688

### DIFF
--- a/graphql/user/mutations.ts
+++ b/graphql/user/mutations.ts
@@ -84,7 +84,17 @@ export class MutateUserResolver {
         section('LOGS', 1);
         result += logs.join('\n');
 
-        logger.info(`An issue was reported from the frontend: \n` + result);
+        const logger = getLogger('IssueReporter');
+        logger.addContext('issureTag', issueTag);
+        logger.addContext('userAgent', userAgent);
+
+        logs.map((log) => logger.info(log));
+        const err: Error = {
+            name: 'IssueReporter',
+            stack: errorStack,
+            message: errorMessage,
+        };
+        logger.error(errorMessage, err);
 
         if (!isDev) {
             await mailjet.sendPure(`Frontend Issue: ${errorMessage}`, result, DEFAULTSENDERS.noreply, 'backend@lern-fair.de', 'Backend', 'Tech-Team');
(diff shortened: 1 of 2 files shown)


graphql/user/mutations.ts
```diff
@@ -84,7 +84,17 @@ export class MutateUserResolver {
         section('LOGS', 1);
         result += logs.join('\n');
 
-        logger.info(`An issue was reported from the frontend: \n` + result);
+        const logger = getLogger('IssueReporter');
+        logger.addContext('issureTag', issueTag);
+        logger.addContext('userAgent', userAgent);
```
dhenkel92, line 89: @Jonasdoubleyou any idea what context might be useful from the `GraphQLContext`?

Jonasdoubleyou: The userID, roles and the sessionID, the sessionID is the same we log in all the GraphQL queries.

graphql/user/mutations.ts
```diff
@@ -27,16 +25,16 @@ class SupportMessage {
...
-    @Mutation((returns) => Boolean)
+    @Mutation(() => Boolean)
     @Authorized(Role.UNAUTHENTICATED)
     @RateLimit('Email Availability', 50 /* requests per */, 5 * 60 * 60 * 1000 /* 5 hours */)
     async isEmailAvailable(@Arg('email') email: string) {
         return await isEmailAvailable(validateEmail(email));
     }
 
-    @Mutation((returns) => String)
+    @Mutation(() => String)
```
Jonasdoubleyou, line 37: using the parameter name to add some meaning to the function is in line with TypeGraphQL's documentation (https://typegraphql.com/docs/resolvers.html) and we followed that pretty consistently. So I would prefer to leave them here.

dhenkel92: I'll add them back, but tbh I don't see the real benefit of having dangling variables for the sake of explanation. Furthermore, it seems to lead to forgotten arguments/variables as the linter cannot care about them right now.

Jonasdoubleyou (approved):

Jonasdoubleyou (approved):

graphql/user/mutations.ts
Jonasdoubleyou: True, also I don't see a reason why it is a function ... Anyways, we can discuss this, but then it should be done consistently across all resolvers

PR #596, opened 2023-04-20
Add Actions for notifying the other user when a match is dissolved

Jonasdoubleyou:


common/match/dissolve.ts
```diff
@@ -39,6 +39,12 @@ export async function dissolveMatch(match: Match, dissolveReason: number, dissol
 
     await Notification.actionTaken(student, 'tutor_match_dissolved', { pupil, matchHash, matchDate, uniqueId });
     await Notification.actionTaken(pupil, 'tutee_match_dissolved', { student, matchHash, matchDate, uniqueId });
+
+    if (dissolver.email === student.email) {
+        await Notification.actionTaken(pupil, 'tutee_match_dissolved_other', { student, matchHash, matchDate, uniqueId });
```
dhenkel92, line 44: What's the difference to the `tutee_match_dissolved` notification above?

Jonasdoubleyou: This one is only triggered for the other person of the match, i.e. when I resolve a match then only my match partner is notified (I don't need to be notified as I triggered the action)

Jonasdoubleyou: No, that depends on the notifications attached to the action.

Jonasdoubleyou: We will probably replace tutee_match_dissolved with tutee_match_dissolved_other in the current notification

dhenkel92: > No, that depends on the notifications attached to the action.

So then it's not a no, but a maybe 😅 



> We will probably replace tutee_match_dissolved with tutee_match_dissolved_other in the current notification

Okay, I see

dhenkel92 (approved):

PR #597, opened 2023-04-23
Fix check if pupil is already a participant

Leo310:
Added missing await and raised the eslint-level for required awaits to error

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,11 +4,7 @@
         "ecmaVersion": 6,
         "sourceType": "module"
     },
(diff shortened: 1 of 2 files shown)


Leo310:
@Jonasdoubleyou maybe raising the linting level to error is not a good idea. There seem to be a lot of places where async isn't needed

Jonasdoubleyou:
agreed, could also do this in a separate PR

Jonasdoubleyou (approved):

PR #599, opened 2023-04-24
feat: notification on new appointment

JenniferFalkenstein:


common/notification/actions.ts
```diff
@@ -314,7 +314,38 @@ const _notificationActions = {
             redirectTo: '/start',
         },
     },
-
+    'student-add-appointment-group': {
```
dhenkel92, line 317: We should stick to the underscore based action naming.

graphql/appointment/mutations.ts
```diff
@@ -85,6 +87,19 @@ export class MutateAppointmentResolver {
         await hasAccessMatch(context, appointment.matchId);
 
         await createMatchAppointment(appointment);
+
+        // send notification
+        if (!context.user) {
```
dhenkel92, line 92: Is it possible to get here without having a user?

LomyW (approved):
nice! 🥳

PR #601, opened 2023-04-25
feat: role authorization for appointments

timidak:
- added new role
- added helper
- refactored public fields


common/appointments/participants.ts
```diff
@@ -0,0 +1,14 @@
+import { Lecture } from '../entity/Lecture';
+import { prisma } from '../prisma';
+import { User } from '../user';
+
+export async function isAppointmentParticipant(lecture: Lecture, user: User): Promise<boolean> {
+    return !!(await prisma.lecture.findFirst({
+        where: {
```
Jonasdoubleyou, line 7: We could select: { id: true } to reduce the amount of data returned from the db

common/entity/Lecture.ts
```diff
@@ -33,4 +42,54 @@ export class Lecture {
 
     @OneToMany((type) => CourseAttendanceLog, (courseAttendanceLog) => courseAttendanceLog.lecture)
     courseAttendanceLog: CourseAttendanceLog[];
+
```
Jonasdoubleyou: rebase

Jonasdoubleyou: (could also leave it like this, the squash commit should look clean anyways)

common/appointments/participants.ts
timidak: Nice one! 👍🏻

Jonasdoubleyou: Yeah one day I request a .exists({ where: { ... } }) feature from Prisma

PR #605, opened 2023-04-26
feat: create and get appointments

timidak:
- add create and get methods for appointments
- add ownership
- add FieldResolver for Appointment
- add FieldResolver for user, subcourse and match


graphql/appointment/fields.ts
```diff
@@ -0,0 +1,188 @@
...
+    @Field((_type) => String, { nullable: true })
+    userId: string;
+    @Field((_type) => Int)
+    id: number;
+    @Field((_type) => String)
+    firstname: string;
+    @Field((_type) => String)
+    lastname: string;
+    @Field((_type) => Boolean)
+    isStudent: boolean;
```
Jonasdoubleyou, line 40: What is the purpose of this field?

graphql/appointment/fields.ts
```diff
@@ -0,0 +1,188 @@
...
+                        in: studentIds,
+                    },
+                },
+                take,
+                skip,
+                select: {
+                    id: true,
+                    firstname: true,
+                    lastname: true,
+                    isStudent: true,
```
Jonasdoubleyou, line 105: What is the meaning of this?

graphql/appointment/fields.ts
```diff
@@ -0,0 +1,188 @@
...
+                        in: pupilIds,
+                    },
+                },
+                take,
+                skip,
+                select: {
+                    id: true,
+                    firstname: true,
+                    lastname: true,
+                    isPupil: true,
```
Jonasdoubleyou, line 123: and this?

graphql/appointment/fields.ts
```diff
@@ -0,0 +1,188 @@
...
+        return isParticipant;
+    }
+    @FieldResolver((returns) => [AppointmentParticipant], { nullable: true })
+    @Authorized(Role.OWNER, Role.APPOINTMENT_PARTICIPANT)
+    @LimitEstimated(30)
+    async participants(@Root() appointment: Appointment, @Arg('take', (type) => Int) take: number, @Arg('skip', (type) => Int) skip: number) {
+        const studentIds = [];
+        const pupilIds = [];
+        appointment.participantIds.forEach((userId) => {
+            const [type, id] = getUserTypeAndIdForUserId(userId);
```
Jonasdoubleyou, line 84: Could we move this whole thing to common/user.ts? Participant and Organizer should both be subsets of User, so we could just do `getUsers(userIds).map(it => { ...subset })` here

graphql/appointment/fields.ts
```diff
@@ -0,0 +1,188 @@
...
+            })
+        ).map((p) => ({ ...p, isPupil: true, userId: getUserIdTypeORM(p) }));
+        return [...studentParticipants, ...pupilParticipants];
+    }
+
+    @FieldResolver((returns) => [Organizer])
+    @Authorized(Role.USER)
+    @LimitEstimated(5)
+    async organizers(@Root() appointment: Appointment, @Arg('take', (type) => Int) take: number, @Arg('skip', (type) => Int) skip: number) {
+        const organizerStudentIds = appointment.organizerIds.map((userId) => parseInt(userId.split('/')[1]));
```
Jonasdoubleyou, line 134: Same here, could we use a generic `getUsers` ?

graphql/appointment/fields.ts
timidak: I wasn't 100% sure if somebody needs this in the frontend. But I just had a look. I'll remove it.

graphql/appointment/fields.ts
timidak: I'm selecting this because I need it in line 108 for `getUserIdTypeORM` so that it can create the correct userId

graphql/appointment/fields.ts
timidak: Same here, line 126

graphql/appointment/fields.ts
Jonasdoubleyou: Aaah, well, I mean we know that it will be "student" no? So I think we should move this to common/user and directly `.map` to the User with the correct userId

common/appointment/create.ts
```diff
@@ -0,0 +1,131 @@
...
+            duration: appointmentToBeCreated.duration,
+            subcourseId: appointmentToBeCreated.subcourseId,
+            appointmentType: lecture_appointmenttype_enum.group,
+            organizerIds: instructors.map((i) => getUserForTypeORM(i.student).userID),
+            participantIds: participants.map((p) => getUserForTypeORM(p.pupil).userID),
+        },
+    });
+};
+
+export const createGroupAppointments = async (subcourseId: number, appointmentsToBeCreated: AppointmentCreateGroupInput[]) => {
```
dhenkel92, line 110: What's the purpose of having two different functions? Couldn't we use `createGroupAppointments`  all the time?

common/appointment/get.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    if (direction === 'next' && cursor) {
+        appointments = await getNextAppointmentsByIdList(user.userID, take, cursor);
+    }
+
+    if (direction === 'last' && cursor) {
+        appointments = await getLastAppointmentsByIdList(user.userID, take, cursor);
+    }
+
+    return appointments;
```
dhenkel92, line 30: I think we can just return `getLastAppointmentsByIdList(user.userID, take, cursor);` here, because direction only has three states where two are covered already.

common/appointment/get.ts
```diff
@@ -0,0 +1,103 @@
...
+                },
+            ],
+        },
+        orderBy: [{ start: 'asc' }],
+        take,
+        skip: 1, // Skipping the cursor object
+        cursor: { id: cursor },
+    });
+};
+const getLastAppointmentsByIdList = async (userId: User['userID'], take: number, cursor: number) => {
```
dhenkel92, line 56: I think we can merge `getLastAppointmentsByIdList` and `getNextAppointmentsByIdList` by introducing another argument.

common/appointment/get.ts
```diff
@@ -0,0 +1,103 @@
...
+        },
+        orderBy: [{ start: 'desc' }],
+        take,
+        skip: 1, // Skipping the cursor object
+        cursor: { id: cursor },
+    });
+    return appointments.reverse();
+};
+
+const getAppointmentsByIdList = async (userId: User['userID'], take, skip): Promise<Appointment[]> => {
```
dhenkel92, line 81: take and skip don't have types

common/appointment/get.ts
```diff
@@ -0,0 +1,103 @@
...
+    return await prisma.lecture.findMany({
+        where: {
+            isCanceled: false,
+            start: { gte: new Date().toISOString() },
```
dhenkel92, line 85: Why do we care about the start date here, but not in the other queries?

common/migration/1682516060693-suffix-participants-and-organizers-with-ids.ts
```diff
@@ -0,0 +1,20 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+
+export class suffixParticipantsAndOrganizersWithIds1682516060693 implements MigrationInterface {
+    name = 'suffixParticipantsAndOrganizersWithIds1682516060693';
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "lecture" DROP COLUMN "organizers"`);
+        await queryRunner.query(`ALTER TABLE "lecture" DROP COLUMN "participants"`);
```
dhenkel92, line 8: Are these new columns or do they already have data in prod?

graphql/appointment/fields.ts
```diff
@@ -0,0 +1,188 @@
...
+export class ExtendedFieldsLectureResolver {
+    @Query((returns) => Appointment)
+    @AuthorizedDeferred(Role.OWNER, Role.APPOINTMENT_PARTICIPANT)
+    async appointment(@Ctx() context: GraphQLContext, @Arg('appointmentId') appointmentId: number) {
+        const appointment = await prisma.lecture.findUniqueOrThrow({ where: { id: appointmentId } });
+        await hasAccess(context, 'Lecture', appointment);
+        return appointment;
+    }
+
+    @Deprecated('use isOrganizer instead')
```
dhenkel92, line 53: Should we introduce a new query which is deprecated already?

common/appointment/create.ts
```diff
@@ -0,0 +1,131 @@
...
+            duration: appointmentToBeCreated.duration,
+            matchId: appointmentToBeCreated.matchId,
+            appointmentType: lecture_appointmenttype_enum.match,
+            organizerIds: [studentUserId],
+            participantIds: [pupilUserId],
+        },
+    });
+};
+
+export const createMatchAppointments = async (matchId: number, appointmentsToBeCreated: AppointmentCreateMatchInput[]) => {
```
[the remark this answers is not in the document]
timidak: I guess from the UX perspective it's better to have different endpoints for different use cases. It is planned to send different notifications if a single appointment or more appointments are created.

common/appointment/create.ts
```diff
@@ -0,0 +1,131 @@
...
+    @Field(() => Int, { nullable: false })
+    subcourseId: number;
+    @Field(() => lecture_appointmenttype_enum)
+    appointmentType: 'group';
+}
+
+export const createMatchAppointment = async (appointmentToBeCreated: AppointmentCreateMatchInput) => {
+    const match = await prisma.match.findUniqueOrThrow({ where: { id: appointmentToBeCreated.matchId } });
+    const student = await prisma.student.findUniqueOrThrow({ where: { id: match.studentId } });
+    const pupil = await prisma.pupil.findUniqueOrThrow({ where: { id: match.pupilId } });
```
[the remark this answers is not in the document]
timidak: How? 🤔

common/appointment/create.ts
timidak: > I guess from the UX perspective it's better to have different endpoints for different use cases. It is planned to send different notifications if a single appointment or more appointments are created.

Same here

common/appointment/get.ts
timidak: I'm not sure if I get the question correctly. 
Because this query is meant to give us `x = take` appointments from now and the other ones give us `x`  appointments from a specific cursor in a specific direction. So we're [paginating cursor based](https://www.prisma.io/docs/concepts/components/prisma-client/pagination#cursor-based-pagination) to be able to scroll upwards and downwards infinitely in a scrollable list.

common/migration/1682516060693-suffix-participants-and-organizers-with-ids.ts
timidak: They shouldn't. I added them with a PR yesterday and they shouldn't be currently in use.

graphql/appointment/fields.ts
timidak: This is an old resolver, but I renamed the file from lectures to appointments.

common/appointment/create.ts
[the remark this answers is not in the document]
dhenkel92: You can join the students and pupils in the match query.
```
    const match = await prisma.match.findUniqueOrThrow({ where: { id: appointmentToBeCreated.matchId }, include: { student: true, pupil: true } });
```

graphql/appointment/fields.ts
dhenkel92: Okay I see.
Interesting that github doesn't show it 😅

common/appointment/get.ts
dhenkel92: Tbh I never worked with cursors, so I'm somewhat confused by the function naming, but the concept looks interesting. 

Could you give me an example how the three functions will be used together?

common/appointment/create.ts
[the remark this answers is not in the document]
dhenkel92: But here we are talking about a different layer. We can still have different graphql endpoints, but the logic will be the same.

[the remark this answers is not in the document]
timidak: Ok, now I get your point! Totally right, will refactor it.

common/appointment/create.ts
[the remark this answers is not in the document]
timidak: Nice one! ty 💯

common/appointment/get.ts
timidak: Sure:
Imagine a list of tiles (a list of appointments is a pretty good use case). First, I want the upcoming appointments that start in the future from "now". So we query 10 items from now. When I reach the bottom of the list, the next 5 appointments after the last appointment (`cursor: lastId, direction: 'next', take: 5`) are queried and merged together with the current list. But this is not the point, because this can be achieved with a offset pagination. But if I now scroll to the top of the list to get to the past appointments, I use the same query again, but with different parameters (`cursor: firstIdInList, direction: 'last', take: 5` ), and I can merge the queryResult with the current list again. 

Did it make it clearer?

timidak: > I'm somewhat confused by the function naming [...]

The naming was a copy and paste artifact 😄 I'll refactor that.

graphql/subcourse/fields.ts
```diff
@@ -468,4 +468,10 @@ export class ExtendedFieldsSubcourseResolver {
     async capacity(@Root() subcourse: Required<Subcourse>) {
         return getCourseCapacity(subcourse);
     }
+
+    @FieldResolver((returns) => [Lecture])
+    @Authorized(Role.UNAUTHENTICATED)
+    async appointments(@Root() subcourse: Subcourse) {
+        return await prisma.lecture.findMany({ where: { subcourseId: subcourse.id } });
```
LomyW, line 475: should return ordered appointments

graphql/subcourse/fields.ts
```diff
@@ -468,4 +468,10 @@ export class ExtendedFieldsSubcourseResolver {
     async capacity(@Root() subcourse: Required<Subcourse>) {
         return getCourseCapacity(subcourse);
     }
+
+    @FieldResolver((returns) => [Lecture])
+    @Authorized(Role.UNAUTHENTICATED)
```
LomyW, line 473:
```suggestion
    @Authorized(Role.ADMIN, Role.OWNER)
```

common/appointment/create.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+export const createMatchAppointments = async (matchId: number, appointmentsToBeCreated: AppointmentCreateMatchInput[]) => {
+    const { pupil, student } = await prisma.match.findUniqueOrThrow({ where: { id: matchId }, include: { student: true, pupil: true } });
+    const studentUserId = getUserIdTypeORM(student);
+    const pupilUserId = getUserIdTypeORM(pupil);
+    return await Promise.all(
+        appointmentsToBeCreated.map(
+            async (appointmentToBeCreated) =>
+                await prisma.lecture.create({
```
Jonasdoubleyou, line 45: Maybe createMany?

Jonasdoubleyou (approved):
nice! Looking forward to the integration tests :)

common/appointment/create.ts
timidak: @LomyW could you bring this change in, in the integration test branch? Would be a charm 🥇

PR #609, opened 2023-04-26
Feat 666

dhenkel92:


common/secret/token.ts
```diff
@@ -105,7 +106,9 @@ export async function requestToken(
 
 // The token returned by this function MAY NEVER be persisted and may only be sent to the user by email
 // If newEmail ist set, the token MUST be sent to that new email
-export async function createSecretEmailToken(user: User, newEmail?: string) {
+export async function createSecretEmailToken(user: User, description?: string): Promise<string>;
+export async function createSecretEmailToken(user: User, description?: string, expiresAt?: Moment): Promise<string>;
```
Jonasdoubleyou, line 110: Are these overloads needed?

Jonasdoubleyou: Also for the EMAIL_TOKEN we 'abuse' the description field to change the email, thus I prefer to keep the name pf the parameter 'newEmail'

dhenkel92: > Are these overloads needed?

Depends. Theoretically, we are using `createSecretEmailToken` only for user actions right now, so we could add a default TTL there as well. Any suggestion?

> Also for the EMAIL_TOKEN we 'abuse' the description field to change the email, thus I prefer to keep the name pf the parameter 'newEmail'

Does this make sense? We will call the function for many different notifications where only one is using the description to store the new email. It doesn't feel right to make the "abuse" the default.

PR #611, opened 2023-04-27
feat: add displayName field resolver for appointments

timidak:
- added resolver
- removed unnecessary TODO comments


graphql/appointment/fields.ts
```diff
@@ -119,5 +117,33 @@ export class ExtendedFieldsLectureResolver {
...
+                    isOrganizer = false;
+                } else {
+                    isOrganizer = (await prisma.lecture.count({ where: { id: appointment.id, organizerIds: { has: context.user.userID } } })) > 0;
+                }
+
+                if (isOrganizer) {
+                    const [tutee] = await getUsers(appointment.participantIds);
+                    return `${tutee.firstname} ${tutee.lastname}`;
+                } else {
+                    const [tutor] = await getUsers(appointment.organizerIds);
```
Jonasdoubleyou, line 137: Do match appointments always only have one organizer and one participant?

Jonasdoubleyou (approved):

graphql/appointment/fields.ts
timidak: Yes, this may change in the future, but for now 1on1 learning support have one organizer and one participant.

PR #612, opened 2023-04-28
test: appointment queries

LomyW:
integration tests for: 
- creating appointments
- query my appointments
- cancel an appointment
- update an appointment


common/appointment/create.ts
```diff
@@ -42,7 +42,7 @@ export const createMatchAppointments = async (matchId: number, appointmentsToBeC
     return await Promise.all(
         appointmentsToBeCreated.map(
             async (appointmentToBeCreated) =>
-                await prisma.lecture.create({
+                await prisma.lecture.createMany({
```
Jonasdoubleyou, line 45: If createMany is used, this shouldn't be in a .map(...)

integration-tests/appointments.ts
```diff
@@ -0,0 +1,172 @@
...
+                        lastname
+                    }
+                    declinedBy
+                }
+            }
+        }
+    `);
+
+    assert.ok(appointments);
+    assert.ok(appointments.some((a) => (a.title = appointmentTitle)));
```
Jonasdoubleyou, line 99: === instead of =

Jonasdoubleyou:
tests fail?

Jonasdoubleyou (approved):

PR #613, opened 2023-04-28
fix: make declinedBy a public field

LomyW:


graphql/authorizations.ts
```diff
@@ -428,21 +428,21 @@ export const authorizationModelEnhanceMap: ModelsEnhanceMap = {
...
-                match: adminOrOwner,
-                matchId: adminOrOwner,
-                declinedBy: adminOrOwner,
-                participantIds: adminOrOwner,
-                organizerIds: adminOrOwner,
-            }
-        ),
+        fields: withPublicFields<
+            Lecture,
+            'id' | 'start' | 'duration' | 'createdAt' | 'updatedAt' | 'title' | 'description' | 'appointmentType' | 'isCanceled' | 'declinedBy'
```
Jonasdoubleyou, line 433: Why is participantIds private but declinedBy is not?

LomyW: For the `AttendeesModal` in the appointment detail view we need the names of the attendees (field resolver for participants and organizers) and also the information, who has declined (declinedBy array from DB), to show it in the modal.

timidak: True, that was my bad, @Jonasdoubleyou! 
Is it possible to check `APPOINTMENT_PARTICPANT` without a custom field resolver?
So my question is can we annotate the auto generated resolvers with `@AuthorizedDeferred`?

Jonasdoubleyou: Yes, @LomyW was faster ;) #622

PR #621, opened 2023-05-04
Feat: Validate appointment start date

LomyW:


common/appointment/create.ts
```diff
@@ -35,13 +36,41 @@ export abstract class AppointmentCreateGroupInput extends AppointmentCreateInput
...
+    if (diffDays >= 6) {
+        return true;
+    }
+    return false;
+};
+
+const isAppointmentFiveMinutesLater = (datestring: Date) => {
+    const now = moment.now();
+    const start = moment(datestring);
+    const same = start.isSame(now, 'date');
```
timidak, line 52:
```suggestion
    const isSameDay = start.isSame(now, 'date');
```

common/appointment/create.ts
```diff
@@ -35,13 +36,41 @@ export abstract class AppointmentCreateGroupInput extends AppointmentCreateInput
...
+    const after = start.isAfter(now);
+    if (same) {
+        const diff = start.diff(now, 'minutes');
+        if (diff >= 5) {
+            return true;
+        }
```
timidak, line 58: The moment the request is hitting the api it might be only 4 minutes?
I'm not sure how to handle that correctly, but going with 5 minutes fails I guess, if we're sending the request to a remote server.

common/appointment/create.ts
```diff
@@ -35,13 +36,41 @@ export abstract class AppointmentCreateGroupInput extends AppointmentCreateInput
...
 export const createMatchAppointments = async (matchId: number, appointmentsToBeCreated: AppointmentCreateMatchInput[]) => {
     const { pupil, student } = await prisma.match.findUniqueOrThrow({ where: { id: matchId }, include: { student: true, pupil: true } });
     const studentUserId = getUserIdTypeORM(student);
     const pupilUserId = getUserIdTypeORM(pupil);
+
     return await Promise.all(
-        appointmentsToBeCreated.map(
-            async (appointmentToBeCreated) =>
+        appointmentsToBeCreated.map(async (appointmentToBeCreated) => {
+            if (isAppointmentFiveMinutesLater(appointmentToBeCreated.start)) {
```
timidak, line 73: This might result in creating some of the appointments, but not all appointments?
Another point is, that we only have to check the first one, if they are sorted correctly.

common/appointment/create.ts
```diff
@@ -35,48 +36,64 @@ export abstract class AppointmentCreateGroupInput extends AppointmentCreateInput
     appointmentType: 'group';
 }
 
+export const isAppointmentOneWeekLater = (appointmentDate: Date) => {
+    const now = moment.now();
+    const start = moment(appointmentDate);
+    const diffDays = start.diff(now, 'days');
+    return diffDays >= 6 ? true : false;
```
dhenkel92, line 43: Shouldn't it be greater than six?

```suggestion
    return diffDays > 6;
```

common/appointment/create.ts
```diff
@@ -35,48 +36,64 @@ export abstract class AppointmentCreateGroupInput extends AppointmentCreateInput
...
+};
+
+export const isAppointmentFiveMinutesLater = (appointmentDate: Date) => {
+    const now = moment.now();
+    const start = moment(appointmentDate);
+    const fiveMinutesFromNow = moment().add(5, 'minutes');
+    const isSameDayAndFiveMinLater = start.isSame(now, 'day') && start.isAfter(fiveMinutesFromNow);
```
dhenkel92, line 50: Any special reason why it has to be on the same day?

LomyW: because a match appointment can now also be created for now + 5 minutes 😄 but if the date is not today, than I don't need to check if the time ist 5 minutes later than now

Jonasdoubleyou: Well this could just be start > fiveMinutesFromNow, if you compare the whole date instead of manually looking at minutes and days ...

PR #625, opened 2023-05-06
Use EMAIL_TOKEN for mail auth tokens

dhenkel92:


common/notification/index.ts
```diff
@@ -223,7 +223,7 @@ export async function getMessage(
         type,
         body: renderTemplate(body, context),
         headline: renderTemplate(headline, context),
-        navigateTo: renderTemplate(navigateTo, context),
+        navigateTo,
```
Jonasdoubleyou, line 226: faulty merge, this should stay

Jonasdoubleyou (commented):
Looks good! Maybe as a followup we could add some secret statistics to https://github.com/corona-school/backend/blob/master/graphql/statistics/fields.ts to get a better visibility into the whole login process

common/notification/index.ts
dhenkel92: Very good catch!

Jonasdoubleyou (approved):

PR #627, opened 2023-05-07
refactor: remove authToken field

dhenkel92:
To old authToken was replaced by the new EMAIL_TOKEN secret. This commit is removing the authToken and all other related fields


web/controllers/tokenController/index.ts
dhenkel92, line 1: I'm not sure if it's okay to remove these. Afaics they've only been used in the `/api/token` REST api, which should be retired anyways, so I decided to delete the code instead of refectoring it.

web/controllers/utils.ts
```diff
@@ -52,6 +51,6 @@ export function checkExpertises(expertises: string[]) {
 
 /* Creates a link which logs in as a certain user and then goes to the path in the frontend
    ATTENTION: Exposing this to someone else than the user is not a good idea! */
-export function createAutoLoginLink(user: Person, path: string) {
-    return `https://my.lern-fair.de/login?token=${user.authToken}&path=${encodeURIComponent(path)}`;
+export function createAutoLoginLink(token: string, path: string) {
+    return `https://my.lern-fair.de/login?token=${token}&path=${encodeURIComponent(path)}`;
```
Jonasdoubleyou, line 55: app.lern-fair.de

web/middleware/auth.ts
```diff
@@ -29,23 +36,22 @@ export function authCheckFactory(optional = false, useQueryParams = false, loadE
                 where: [
                     {
                         authToken: hashToken(token),
```
Jonasdoubleyou, line 38: How can we use a dropped field here?

Jonasdoubleyou: I think we can remove the auth check factory entirely, as all remaining REST Apis should be unauthenticated / use the GraphQL Session

dhenkel92: Good Catch!

dhenkel92: > How can we use a dropped field here?

Typescript is never as safe as it promises 😞 
The where attibute includes the `ObjectLiteral` which basically accapts any object
```
    where?: FindConditions<Entity>[] | FindConditions<Entity> | ObjectLiteral | string;

    export interface ObjectLiteral {
      [key: string]: any;
    }
```

> as all remaining REST Apis should be unauthenticated / use the GraphQL Session

What you mean is that for example the user API is getting/validating the user by it's session already? So I could just remove the `authCheckFactory` middleware everywhere?

Jonasdoubleyou: No, we could just remove the user API.

dhenkel92: Then I would propose to keep it as it is for now and remove the old apis in different commits. 
The broken `authCheckFactory` will make sure that no user data will be leaked until then.

Jonasdoubleyou: Mm can we explicitly disable the authCheckFactory? Not a fan of broken authentication

dhenkel92: Done

Jonasdoubleyou (approved):

PR #628, opened 2023-05-07
Fix undefined bug in MessageTranslations

realmayus:
Closes https://github.com/corona-school/project-user/issues/714

### DIFF
--- a/utils/helpers.ts
+++ b/utils/helpers.ts
@@ -3,7 +3,9 @@ import { Context } from '../common/notification/types';
 import { getLogger } from '../common/logger/logger';
 
 export const renderTemplate = (template: string, context: Partial<Context>, strict: boolean = false) => {
+    const log = getLogger('Template Rendering');
     if (!template) {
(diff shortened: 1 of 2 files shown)


Jonasdoubleyou (dismissed):

utils/helpers.ts
```diff
@@ -1,4 +1,10 @@
 import { compile } from 'handlebars';
 import { Context } from '../common/notification/types';
+import { getLogger } from '../common/logger/logger';
 
-export const renderTemplate = (template: string, context: Partial<Context>, strict: boolean = false) => compile(template, { strict })(context);
+export const renderTemplate = (template: string, context: Partial<Context>, strict: boolean = false) => {
+    if (!template) {
+        return undefined;
```
dhenkel92, line 7: Can the template be empty by exception or is the system designed like that?
If it's an exception I would propose to add a log message to keep track of it.

dhenkel92 (approved):

PR #632, opened 2023-05-07
Session Logging

Jonasdoubleyou:
Introduces the concept of a Session (i.e. a User-App Session) with multiple Transactions (i.e. GraphQL Queries) and uses AsyncLocalStorage to retain the transaction across continuations. The transactionID is attached to every log line to better trace logs across components.

resolves https://github.com/corona-school/project-user/issues/702

### DIFF
--- a/common/logger/logger.ts
+++ b/common/logger/logger.ts
@@ -1,5 +1,6 @@
 import { isCommandArg } from '../util/basic';
 import { configure, addLayout, getLogger as getlog4jsLogger, Logger as Log4jsLogger } from 'log4js';
+import { getCurrentTransaction } from '../session';
 
 addLayout('json', function () {
     return function (logEvent) {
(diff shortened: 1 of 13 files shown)


common/logger/logger.ts
```diff
@@ -24,7 +25,22 @@ if (process.env.LOG_FORMAT === 'json') {
 
 configure({
     appenders: {
-        out: { type: 'stdout', layout: { type: 'coloured' } },
+        out: {
+            type: 'stdout',
+            layout: {
+                type: 'pattern',
+                pattern: '%[%x{transaction}%c%] %m%n',
```
Jonasdoubleyou, line 32: We can discuss which format we configure here, this is only used for local development anyways

graphql/me/mutation.ts
```diff
@@ -225,6 +225,8 @@ class BecomeStatePupilInput implements BecomeStatePupilData {
     gradeAsInt?: number;
 }
 
+const logger = getLogger('Me');
```
dhenkel92, line 228: To keep it consistent.
```suggestion
const logger = getLogger('Me Mutations');
```

dhenkel92 (dismissed):
🤩

graphql/logging.ts
```diff
@@ -31,10 +32,9 @@ export const GraphQLLogger: any = {
                 if (!unexpected) {
                     logger.info(`Expected Errors occurred`, { errors: requestContext.errors.map((it) => `  - ${it.name} (${it.message})`) });
                 } else {
-                    const errorLogger = logInContext(`GraphQL Error`, requestContext.context as GraphQLContext);
-                    errorLogger.addContext('uid', uid);
+                    logger.addContext('uid', uid);
```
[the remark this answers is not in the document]
Jonasdoubleyou: Yeah but were in a callback, so probably we need to set the context again? (Also we probably need to clean up the context more often)

dhenkel92 (approved):

PR #633, opened 2023-05-08
feat: graphql mutations skeleton for chat functionality

JenniferFalkenstein:


graphql/me/fields.ts
```diff
@@ -3,7 +3,19 @@ import { getSessionUser, GraphQLUser } from '../authentication';
 import { GraphQLContext } from '../context';
 import { Role } from '../authorizations';
 import { UserType } from '../types/user';
+import { Field, ObjectType } from 'type-graphql';
 
+@ObjectType()
+class Contact {
```
Jonasdoubleyou, line 9: Can't we use the User (UserType) here?

LomyW: I think that the `Contact` type would maybe extended by fields e.g. `contactReason`, if we add a own `contact` table.

Jonasdoubleyou: Well how about:

class Contact {
  user: UserType;
  contactReason: ...;
  chatId: string;
}

graphql/me/fields.ts
```diff
@@ -17,4 +27,16 @@ export class FieldMeResolver {
     myRoles(@Ctx() context: GraphQLContext): string[] {
         return context.user?.roles ?? [];
     }
+
+    @Query((returns) => [String])
```
Jonasdoubleyou, line 31: [Contact] ?

Jonasdoubleyou (approved):

PR #634, opened 2023-05-09
feat: talkjs service functions

JenniferFalkenstein:


common/chat/helper.ts
```diff
@@ -0,0 +1,16 @@
+const parseSlashToUnderscore = (userId: string): string => {
```
Jonasdoubleyou, line 1: I'd prefer a name that conveys the semantic meaning, i.e. userIdToTalkJS(...)

common/chat/user.ts
```diff
@@ -0,0 +1,88 @@
+/* eslint-disable camelcase */
+import { GraphQLUser } from '../user/session';
+import dotenv from 'dotenv';
+import { checkResponseStatus, parseSlashToUnderscore } from './helper';
+
+dotenv.config();
+
+/**
+ * NOTE: `id` ist not the same as `userId`! It's a transformed version, e.g.
```
Jonasdoubleyou, line 9: It would also be good to document the 'why' ... "as TalkJS' id field must not contain slashes"

graphql/me/fields.ts
```diff
@@ -17,4 +18,12 @@ export class FieldMeResolver {
     myRoles(@Ctx() context: GraphQLContext): string[] {
         return context.user?.roles ?? [];
     }
+
+    @Query((returns) => String)
```
Jonasdoubleyou, line 22: Could also be a FieldResolver on me

Jonasdoubleyou (commented):
Looks good, but the unit tests aren't really unit tests if they need to talk to TalkJS. We could either mock on the HTTP layer or disable the tests for now

Jonasdoubleyou (dismissed):

Jonasdoubleyou (approved):

PR #642, opened 2023-05-15
Shutdown unused APIs

Jonasdoubleyou:
resolves https://github.com/corona-school/project-user/issues/689

- Project Experts was a project that was stopped a while ago and has not made it to the new User App
- Interest Confirmations now go through the GraphQL stack (with an important message in the User App)
- Match dissolve now goes through GraphQL
- Project Matching was a project that was stopped a while ago and has not made it to the new User App (but partial support exists in GraphQL)
- Registration / User Settings now goes through the GraphQL stack


Leo310 (approved):
Noice, removing legacy stuff is the best

realmayus (approved):
Thank you! 🎉

PR #645, opened 2023-05-16
feat(file-store): remove tmp-files after successful download and processing

Leo310:
Resolves backend part of https://github.com/corona-school/project-user/issues/731

I think we could also increase the upload filesize from 5mb to 10mb.


Jonasdoubleyou (dismissed):
nice!

Jonasdoubleyou:
Sure, lets increase it tp something more realistic, i.e. 50MB

Leo310:
Do you mean 50mb per file? Then we would need a capacity of 2.5gb only for file storing. We don't have that. I thought about storing 50*10mb files -> max 500mb

PR #648, opened 2023-05-18
Entity roles

Jonasdoubleyou:
Introduce the concept of "entity roles" that are evaluated per entity, cache the role once it was evaluated (so that it can be reused when checking permissions for multiple field resolvers on the same entity). Also add the `@ImpliesRoleOnResult` decorator to skip entity role checks when the role is already clear from the context.

-----

When querying:

```gql
query { 
   me { 
      pupil { 
         subcoursesJoined { # imply role on result
            otherParticipants { firstname } # check role from cache
         } 
      } 
   }
}
```

The average response time drops from **35ms to 15ms** as a lot of entity role checks can be skipped ([details](https://github.com/corona-school/backend-loader))

### DIFF
--- a/graphql/authorizations.ts
+++ b/graphql/authorizations.ts
@@ -58,6 +58,26 @@ export function AuthorizedDeferred(...requiredRoles: Role[]) {
     });
 }
 
+export function ImpliesRoleOnResult(role: Role) {
+    assert(
+        entityRoles.some((it) => it.role === role),
+        'Only entity roles can be implied'
+    );
+
+    return createMethodDecorator<GraphQLContext>(async ({ args, root, info, context }, next) => {
+        const result = await next();
+        if (Array.isArray(result)) {
+            for (const it of result) {
+                storeDeterminedEntityRole(context, role, it, true);
+            }
+        } else {
+            storeDeterminedEntityRole(context, role, result, true);
+        }
+
+        return result;
+    });
+}
+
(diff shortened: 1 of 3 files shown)


graphql/authorizations.ts
```diff
@@ -113,6 +155,75 @@ async function accessCheck(context: GraphQLContext, requiredRoles: Role[], model
...
+// As we traverse multiple edges from the same entity, we might need to check the same entity roles for the same entity again,
+// as these checks can be costly involving a roundtrip to the db, we cache the evaluated role in the GraphQL context
+
+// undefined -> did not yet checked role
+// true -> user has role on entity
+// false -> user does not have role on entity
+type EvaluatedEntityRoles = { [role in Role]?: boolean };
+type RolesForEntities = Map<any /* root entity */, EvaluatedEntityRoles>;
+
+const EVALUATED_ROLES = Symbol();
```
realmayus, line 173: Out of curiosity, why is a `Symbol` used here? I read it's supposed to guarantee uniqueness, but why can't we just use e.g. `context.evaluatedRoles` to access the map instead? 🙂

realmayus (dismissed):
Very nice! :)

graphql/authorizations.ts
Jonasdoubleyou: Cause I wanted to make sure that this attribute is only accessed and modified in authorizations.ts, as writing it is highly dangerous.

PR #649, opened 2023-05-21
Cleanup Web

Jonasdoubleyou:
- Load Browser on demand when rendering PDF to simplify setup, drop the --noPDF option
- Move Server setup to dedicated file to make index.ts smaller
- Drop hpp, we barely have REST APIs anymore so no need for that
- Drop cleanup actions, the single action we have is nonsense
- Move Routers to Controllers


realmayus (approved):
Nice!

Leo310 (approved):

PR #652, opened 2023-05-26
feat: create matchee chat

LomyW:


graphql/chat/mutations.ts
```diff
@@ -27,8 +27,8 @@ export class MutateChatResolver {
             },
         };
 
-        await getOrCreateConversation(matchees, conversationInfos);
-        return true;
+        const conversation = await getOrCreateConversation(matchees, conversationInfos);
+        return conversation.id;
```
timidak, line 31: LGTM, lets just check this with @Jonasdoubleyou, hence the "convention" in this code base is to return true. But in this case seems valid to return the conversation.id as we need it in the frontend to access the created convo.

timidak (commented):
LGTM 👍🏻

Jonasdoubleyou (approved):

graphql/chat/mutations.ts
Jonasdoubleyou: Yeah that convention is mainly for "void" methods

PR #655, opened 2023-05-27
fix: return conversationId of participants chat

LomyW:


Jonasdoubleyou (approved):

graphql/chat/mutations.ts
```diff
@@ -53,8 +53,8 @@ export class MutateChatResolver {
         };
 
         if (allowed) {
-            await getOrCreateConversation([user, participantUser], conversationInfos);
-            return true;
+            const conversation = await getOrCreateConversation([user, participantUser], conversationInfos);
+            return conversation.id;
         }
         throw new Error('Participant is not allowed to create conversation.');
```
timidak, line 59: I think this is misleading, because participants and instructors of a subcourse should be able to create conversations with each other.

PR #660, opened 2023-05-31
refactor: more resilient course-promotion

Leo310:
Resolves https://github.com/corona-school/project-user/issues/691

**Open questions:**
1. Where should I put the shuffle function?
Because [shuffling directly on the prisma](https://github.com/prisma/prisma/discussions/5886) query is not possible atm.

2. What should be the predictedPupilResponseRate?

3. What Notification state should be passed when calling bulkCreateConcreteNotifications in bulkactionTaken()?

### DIFF
--- a/common/mails/courses/index.ts
+++ b/common/mails/courses/index.ts
@@ -17,6 +17,7 @@ import { getCourseCapacity, isParticipant } from '../../courses/participants';
 import { getCourseImageURL } from '../../courses/util';
 import { createSecretEmailToken } from '../../secret';
 import { getCourse } from '../../../graphql/util';
+import { shuffleArray } from '../../../common/util/basic';
 
 const logger = getLogger('Course Notification');
 
@@ -234,19 +235,8 @@ const isPromotionValid = (publishedAt: Date, capacity: number, alreadyPromoted: 
     return capacity < 0.75 && alreadyPromoted === false && (daysDiff === null || daysDiff > 3);
 };
 
-function shuffleArray<T>(array: T[]): T[] {
-    const shuffledArray = [...array]; // Create a copy of the original array
-
-    for (let i = shuffledArray.length - 1; i > 0; i--) {
-        const j = Math.floor(Math.random() * (i + 1));
-        [shuffledArray[i], shuffledArray[j]] = [shuffledArray[j], shuffledArray[i]];
-    }
-
-    return shuffledArray;
-}
-
 export async function sendPupilCoursePromotion(subcourse: Prisma.subcourse) {
-    const predictedPupilResponseRate = 10; // in percent
+    const predictedPupilResponseRate = 5; // in percent
 
     const courseCapacity = await getCourseCapacity(subcourse);
     const { alreadyPromoted, publishedAt } = subcourse;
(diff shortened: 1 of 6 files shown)


Jonasdoubleyou:
1. some generic function common/util, then shuffle the array after fetching all pupils from Prisma? Fetching a few thousand pupils at once shouldn't be a problem
2. I guess we should start at something like 5%, then adjust after some weeks
3. DELAYED, as we want the background job to pick up the concrete notifications and send them out later

common/mails/courses/index.ts
```diff
@@ -234,40 +234,64 @@ const isPromotionValid = (publishedAt: Date, capacity: number, alreadyPromoted:
     return capacity < 0.75 && alreadyPromoted === false && (daysDiff === null || daysDiff > 3);
 };
 
-export async function sendPupilCourseSuggestion(course: Course | Prisma.course, subcourse: Prisma.subcourse, actionId: ActionID) {
-    // TODO: Reenable once we stabilized course promotions
-    return;
+function shuffleArray<T>(array: T[]): T[] {
```
Jonasdoubleyou, line 237: Could be moved to /common/util

Jonasdoubleyou (dismissed):
very nice!

common/mails/courses/index.ts
Jonasdoubleyou: Also just for fun: This could be implemented slightly more efficient ;) https://github.com/jashkenas/underscore/pull/2158 ... However for a few thousand elements it probably won't matter

Leo310: so you mean using a new approach right? because I dont see how you could tweake the current approach to make it more efficient. ([reference for current shuffle](https://stackoverflow.com/questions/2450954/how-to-randomize-shuffle-a-javascript-array))

common/mails/courses/index.ts
```diff
@@ -234,40 +235,53 @@ const isPromotionValid = (publishedAt: Date, capacity: number, alreadyPromoted:
...
-    const context = await getNotificationContextForSubcourse(course, subcourse);
-
-    if (actionId === 'available_places_on_subcourse' && !isPromotionValid(publishedAt, courseCapacity, alreadyPromoted)) {
-        throw new Error(`Cannot send course promotion for Subcourse(${subcourse.id})`);
-    }
-
-    let sentCount = 0;
-    for (let pupil of pupils) {
+    const courseSubject = course.subject;
+    const filteredPupils = pupils.filter(async (pupil) => {
```
Jonasdoubleyou, line 260: A Promise is always truthy, the filter won't work. Maybe we can create a Set of pupil ids of the Course participants, then do `.filter(pupil => !participantIds.has(pupil.id))`

Jonasdoubleyou: Alternatively we could also filter on the database.

common/mails/courses/index.ts
```diff
@@ -234,40 +235,53 @@ const isPromotionValid = (publishedAt: Date, capacity: number, alreadyPromoted:
...
         const subjects = parseSubjectString(pupil.subjects);
+        const pupilAlreadyParticipant = await isParticipant(subcourse, pupil);
         const isPupilsSubject = subjects.some((subject) => subject.name == courseSubject);
-        if (!courseSubject || isPupilsSubject) {
-            await Notification.actionTaken(pupil, actionId, context);
-            sentCount += 1;
-        }
-    }
+        return !courseSubject || (!pupilAlreadyParticipant && isPupilsSubject);
```
Jonasdoubleyou, line 264: (!courseSubject || isPupilSubject) && !isAlreadyParticipant

PR #663, opened 2023-06-01
feat: announcement group chat

LomyW:


Jonasdoubleyou (commented):
linter errors to be addressed

common/courses/states.ts
```diff
@@ -129,15 +131,41 @@ export async function editSubcourse(subcourse: Subcourse, update: Partial<Subcou
     if (!can.allowed) {
         throw new Error(`Cannot edit Subcourse(${subcourse.id}) reason: ${can.reason}`);
     }
+    const participantCount = await prisma.subcourse_participants_pupil.count({ where: { subcourseId: subcourse.id } });
 
     const isMaxParticipantsChanged: boolean = Boolean(update.maxParticipants);
+    const isGroupChatTypeChanged: boolean = Boolean(update.groupChatType);
```
timidak, line 137: Even if it's not a mandatory field it could be that the groupChatType is present but hasn't changes. We should check if the last type equals the current value present in the update.

common/courses/states.ts
```diff
@@ -129,15 +131,41 @@ export async function editSubcourse(subcourse: Subcourse, update: Partial<Subcou
...
 
     if (isMaxParticipantsChanged) {
-        const participantCount = await prisma.subcourse_participants_pupil.count({ where: { subcourseId: subcourse.id } });
         if (update.maxParticipants < participantCount) {
             throw new PrerequisiteError(`Decreasing the number of max participants below the current number of participants is not allowed`);
         }
     }
+    if (isGroupChatTypeChanged) {
+        //* if the subcourse has already an conversation, the conversation has to be updated
+        if (subcourse.conversationId !== null) {
```
timidak, line 146: What if `conversationId` is `null`? Is this even possible?

common/courses/states.ts
```diff
@@ -129,15 +131,41 @@ export async function editSubcourse(subcourse: Subcourse, update: Partial<Subcou
...
+                await markConversationAsWriteable(subcourse.conversationId);
+                await updateConversation(conversationToBeUpdated);
+            } else if (update.groupChatType === ChatType.ANNOUNCEMENT) {
+                const conversationToBeUpdated: { id: string } & ConversationInfos = {
+                    id: subcourse.conversationId,
+                    custom: {
+                        type: 'announcement',
+                    },
+                };
+                await markConversationAsReadOnly(subcourse.conversationId);
```
timidak, line 164: Just a question: in some cases we want to mark the whole conversation as readonly and in this case we want to mark only the participants (not the instructors) as readonly, is this handled by this function? Have you thought of this?

graphql/chat/mutations.ts
```diff
@@ -65,11 +66,14 @@ export class MutateChatResolver {
             subject: subcourse.course.name,
             custom: {
                 start: subcourse.lecture[0].start.toISOString(),
-                type: 'course',
+                type: groupChatType === ChatType.ANNOUNCEMENT ? 'announcement' : 'course',
```
timidak, line 69: Why are we using different enums and values in similar places? I so far saw:
- announcement
- course
- normal
What's the matter with those types?

LomyW: https://github.com/corona-school/backend/pull/686

common/chat/conversation.ts
```diff
@@ -251,12 +251,44 @@ async function removeParticipant(user: User, conversationId: string): Promise<vo
     }
 }
 
+function isStudentChatMember(memberUserId: string): boolean {
+    if (memberUserId.includes('student')) {
+        return true;
+    } else {
+        return false;
+    }
```
timidak, line 259: just return `memberUserId.includes('student')`

common/chat/conversation.ts
```diff
@@ -251,12 +251,44 @@ async function removeParticipant(user: User, conversationId: string): Promise<vo
...
+                    'Content-Type': 'application/json',
+                },
+                body: JSON.stringify({
+                    access: 'Read',
+                }),
+            });
+            await checkResponseStatus(response);
+        }
+    } catch (error) {
+        throw new Error(error);
```
timidak, line 281:
```suggestion
       throw new Error("Could not mark conversation as readonly.", { cause: error })
```

timidak: Rethrowing is useless unless you want to throw an own error. 🙂

timidak (dismissed):
LGTM 👍🏻

timidak (approved):

PR #667, opened 2023-06-03
End-to-end integration tests

Jonasdoubleyou:
- Run the Webserver from within the test process itself, to be able to mock functions (among other benefits)
- Introduce a logging mode "brief" to not make logging inside tests too verbose
- Mock `fetch`, to make sure tests are always self containing and do not call external systems, allow fetch calls to be mocked with `expectFetch`
- Make `npm run integration-tests` more or less self containing (set up all relevant env variables), as it isn't that configurable now anyways

### DIFF
--- a/common/prisma/index.ts
+++ b/common/prisma/index.ts
@@ -3,4 +3,6 @@ import { PrismaClient } from '@prisma/client';
 /* In the development environment, also log 'query'
    which is the SQL query sent to the database.
    Through that, queries can be optimized for performance */
-export const prisma = new PrismaClient(process.env.ENV !== 'dev' ? undefined : { log: ['query', 'info', `warn`, `error`] });
+export const prisma = new PrismaClient(
+    process.env.ENV !== 'dev' || process.env.LOG_FORMAT === 'brief' ? undefined : { log: ['query', 'info', `warn`, `error`] }
+);
--- a/integration-tests/base.ts
+++ b/integration-tests/base.ts
@@ -1,6 +1,8 @@
 import { randomBytes } from "crypto";
 import { GraphQLClient } from "graphql-request";
 
+import "./mock";
+
 import * as WebServer from "../web";
 
 /* -------------- Configuration ------------------- */
--- a/integration-tests/base.ts
+++ b/integration-tests/base.ts
@@ -15,9 +15,9 @@ const silent = process.env.INTEGRATION_SILENT === "true";
 
 /* -------------- Utils --------------------------- */
 
-const blue = (msg: string) => '\u001b[94m' + msg + '\u001b[39m';
-const red = (msg: string) => '\u001b[31m' + msg + '\u001b[39m';
-const green = (msg: string) => '\u001b[32m' + msg + '\u001b[39m';
+export const blue = (msg: string) => '\u001b[94m' + msg + '\u001b[39m';
+export const red = (msg: string) => '\u001b[31m' + msg + '\u001b[39m';
+export const green = (msg: string) => '\u001b[32m' + msg + '\u001b[39m';
 
 console.log(
     blue(`\n\nBackend Integration Tests\n`) +
--- a/integration-tests/base.ts
+++ b/integration-tests/base.ts
@@ -1,10 +1,12 @@
 import { randomBytes } from "crypto";
 import { GraphQLClient } from "graphql-request";
 
+import * as WebServer from "../web";
+
 /* -------------- Configuration ------------------- */
 
 const APP = "lernfair-backend-dev";
-const URL = process.env.INTEGRATION_TARGET ?? `https://${APP}.herokuapp.com/apollo`;
+const URL = process.env.INTEGRATION_TARGET ?? `http://localhost:${process.env.PORT ?? 5000}/apollo`;
 const ADMIN_TOKEN = process.env.ADMIN_AUTH_TOKEN;
 
 const silent = process.env.INTEGRATION_SILENT === "true";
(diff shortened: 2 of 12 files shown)


integration-tests/matching.ts
```diff
@@ -35,6 +36,194 @@ void test('Pupil Request Match', async () => {
...
+    assert.strictEqual(s1.matches[0].pupil.firstname, pupil.firstname);
+    assert.strictEqual(s1.matches[0].pupil.lastname, pupil.lastname);
+
+    return { id: s1.matches[0].id, pupil, student, pupilClient, studentClient };
+});
+
+void test('Create Chat for Match', async () => {
+    const { id, pupil, student, pupilClient, studentClient } = await match1;
+
+    // Creating a Chat ensures all users have accounts:
```
Jonasdoubleyou, line 136: That's how the mocking would look like then:

common/prisma/index.ts
```diff
@@ -3,4 +3,6 @@ import { PrismaClient } from '@prisma/client';
 /* In the development environment, also log 'query'
    which is the SQL query sent to the database.
    Through that, queries can be optimized for performance */
-export const prisma = new PrismaClient(process.env.ENV !== 'dev' ? undefined : { log: ['query', 'info', `warn`, `error`] });
+export const prisma = new PrismaClient(
+    process.env.ENV !== 'dev' || process.env.LOG_FORMAT === 'brief' ? undefined : { log: ['query', 'info', `warn`, `error`] }
```
dhenkel92, line 7: I don't think that the log format should control which logs are published. Could we use a different environment for the integration tests?

integration-tests/base.ts
```diff
@@ -133,16 +143,19 @@ export async function finalizeTests() {
...
     }
 
-    console.log(`\n\nsummary:`);
-
     const durationAll = Date.now() - startAll;
 
     if (failureCount === 0) {
-        console.log(green(`  all tests SUCCEEDED in ${durationAll}ms`));
+        console.log(blue('\n\n-------------------- TEARDOWN ------------------------'));
+        await WebServer.shutdown();
```
dhenkel92, line 150: We only want to gracefully shutdown the server if there is no failure?

integration-tests/mock.ts
```diff
@@ -0,0 +1,51 @@
...
+    url: string; // use some/path/* as a wildcard
+    body?: string; // Use * as a wildcard
+    responseStatus: number;
+    response?: string | object;
+}
+
+let expectedCalls: MockedFetch[] = [];
+
+export function expectFetch(mocked: MockedFetch) {
+    expectedCalls.push(mocked);
```
dhenkel92, line 16: Should we add a reset mechanism, so that every test starts with a clean state? Otherwise, test2 might still have the `expectFetch` elements from test1.

integration-tests/mock.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export function setupFetchMock() {
+    const originalFetch = global.fetch;
+
+    async function mockedFetch(resource: RequestInfo, options: RequestInit) {
+        const url = "" + resource;
+
+        const mock = expectedCalls.find(it => (it.url === url || wcmatch(it.url)(url)) && it.method === (options.method ?? "GET"));
```
dhenkel92, line 25: What if there are multiple matching results within the list? It could be on purpose, but especially with wildcards it's hard to predict what's used in which call.

integration-tests/mock.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        if (mock) {
+            expectedCalls = expectedCalls.filter(it => it !== mock);
```
dhenkel92, line 28: Could you please add a comment to explain this line?

integration-tests/matching.ts
```diff
@@ -35,6 +36,194 @@ void test('Pupil Request Match', async () => {
...
+
+      expectFetch({
+        "url": "https://api.talkjs.com/v1/mocked-talkjs-appid/conversations/*/messages",
+        "method": "POST",
+        "body": "[{\"text\":\"Willkommen im Lern-Fair Chat!\",\"type\":\"SystemMessage\",\"custom\":{\"type\":\"first\"}}]",
+        "responseStatus": 200
+      });
+
+
+    const { matchChatCreate: conversionID } = await pupilClient.request(`
```
dhenkel92, line 222: Should we check if the response make sense or is it enough that it didn't crash?

integration-tests/mock.ts
Jonasdoubleyou: Well I wasn't sure if the order is always deterministic, but I guess we can take care of that once we have that case. Changed this to make it strictly sequential (and easier to follow).

integration-tests/mock.ts
Jonasdoubleyou: good point, added a check

integration-tests/base.ts
Jonasdoubleyou: fair enough, we could shutdown before exiting

common/prisma/index.ts
Jonasdoubleyou: fair, added ENV=test

integration-tests/matching.ts
timidak: Are we mocking based on the url as identifier or is it mocked sequentially (so the first expectFetch mocks the first fetch call and so on)?

integration-tests/mock.ts
```diff
@@ -0,0 +1,72 @@
...
+    body?: string;
+
+    // The then returned response:
+    responseStatus: number;
+    // If an object is passed, the response is JSON
+    response?: string | object;
+}
+
+// A FIFO queue of expected calls:
+let expectedCalls: MockedFetch[] = [];
```
timidak, line 21: Nevermind, found it here 😄

integration-tests/matching.ts
```diff
@@ -35,6 +36,195 @@ void test('Pupil Request Match', async () => {
...
+      });
+
+        // TODO: Remove duplicate fetch
+      expectFetch({
+        url: "https://api.talkjs.com/v1/mocked-talkjs-appid/conversations/*",
+        method: "GET",
+        responseStatus: 404
+      });
+
+      // TODO: Why PUT twice?
```
timidak, line 183: The API design of TalkJS is a bit weird, they use PUT for creating (instead of POST) and updating conversations

common/prisma/index.ts
Leo310: huch habs ausversehen unresolved 😬

realmayus (commented):
Looks great so far, thank you! :)

Jonasdoubleyou:
So good to go @dhenkel92 @timidak @realmayus ?

PR #669, opened 2023-06-05
fix(IssueReporter): add issue reporter logs to error tracking

dhenkel92:
This PR will move the issue reporter errors to Datadog's [error tracking](https://app.datadoghq.eu/logs/error-tracking?from_ts=1685980314420&to_ts=1685994714420&live=true), so that we don't get notified for the same error again and again.

### DIFF
--- a/web/server.ts
+++ b/web/server.ts
@@ -38,7 +38,7 @@ app.use((req, res, next) => {
 });
 
 // Parse Cookies and JSON Bodies:
-app.use(bodyParser.json());
+app.use(bodyParser.json({ limit: '1mb' }));
 app.use(cookieParser());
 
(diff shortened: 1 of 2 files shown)


web/server.ts
```diff
@@ -38,7 +38,7 @@ app.use((req, res, next) => {
 });
 
 // Parse Cookies and JSON Bodies:
-app.use(bodyParser.json());
+app.use(bodyParser.json({ limit: '1mb' }));
```
dhenkel92, line 41: The default is 100kb. I had to increase it, because the payload of the issue reporter is pretty big.

dhenkel92:
https://github.com/corona-school/user-app/pull/274

Jonasdoubleyou (approved):

PR #678, opened 2023-06-08
feature: move GraphQL error logging to formatError

dhenkel92:
Unfortunately, the didEncounterErrors hook doesn't have a lot of information about which kind of errors did happen. This means, that it's alsomost impossible to distinguish them and decide which to log how.

Therfore, I moved the logging to the formatError hook, because the given error is always wrapped into a class with context, like ValidationError. This helps us to devide when to log what.

### DIFF
--- a/graphql/logging.ts
+++ b/graphql/logging.ts
@@ -1,10 +1,6 @@
 import { GraphQLRequestContext } from 'apollo-server-plugin-base';
 import { isDev } from '../common/util/environment';
 import { getLogger } from '../common/logger/logger';
-import { toPublicToken } from './authentication';
-import { Role } from './authorizations';
-import { GraphQLContext } from './context';
-import { isUnexpectedError } from './error';
 import { v4 as uuidv4 } from 'uuid';
 
 const logger = getLogger('GraphQL Query');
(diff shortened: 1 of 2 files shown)


graphql/logging.ts
```diff
@@ -26,18 +22,6 @@ export const GraphQLLogger: any = {
         logger.info(`Started processing query`, { query });
 
         const handler: any = {
-            // Actually GraphQLRequestListener, but we're on v2 and not on v3
-            didEncounterErrors(requestContext: GraphQLRequestContext) {
-                const unexpected = requestContext.errors.some(isUnexpectedError);
```
dhenkel92, line 31: The errors in this case always have the same type `GraphQLError` and often miss the originalError field, which makes it hard to distinguish.

dhenkel92: I've checked the logs of the recent 15 days and there wasn't a single `Unexpected Errors`. In combination with type findings I would assume that the current logic doesn't work as expected.

We could log all the errors as info, so that we don't loose the information.

Jonasdoubleyou (approved):

realmayus (approved):

PR #684, opened 2023-06-10
feat(screening-entity): add job status data field

Leo310:
Resolves part of https://github.com/corona-school/project-user/issues/750

### DIFF
--- a/common/entity/ScreeningTypes.ts
+++ b/common/entity/ScreeningTypes.ts
@@ -1,0 +1,19 @@
+export enum JobStatus {
+    STUDENT = 'Schüler:in',
+    PUPIL = 'Student:in',
+    EMPLOYEE = 'Angestellte:r',
+    RETIREE = 'Rentner:in',
+    MISC = 'Sonstige',
+}
+
+export enum KnowsCoronaSchoolFrom {
+    FRIENDS = 'Persönliche Empfehlung/Freund:innen',
+    VOLUNTEEREXCHANGE = 'Ehrenamtsbörse',
+    INSTAGRAM = 'Instagram',
+    LINKEDIN = 'LinkedIn',
+    SEARCHENGINE = 'Suchmaschine',
+    PRINTMEDIA = 'Printmedien (Flyer, Poster etc.)',
+    PRESS = 'Presse',
+    SCHOLARSHIP = 'Stipendium',
+    OTHER = 'Sonstiges',
+}
--- a/common/entity/ScreeningTypes.ts
+++ b/common/entity/ScreeningTypes.ts
@@ -1,19 +1,7 @@
 export enum JobStatus {
-    STUDENT = 'Schüler:in',
-    PUPIL = 'Student:in',
-    EMPLOYEE = 'Angestellte:r',
-    RETIREE = 'Rentner:in',
-    MISC = 'Sonstige',
-}
-
-export enum KnowsCoronaSchoolFrom {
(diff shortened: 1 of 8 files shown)


common/entity/ScreeningTypes.ts
```diff
@@ -0,0 +1,19 @@
+export enum JobStatus {
+    STUDENT = 'Schüler:in',
+    PUPIL = 'Student:in',
+    EMPLOYEE = 'Angestellte:r',
+    RETIREE = 'Rentner:in',
+    MISC = 'Sonstige',
+}
+
+export enum KnowsCoronaSchoolFrom {
```
Leo310, line 9: This is not in use yet. Any naming suggestions for the database field? Because knowsCoronaSchoolFrom: string already exists

common/student/screening.ts
```diff
@@ -5,10 +5,12 @@ import { getLogger } from '../../common/logger/logger';
 import { createRemissionRequest } from '../remission-request';
 import { getTransactionLog } from '../transactionlog';
 import CoCCancelledEvent from '../transactionlog/types/CoCCancelledEvent';
+import { screening_jobstatus_enum, instructor_screening_jobstatus_enum } from '../../graphql/generated';
 
 interface ScreeningInput {
     success: boolean;
     comment?: string;
+    jobStatus?: screening_jobstatus_enum | instructor_screening_jobstatus_enum;
```
Leo310, line 13: prisma generates two different generated/enums out of the one entity/JobStatus enum, which I cannot use here for some reason (getting type error). That is why I have this kinda weird or statement for one and the same thing.

prisma/schema.prisma
```diff
@@ -1025,3 +1027,19 @@ enum chat_type {
   NORMAL
   ANNOUNCEMENT
 }
+
+enum instructor_screening_jobstatus_enum {
```
Leo310, line 1031: Also prisma isnt really a fan of umlaute.
image
Not sure how if this will be display correctly in retool

common/student/screening.ts
Jonasdoubleyou: I recently found out that one can pass an enum name to the Column(...) declaration, in which case there will be only one enum created in the DB, this npx db pull will also only create one enum

prisma/schema.prisma
Jonasdoubleyou: I would prefer if we would keep the enums english, translation can be done in the UI

common/entity/ScreeningTypes.ts
Jonasdoubleyou: I guess we can just continue to use the existing varchar and write some well known strings into it so that COUNT(*) GROUP BY works

Leo310: With writing some well-known strings into it, you mean handling these options in retool without changing anything in the backend right?

common/student/screening.ts
Leo310: Ok cool will check it out

Leo310: Ok cool will check it out

Leo310: Ok cool will check it out

common/entity/ScreeningTypes.ts
Jonasdoubleyou: yes exactly

Jonasdoubleyou: Thats also easier to adapt in the future when requirements change

Jonasdoubleyou (approved):

PR #686, opened 2023-06-12
fix: chat types

LomyW:


common/migration/1686567733355-chat-type-enum.ts
```diff
@@ -0,0 +1,11 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class chatTypeEnum1686567733355 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "subcourse" ADD "groupChatType" "public"."chat_type" NOT NULL DEFAULT 'course'`);
```
Jonasdoubleyou, line 5: missing migration for the enum type

Jonasdoubleyou (changes requested):

common/migration/1686567733355-chat-type-enum.ts
Jonasdoubleyou: I guess we also don't have to touch the subcourse.groupChatType then, the rename should also be reflected on the default value.

PR #700, opened 2023-06-22
fix: appointment pagination and managing empty states in FE

timidak:
- enhanced pagination to be able to get skip values
- added edges to get if a user has appointments and the last appointment id present


graphql/user/fields.ts
```diff
@@ -169,6 +169,18 @@ export class UserFieldsResolver {
         return getAppointmentsForUser(user, take, skip, cursor, direction);
     }
```
Jonasdoubleyou, line 171: Shouldn't take and skip be mandatory then in appointments(...)?

Jonasdoubleyou (approved):

graphql/user/fields.ts
timidak: I defaulted in the getAppointmentsForUser, but you're right the API is cleaner this way.
Will try it locally in a sec and then push the update.

Jonasdoubleyou (dismissed):

Jonasdoubleyou (approved):

PR #702, opened 2023-06-24
Type Notification.actionTaken's context

Jonasdoubleyou:
We have an insane amount of bugs at that interface

resolves https://github.com/corona-school/project-user/issues/549


graphql/appointment/mutations.ts
```diff
@@ -92,8 +91,7 @@ export class MutateAppointmentResolver {
         for await (const participant of participants) {
             await Notification.actionTaken(participant.pupil, 'student_add_appointment_group', {
                 student: student,
-                user: participant,
```
dhenkel92, line 95: /nit let's remove the await in line 92 as well

common/courses/participants.ts
```diff
@@ -268,7 +268,7 @@ export async function leaveSubcourse(subcourse: Subcourse, pupil: Pupil) {
     logger.info(`Pupil(${pupil.id}) left Subcourse(${subcourse.id})`);
     await logTransaction('participantLeftCourse', pupil, { subcourseID: subcourse.id });
 
-    const course = prisma.course.findUnique({ where: { id: subcourse.courseId } });
+    const course = await prisma.course.findUnique({ where: { id: subcourse.courseId } });
```
dhenkel92, line 271: We always find new places like this 😞

dhenkel92 (dismissed):

common/courses/participants.ts
Jonasdoubleyou: jup, need better linters. It's so hard to spot these in code review ...

PR #707, opened 2023-06-26
fix: refresh declined or cancelled appointments

LomyW:
- make `matchId` accessable for appointment participants


timidak (approved):
Simple yet good 👍🏻

PR #710, opened 2023-06-28
fix(migration): change instructor_screening_jobstatus_enum values to match prisma

dhenkel92:
Follow up of [this pr](https://github.com/corona-school/backend/pull/684/files#diff-3c1eda158e747f9172dc1aacd07b21a3f087105ba55efda93d86ce0787f6aadeR1-R23) to rename the instructor screening enum values as well.

### DIFF
--- a/common/migration/1687941994171-updateInstructorScreeningEnumType.ts
+++ b/common/migration/1687941994171-updateInstructorScreeningEnumType.ts
@@ -1,0 +1,21 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class updateInstructorScreeningEnumType1687941994171 implements MigrationInterface {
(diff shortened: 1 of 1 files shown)


common/entity/InstructorScreening.ts
```diff
@@ -17,7 +17,7 @@ export class InstructorScreening {
     })
     comment: string;
 
-    @Column({ type: 'enum', enum: JobStatus, nullable: true, enumName: 'screening_jobstatus_enum' })
+    @Column({ type: 'enum', enum: JobStatus, nullable: true, enumName: 'instructor_screening_jobstatus_enum' })
```
Jonasdoubleyou, line 20: actually we wanted to share the enum for both

Jonasdoubleyou (approved):

PR #711, opened 2023-06-28
Add lastLogin property to users

realmayus:


common/user/index.ts
```diff
@@ -294,6 +294,18 @@ export async function getUsers(userIds: User['userID'][]): Promise<User[]> {
...
 
+export async function updateLastLogin(user: User) {
+    if (user.pupilId) {
+        await prisma.pupil.update({ where: { id: user.pupilId }, data: { lastLogin: new Date() } });
+    }
+    if (user.studentId) {
+        await prisma.student.update({ where: { id: user.studentId }, data: { lastLogin: new Date() } });
+    }
+    if (user.screenerId) {
+        await prisma.student.update({ where: { id: user.screenerId }, data: { lastLogin: new Date() } });
```
dhenkel92, line 305: should it be prisma.screener.update?

graphql/authentication.ts
```diff
@@ -127,6 +127,7 @@ export async function loginAsUser(user: User, context: GraphQLContext, noSession
         await userSessions.set(context.sessionToken, context.user);
         logger.info(`[${context.sessionToken}] User(${user.userID}) successfully logged in`);
     }
+    await updateLastLogin(user);
```
dhenkel92, line 130: Can we make sure that it's not triggered when we are using the admin token?

common/user/index.ts
```diff
@@ -294,6 +294,18 @@ export async function getUsers(userIds: User['userID'][]): Promise<User[]> {
     return [...students, ...pupils];
 }
 
+export async function updateLastLogin(user: User) {
```
dhenkel92, line 297: What about the metnor?

common/entity/Person.ts
```diff
@@ -58,6 +58,13 @@ export abstract class Person {
 
     @Column({ nullable: true, type: 'json' })
     notificationPreferences: any;
+
+    @Column({
+        type: 'timestamp',
+        nullable: true,
+        default: new Date(),
+    })
+    lastLogin?: Date;
```
dhenkel92, line 67: I think we also need a migration to add the field in production

common/user/index.ts
realmayus: I don't think Mentors are a thing anymore. I did a quick search and only was able to find them in the dev.ts setup, but there don't seem to be any actual usages. The `userType` doesn't specify mentors: https://github.com/corona-school/backend/blob/894a832bba8a1db12cc1ecbb4104e26dfa6e43b6/common/user/index.ts#L157

common/entity/Person.ts
realmayus: I always forget those 😅

Jonasdoubleyou: Why is this nullable while the migration creates a NOT NULL column?

common/user/index.ts
Jonasdoubleyou: I guess we should create an issue to remove it fully from the backend

graphql/authorizations.ts
```diff
@@ -401,6 +401,7 @@ export const authorizationModelEnhanceMap: ModelsEnhanceMap = {
             match: nobody,
             _count: nobody,
             waiting_list_enrollment: nobody,
+            lastLogin: nobody,
```
Jonasdoubleyou, line 404: I think we could release it to onlyAdminOrOwner. This might also be useful for support

common/entity/Person.ts
realmayus: So you think there shouldn't be a default value at all? I think having a default makes sense, as the redaction job can just work with the default values, such that accounts that don't log in after the migration will be redacted in 3 months. 🙂 
I agree though, it's an oversight to have unmatching properties.

dhenkel92: I think all the current users should have `now()` as default as we don't have any other source to firgure out their last login.
For new users we should think about it tho. If allowing null, we could see how many users never logged in. Maybe this is also possible with other metrics :-D

prisma/schema.prisma
```diff
@@ -259,6 +259,7 @@ model mentor {
   isRedacted                   Boolean                 @default(false)
   lastTimeCheckedNotifications DateTime?               @default(dbgenerated("'1970-01-01 00:00:00'::timestamp without time zone")) @db.Timestamp(6)
   notificationPreferences      Json?                   @db.Json
+  lastLogin                    DateTime?               @default(now()) @db.Timestamp(6)
```
dhenkel92, line 262: Can we also make sure that prisma is updating/setting the createAt/updatedAt fields?

realmayus: `createdAt` is set by the `@default(now())` annotation afaik, but `updatedAt` is a TypeORM relic that's not supported by Prisma iirc. That's the reason why we introduced `lastLogin`, as `updatedAt` doesn't accurately represent the last time someone changed an object in the DB which caused trouble in the automated account redaction job.

dhenkel92: It seems like there is [an option](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#updatedat) in Prisma for this.
It would still be cool to keep the updatedAt, just to keep track of changes.

realmayus: Hmm indeed, but TypeORM is our "source of truth" right now and we pull the prisma schema from the database. So we'd have to manually add `@updatedAt` every time we pull the schema 🤔 @Jonasdoubleyou is that a feasible approach? or should we wait until we ditch TypeORM

dhenkel92: Okay, that makes a lot of sense. Let's skip it until we finally got rid of TypeORM 👍

dhenkel92 (approved):

PR #719, opened 2023-07-06
fix: last appointment date

LomyW:
**Description**
The date of the last appointment that was created is always set to '02:00' on the next day.

**What was done?**
- [x] fixed zoom scheduled meeting end date with using moment to set end date time

### DIFF
--- a/common/zoom/zoom-scheduled-meeting.ts
+++ b/common/zoom/zoom-scheduled-meeting.ts
@@ -80,7 +80,7 @@ const createZoomMeeting = async (zoomUsers: ZoomUser[], startTime: Date, isCours
                     waiting_room: isCourse ? true : false,
                     breakout_room: isCourse ? true : false,
                     recurrence: endDateTime && {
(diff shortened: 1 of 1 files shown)


common/zoom/zoom-scheduled-meeting.ts
```diff
@@ -80,7 +80,7 @@ const createZoomMeeting = async (zoomUsers: ZoomUser[], startTime: Date, isCours
                     waiting_room: isCourse ? true : false,
                     breakout_room: isCourse ? true : false,
                     recurrence: endDateTime && {
-                        end_date_time: new Date(endDateTime.setHours(24, 0, 0, 0)),
+                        end_date_time: moment(endDateTime).tz(tz).format('YYYY-MM-DDTHH:mm:ss'),
```
timidak, line 83: What you actually did was fixing the passing by reference of a date which that gets maninpulated here and manipulates the last date before creating the appointment. But good job! 👍🏻

timidak (approved):

PR #723, opened 2023-07-09
Cache compiled message templates of Notifications

Jonasdoubleyou:
This saves a DB roundtrip and template parsing for every rendered notification (which can be quite a lot when a user opens a full notification inbox).

Additionally fixes:
- Special characters aren't HTML encoded anymore (the frontend should do this instead, when it embeds the data somewhere)
- Unified sample context logic between campaigns and message rendering (distinguish the context coming in and going out to the template, fix the edge case where a notification has neither actions nor a sample_context)

resolves https://github.com/corona-school/project-user/issues/787


common/notification/messages.ts
```diff
@@ -42,6 +48,68 @@ export async function getMessageForNotification(
...
+export async function getMessage(
+    concreteNotification: ConcreteNotification,
+    language: TranslationLanguage = TranslationLanguage.DE
+): Promise<NotificationMessageType | null> {
+    const template = await loadTemplate(concreteNotification.notificationID, language);
+    if (!template) {
+        return null;
+    }
+
+    const legacyUser = await getUserTypeORM(concreteNotification.userId);
```
dhenkel92, line 101: /nit Should we use prisma to push the migration forward?

common/notification/messages.ts
```diff
@@ -42,6 +48,68 @@ export async function getMessageForNotification(
...
 
+type CachedMessageTemplate = {
+    type: NotificationType;
+    body: (context: NotificationContext) => string;
+    headline: (context: NotificationContext) => string;
+    navigateTo?: (context: NotificationContext) => string;
+    modalText?: (context: NotificationContext) => string;
+};
+
+const cachedTemplates = new Map</* notification id + language */ string, CachedMessageTemplate | null /* = no message */>();
```
dhenkel92, line 59: This would mean that we have to restart the backend for every notification change.
This doesn't happen very often, but it somewhat works against the initial idea to have a dynamic system that can be configured via retool.
Theoretically, we could create a TTL cache to automatically refresh the cache from time to time.

common/notification/notification.ts
```diff
@@ -261,9 +258,21 @@ export function getSampleContext(notification: Notification): SampleContext {
...
+
     return sampleContexts.reduce(subset) as SampleContext;
 }
 
+// Returns a sample context for a notification, this can be used to validate
+// templates which should be rendered with concrete notification contexts,
+// as it contains at least the subset of the available fields in the context,
+// which will definitely be available
+export function getSampleContext(notification: Notification): Context {
+    return { ...getSampleContextExternal(notification), user: sampleUser as any, USER_APP_DOMAIN: 'https://app.lern-fair.de' };
```
dhenkel92, line 273: We should use the `USER_APP_DOMAIN` env variable just in case.

common/notification/messages.ts
Jonasdoubleyou: yes, but touching the noitification system in that way might potentially break all notifications, so I want to do that refactoring in a separate step.

common/notification/messages.ts
```diff
@@ -115,4 +182,7 @@ export async function setMessageTranslation({
             data: { language, notificationId: notification.id, template: { body, headline, modalText }, navigateTo },
         }),
     ]);
+
+    cachedTemplates.delete(notification.id + ':' + language);
```
Jonasdoubleyou, line 186: Here we invalidate ;)

common/notification/messages.ts
Jonasdoubleyou: Not quite, when a Message Template is updated with setMessageTranslation, we invalidate the cache ... So no need to restart. There is also an integration test for that case :)

dhenkel92: For such systems, I'm mostly afraid about the future. We must remember to invalidate the cache every time we update the translation.
In this case, it will not be a problem, but as soon as updates are done in other files, it's getting complex :)

PR #725, opened 2023-07-09
Cleanup some Legacy

Jonasdoubleyou:
Sunset the following programs:
- Mentoring
- Jufo
- Project Matching

They are gone from the User App, but some stuff was still present in the backend (and even partially ported to GraphQL).
If we want to reenable these programs somewhen, we need to rewrite them from scratch anyways.

Additionally:
- Fully remove the TypeORM based transaction log (Prisma only)
- Drop a lot of code related to old jobs
- Drop some methods present in TypeORM entities


dhenkel92 (approved):
"Some legacy" `−3,768 lines` 😆

web/dev.ts
```diff
@@ -334,8 +325,6 @@ export async function setupDevDB() {
     s6.isInstructor = false;
     s6.isProjectCoach = true;
```
Maascha, line 326: don't we have to remove this too?

web/dev.ts
```diff
@@ -103,9 +96,7 @@ export async function setupDevDB() {
     p.isParticipant = false;
     p.isPupil = false;
     p.active = true;
-    p.isJufoParticipant = TuteeJufoParticipationIndication.YES;
     p.isProjectCoachee = true;
```
Maascha, line 99: don't we have to remove this too?

web/dev.ts
Jonasdoubleyou: Yeah, we could also remove them. Will do in some followup PR, this is already large enough. We also need to discuss whether we drop the data from the database.

common/entity/ExpertData.ts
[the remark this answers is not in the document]
Jonasdoubleyou: I would only remove them if we also drop the data from the database, otherwise our ORM and DB wouldn't be in sync.

PR #726, opened 2023-07-09
Add Slack Integration

Jonasdoubleyou:
resolves https://github.com/corona-school/project-user/issues/814

### DIFF
--- a/common/slack/index.ts
+++ b/common/slack/index.ts
@@ -1,5 +1,5 @@
-import { getLogger } from 'log4js';
 import { SlackBlock } from './blocks';
+import { getLogger } from '../logger/logger';
 
 export interface SlackChannel {
     name: string;
@@ -23,15 +23,15 @@ const logger = getLogger('Slack');
 
 export async function sendToSlack(channel: SlackChannel, message: SlackMessage) {
     if (!channel.webhookURL) {
-        logger.warn('Cannot send message to ' + channel.name + ' as webhook is missing');
+        logger.warn('Cannot send message to Slack as webhook is missing', { channel: channel.name });
         return;
     }
 
     const request = await fetch(channel.webhookURL, { method: 'POST', body: JSON.stringify(message), headers: { 'content-type': 'application/json' } });
     if (request.ok) {
-        logger.info('Sent Slack Message to ' + channel.name, { message });
+        logger.info('Sent Slack Message', { message, channel: channel.name });
     } else {
         // Unlike other places we do not fail here with an error, as the Slack integration is usually optional
-        logger.error('Failed to send Slack message', { status: request.status, text: await request.text() }, JSON.stringify(message, null, 2));
+        logger.error('Failed to send Slack message', { status: request.status, text: await request.text(), message: JSON.stringify(message, null, 2) });
     }
 }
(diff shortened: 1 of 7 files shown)


jobs/slack-statistics.ts
```diff
@@ -0,0 +1,27 @@
...
+import { table } from "../common/slack/blocks";
+
+export async function postStatisticsToSlack() {
+    const lastMonthBegin = new Date();
+    lastMonthBegin.setMonth(lastMonthBegin.getMonth() - 1);
+    lastMonthBegin.setDate(1);
+    const begin = lastMonthBegin.toISOString();
+
+    const lastMonthEnd = new Date();
+    lastMonthEnd.setDate(-1);
```
dhenkel92, line 12: If we execute the job manually, it will use the first of last month till yesterday. I would do something similar to the above just to ensure we always select the whole last month.

jobs/slack-statistics.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    const pupilRegisteredCount = await prisma.pupil.count({ where: { createdAt: { gte: begin, lte: end } }});
+    const studentsRegisteredCount = await prisma.student.count({ where: { createdAt: { gte: begin, lte: end }}});
+
+    await sendToSlack(SLACK_CHANNEL.PublicStatistics, {
+        blocks: [
+            table(`Statistiken vom ${begin} zum ${end}`, "Name", "Wert", [
+                ["Anzahl registrierter Schüler", "" + pupilRegisteredCount],
+                ["Anzahl registrierter Helfer", "" + studentsRegisteredCount]
+            ])
```
dhenkel92, line 24: This is an example for now?

common/slack/index.ts
```diff
@@ -0,0 +1,37 @@
+import { getLogger } from 'log4js';
```
dhenkel92, line 1:
```suggestion
import { getLogger } from '../logger/logger';
```

common/slack/index.ts
```diff
@@ -0,0 +1,37 @@
...
+        name: 'publ-statistics',
+        webhookURL: process.env.SLACK_PUBL_STATISTICS,
+    },
+};
+
+const logger = getLogger('Slack');
+
+export async function sendToSlack(channel: SlackChannel, message: SlackMessage) {
+    if (!channel.webhookURL) {
+        logger.warn('Cannot send message to ' + channel.name + ' as webhook is missing');
```
dhenkel92, line 26: We should try to not use variables in log messages. Otherwise, it's hard to search for all logs related to this event.

jobs/slack-statistics.ts
Jonasdoubleyou: Not quite, actually `setDate` takes the absolute day of the month, but handles over and underflows gracefully, so -1 is indeed the last date of the previous month. However I just noticed that we actually skip one whole day that way, will have to adapt this ...

jobs/slack-statistics.ts
Jonasdoubleyou: Yeah, I thought it might be good to split this up into two parts, as the rest of the task would also be a "good first issue"

dhenkel92 (approved):

PR #727, opened 2023-07-10
feat: system messages for new chat

LomyW:
**Description**
System message for courses (when a new group chat is created) and 1:1 chats should send new message

**What was done?**
- [x] adjust first system message


common/chat/conversation.ts
```diff
@@ -15,6 +15,11 @@ const TALKJS_API_URL = `https://api.talkjs.com/v1/${process.env.TALKJS_APP_ID}`;
 const TALKJS_CONVERSATION_API_URL = `${TALKJS_API_URL}/conversations`;
 const TALKJS_API_KEY = process.env.TALKJS_API_KEY;
 
+const ONE_ON_ONE_SYSTEM_MESSAGE =
+    'Standardmäßig werdet ihr über neue Chat-Nachrichten per E-Mail informiert. In den Einstellungen könnt ihr eure Benachrichtigungen anpassen. Wir erwarten, dass alle Chat-Nachrichten respektvoll sind. Falls es Beleidigungen oder andere Probleme gibt, meldet euch bitte unter sorgen-eule@lern-fair.de bei uns.';
+const GROUP_CHAT_SYSTEM_MESSAGE =
+    'In diesem Gruppen-Chat können Kursleiter:innen wichtige Ankündigungen und weitere Informationen zum ihrem Kurs mit allen Schüler:innen teilen. Dabei können Kursleiter:innen einstellen, ob auch Schüler:innen in den Gruppen-Chat schreiben dürfen. Der Chat wird 30 Tage nach Ende des Kurses inaktiv. Wir erwarten, dass alle Chat-Nachrichten respektvoll sind. Falls es Beleidigungen oder andere Probleme gibt, meldet euch bitte unter sorgen-eule@lern-fair.de bei uns.';
+
```
timidak, line 22: Can we localize that, with an object? So if other languages will be used in the future, we at least serve a concept. 
const ONE_ON_ONE_SYSTEM_MESSAGE = {
  de: "...."
};

Do we have any other options for localization? @Jonasdoubleyou

timidak: We could put this in another file `chat/localization.ts`, as there are some more messages to come.

Jonasdoubleyou: Having some "i18n like file" sounds good to me :)

integration-tests/matching.ts
```diff
@@ -200,7 +205,7 @@ void test('Create Chat for Match', async () => {
     expectFetch({
         url: 'https://api.talkjs.com/v1/mocked-talkjs-appid/conversations/*/messages',
         method: 'POST',
-        body: '[{"text":"Willkommen im Lern-Fair Chat!","type":"SystemMessage","custom":{"type":"first"}}]',
+        body: '[{"text":"Standardmäßig werdet ihr über neue Chat-Nachrichten per E-Mail informiert. In den Einstellungen könnt ihr eure Benachrichtigungen anpassen. Wir erwarten, dass alle Chat-Nachrichten respektvoll sind. Falls es Beleidigungen oder andere Probleme gibt, meldet euch bitte unter sorgen-eule@lern-fair.de bei uns.","type":"SystemMessage","custom":{"type":"first"}}]',
```
Jonasdoubleyou, line 208: Could also just do "text": "*" so that we do not have to adapt the test each time the message is changed ...

integration-tests/matching.ts
```diff
@@ -3,6 +3,11 @@ import { pupilOne, studentOne } from './user';
 import * as assert from 'assert';
 import { expectFetch } from './mock';
 
+const ONE_ON_ONE_SYSTEM_MESSAGE =
```
Jonasdoubleyou, line 6: unused?

Jonasdoubleyou (approved):

Jonasdoubleyou (dismissed):

Jonasdoubleyou (approved):

PR #733, opened 2023-07-12
Add some sample Queries to the Playground

Jonasdoubleyou:
Might make it easier for new folks to onboard to the Backend.

### DIFF
--- a/graphql/playground.ts
+++ b/graphql/playground.ts
@@ -1,0 +1,207 @@
+import { ApolloServerPluginLandingPageGraphQLPlayground } from 'apollo-server-core';
+import { isDev } from '../common/util/environment';
+
+const userHeader = { authorization: 'Bearer <somerandombearer>' };
+const adminHeader = { authorization: "Basic <btoa('admin' + password)>" };
+
+const endpoint = '/apollo';
+
+const loginQuery = `mutation {
+   # Place some random Bearer token in the tab HTTP Headers, then log in using one of these mutations:
+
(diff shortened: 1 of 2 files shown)


graphql/playground.ts
```diff
@@ -0,0 +1,207 @@
+import { ApolloServerPluginLandingPageGraphQLPlayground } from 'apollo-server-core';
+import { isDev } from '../common/util/environment';
+
+const userHeader = { authorization: 'Bearer <somerandombearer>' };
+const adminHeader = { authorization: "Basic <btoa('admin' + password)>" };
+
+const endpoint = '/apollo';
+
+const loginQuery = `mutation {
+   # Place some random Bearer token in the tab HTTP Headers, then log in using one of these mutations:
```
dhenkel92, line 10: Let's put the actual header as example, so that we can copy-paste it & just replace the token.

dhenkel92 (approved):
That's a really nice improvement for our on-boarding 👍

graphql/playground.ts
Jonasdoubleyou: The header is already prefilled :)

Jonasdoubleyou: Unfortunately we cannot generate a random token as the tabs are shared by all users ...

dhenkel92: Ahh right, haven't seen the header. Nice!

realmayus:
Very nice! 👍

PR #734, opened 2023-07-12
Remove some unused Mails

Jonasdoubleyou:
Also remove deactivating project machtes, they are unused anyways.

Additionally remove the current guest invitation, that was built for BBB anyways


dhenkel92 (approved):
Quite hard to review, but I guess it's fine 😅 
Many things that could break should be covered type tsc

PR #738, opened 2023-07-16
Cleanup dependencies

Jonasdoubleyou:
https://github.com/corona-school/project-user/issues/449


jobs/scheduler.ts
```diff
@@ -1,28 +1,26 @@
 import cron from 'cron';
-import { Mutex } from 'async-mutex';
 import { getLogger } from '../common/logger/logger';
 import { Connection, createConnection, EntityManager } from 'typeorm';
 import { CSCronJob } from './types';
 
 const logger = getLogger();
 
-const activeConnectionMutex = new Mutex();
```
realmayus, line 9: Why do we no longer need the mutex?

realmayus (approved):

jobs/scheduler.ts
Jonasdoubleyou: Because as JavaScript guarantees that only one Job ever runs at the same time, if we synchronously write into the variable we do not need to synchronize.

PR #739, opened 2023-07-16
Prisma migrations

Jonasdoubleyou:
resolves https://github.com/corona-school/project-user/issues/353

After Deployment we need to establish the baseline migration using `npx prisma migrate resolve --applied 20230716161726_initial` (as [documented](https://www.prisma.io/docs/guides/migrate/developing-with-prisma-migrate/add-prisma-migrate-to-a-project#baseline-your-production-environment)).


.github/workflows/node.js.yml
```diff
@@ -13,14 +13,23 @@ jobs:
...
-          POSTGRES_PASSWORD: coronanervt
-          POSTGRES_DB: coronaschool-test
+          POSTGRES_USER: test
+          POSTGRES_PASSWORD: test
+          POSTGRES_DB: integration
         ports:
-          - 5432:5432
+          - 5001:5432
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+      postgres-verify-schema:
```
Jonasdoubleyou, line 25: I couldn't find a good way to start a single Postgres docker-container with multiple DBs, and using the same DB for different things could be error prone ... So two DBs it is ...

.github/workflows/node.js.yml
```diff
@@ -44,11 +53,27 @@ jobs:
...
         run: npx prisma generate
 
-      - run: npm run build --if-present
-      - run: npm test
-      - run: npm run integration-tests
+      - name: Build Backend
+        run: npm run build --if-present
+
+      - name: Verify Prisma Schema against Prisma Migrations
+        run: npm run db:verify-schema
```
Jonasdoubleyou, line 60: We now have a  CI step that validates the schema against the migration, so one cannot push invalid migrations anymore (or forget them)

app.json
```diff
@@ -11,7 +11,7 @@
     "formation": {
         "web": {
             "quantity": 1,
-            "size": "hobby"
+            "size": "basic"
```
Jonasdoubleyou, line 14: Apparently Heroku dropped the hobby plan, so we need to use basic instead (review apps are broken at the moment on all PRs)

common/match/pool.ts
```diff
@@ -522,7 +522,7 @@ export async function getInterestConfirmationRate() {
         where: { status: InterestConfirmationStatus.CONFIRMED },
     });
 
-    return confirmedInterestConfirmations / totalInterestConfirmations;
+    return confirmedInterestConfirmations / totalInterestConfirmations || 0;
```
Jonasdoubleyou, line 525: We now run the integration tests without seeding the DB with test data (as the integration tests should be self-containting), so we now divide by zero here (which we did not handle properly

integration-tests/course.ts
```diff
@@ -156,21 +156,22 @@ export const subcourseOne = test('Create Subcourse', async () => {
 
 void test('Search further instructors', async () => {
     const { client } = await screenedInstructorOne;
+    const { instructor: instructor2 } = await screenedInstructorTwo;
 
     // Partial searches yield no result to not leak infos
-    const partialSearch = await client.request(`query { otherInstructors(search: "melanie", take: 100, skip: 0) { id }}`);
+    const partialSearch = await client.request(`query { otherInstructors(search: "${instructor2.firstname}", take: 100, skip: 0) { id }}`);
```
Jonasdoubleyou, line 162: Here we also relied on test data from dev.ts ...

package.json
```diff
@@ -20,23 +20,21 @@
...
+    "db:create-migration": "prisma migrate dev",
+    "db:reset": "prisma migrate reset",
+    "db:seed": "node -r dotenv/config ./built/seed-db.js",
+    "db:deploy-migrations": "prisma migrate deploy"
   },
   "dependencies": {
     "@aws-sdk/client-s3": "^3.369.0",
     "@aws-sdk/s3-request-presigner": "^3.137.0",
-    "@prisma/client": "4.3.0",
+    "@prisma/client": "4.15.0",
```
Jonasdoubleyou, line 37: While at it I also updated Prisma ...

prisma/migrations/20230716161726_initial/migration.sql
```diff
@@ -0,0 +1,1045 @@
+-- MANUAL CHANGES:
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
```
Jonasdoubleyou, line 2: For some reason we use a Postgres Plugin to generate UUIDs on the database side, which for sure Prisma did not know. Thus `prisma db push` also does not work, but with the changes here `prisma migrate reset` works

prisma/schema.prisma
```diff
@@ -388,7 +388,7 @@ model pupil {
   matchingPriority                             Int                                            @default(0)
   lastUpdatedSettingsViaBlocker                DateTime?                                      @db.Timestamp(6)
   teacherEmailAddress                          String?                                        @db.VarChar
-  registrationSource                           pupil_registrationsource_enum                  @default(dbgenerated("0"))
+  registrationSource                           pupil_registrationsource_enum                  @default(normal)
```
Jonasdoubleyou, line 391: I have no idea how this got into the current state, but `dbgenerated("0")` doesn't make sense as that tries to set the enum to an integer.

release.sh
```diff
@@ -0,0 +1,21 @@
+echo "Starting Release for Environment '$ENV'"
+
+# ----------- STAGING & REVIEW APPS --------------
+if [ \"$ENV\" = 'dev' ]; then
+  echo "=========== Development Release =============="
+  echo "----------- Set up DB Schema -----------------"
+  npm run db:reset -- --force
+  echo "----------- Seed DB with sample users --------"
+  npm run db:seed
```
Jonasdoubleyou, line 9: This moves seeding of the database from the web dyno's start to the release phase, so in DEV we from now on only reset when we deploy and not at each restart anymore (I find that cleaner and makes it easier to test for multiple days)

release.sh
```diff
@@ -0,0 +1,22 @@
...
+  echo "=========== DONE ============================="
+fi
+
+# ----------- PRODUCTION -------------------------
+if [ "$ENV" = 'production' ]; then
+  echo "=========== Development Release =============="
+  echo "----------- Run Migrations -----------------"
+  echo "TODO"
+  # npm run db:deploy-migrations
```
Jonasdoubleyou, line 18: Will turn on productive migrations in a followup change

dhenkel92 (approved):
Amazing 🙂 
Another step towards removing typeorm

Jonasdoubleyou:
PR got too big, and I do not want to bring in Prisma based migrations while we also roll out Chat. Will bring in the trivial changes through a separate PR and will postpone the move to Prisma (but one can just diff this branch with master).

PR #744, opened 2023-07-20
quickfix: get also match appointments

LomyW:
### Description
Currently, match appointments don't appear in the FE appointment list and also not on the dashboard.

### DIFF
--- a/common/appointment/get.ts
+++ b/common/appointment/get.ts
@@ -9,7 +9,6 @@ export const hasAppointmentsForUser = async (user: User): Promise<boolean> => {
     const appointmentsCount = await prisma.lecture.count({
         where: {
             isCanceled: false,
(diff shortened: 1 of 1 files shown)


common/appointment/get.ts
```diff
@@ -112,7 +109,6 @@ const getAppointmentsForUserFromNow = async (userId: User['userID'], take: numbe
     const appointmentsFromNow = await prisma.lecture.findMany({
         where: {
             isCanceled: false,
-            subcourse: { published: true },
```
Jonasdoubleyou, line 115: I think what you actually want is OR: [{ subcourse: null }, { subcourse: { ... }]

Jonasdoubleyou (approved):
(but feel free to bring this fix in as is)

PR #753, opened 2023-07-25
feat: cancel appointments on cancel subcourse

LomyW:


Jonasdoubleyou (approved):

common/appointment/cancel.ts
```diff
@@ -17,32 +18,37 @@ export async function cancelAppointment(user: User, appointment: Appointment) {
...
-        case AppointmentType.internal:
-        case AppointmentType.legacy:
-            break;
+            case AppointmentType.internal:
+            case AppointmentType.legacy:
+                break;
+        }
     }
+
+    await deleteZoomMeeting(appointment);
```
Jonasdoubleyou, line 53: Ah, deleteZoomAppointment assumes zoomMeetingId exists, we should check that here

Jonasdoubleyou (changes requested):

common/appointment/cancel.ts
```diff
@@ -17,32 +19,40 @@ export async function cancelAppointment(user: User, appointment: Appointment) {
...
-            break;
+            case AppointmentType.internal:
+            case AppointmentType.legacy:
+                break;
+        }
+    }
+    if (appointment.zoomMeetingId) {
+        await deleteZoomMeeting(appointment);
+    } else {
+        throw new RedundantError('Appointment has no Zoom Meeting');
```
Jonasdoubleyou, line 56: Is this an error condition? If we cancel the appointment without a zoom meeting, then just cancelling it is fine?

Jonasdoubleyou (approved):

PR #754, opened 2023-07-26
feat: notifications for missed chat messages

LomyW:
### Description
A user should get notifications from Lern-Fair if the user missed some chat messages

### What was done?
- [x] implemented webhook
- [x] added notification actions for 1:1 chats and group chats 

### Tests
- [x] tested if a `concreteNotification` is created after a notification is triggered by TalkJS
- [x] tested in-app notification if preference is set

### DIFF
--- a/common/notification/actions.ts
+++ b/common/notification/actions.ts
@@ -412,6 +412,31 @@ const _notificationActions = {
             appointment: sampleAppointment,
         },
     },
+    missed_one_on_one_chat_message: {
+        description: 'Missed message in 1:1 chat',
+        sampleContext: {
+            sender: sampleUser,
+            recipient: sampleUser,
+            conversationId: sampleAppointment,
+            message: '',
+            totalUnread: '',
+            matchId: '',
+            loginToken: '',
+        },
+    },
+    missed_course_chat_message: {
+        description: 'Missed message in group chat',
+        sampleContext: {
+            sender: sampleUser,
+            recipient: sampleUser,
+            conversationId: sampleAppointment,
+            message: '',
+            totalUnread: '',
+            courseId: '',
+            courseName: '',
+            loginToken: '',
+        },
+    },
 
     user_authenticate: DEPRECATED,
     user_login_email: DEPRECATED,
--- a/common/notification/actions.ts
+++ b/common/notification/actions.ts
@@ -43,6 +43,23 @@ const sampleAppointment = {
     },
 };
 
+const sampleMissedCourseMessage = {
+    sender: 'Leon',
+    conversationId: '1',
+    message: 'Hey',
+    totalUnread: '1',
+    courseId: '1',
+    courseName: 'Gen Z',
+};
+
+const sampleMissedOneOnOneMessage = {
+    sender: 'Max',
+    conversationId: '2',
+    message: 'Hey!',
+    totalUnread: '1',
+    matchId: '1',
+};
+
 const DEPRECATED = {
     description: 'DEPRECATED - DO NOT USE',
 };
@@ -415,26 +432,15 @@ const _notificationActions = {
     missed_one_on_one_chat_message: {
         description: 'Missed message in 1:1 chat',
         sampleContext: {
-            sender: sampleUser,
-            recipient: sampleUser,
-            conversationId: sampleAppointment,
-            message: '',
-            totalUnread: '',
-            matchId: '',
-            loginToken: '',
+            user: sampleUser,
+            message: sampleMissedOneOnOneMessage,
         },
     },
     missed_course_chat_message: {
         description: 'Missed message in group chat',
         sampleContext: {
-            sender: sampleUser,
-            recipient: sampleUser,
-            conversationId: sampleAppointment,
-            message: '',
-            totalUnread: '',
-            courseId: '',
-            courseName: '',
-            loginToken: '',
+            user: sampleUser,
+            message: sampleMissedCourseMessage,
         },
     },
 
--- a/web/controllers/chatNotificationController/index.ts
+++ b/web/controllers/chatNotificationController/index.ts
@@ -6,6 +6,7 @@ import { getPupil, getStudent, getUser } from '../../../common/user';
 import * as Notification from '../../../common/notification';
 import { pupil as Pupil, student as Student } from '@prisma/client';
 import { ChatType, getChatType, getNotificationContext, verifyChatUser } from './util';
+import { createHmac } from 'crypto';
 
 const logger = getLogger('ChatNotification');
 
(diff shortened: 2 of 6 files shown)


common/notification/actions.ts
```diff
@@ -412,6 +412,31 @@ const _notificationActions = {
             appointment: sampleAppointment,
         },
     },
+    missed_one_on_one_chat_message: {
```
LomyW, line 415: action for missed 1:1 messages

common/notification/actions.ts
```diff
@@ -412,6 +412,31 @@ const _notificationActions = {
...
+            sender: sampleUser,
+            recipient: sampleUser,
+            conversationId: sampleAppointment,
+            message: '',
+            totalUnread: '',
+            matchId: '',
+            loginToken: '',
+        },
+    },
+    missed_course_chat_message: {
```
LomyW, line 427: action for missed group chat messages

web/controllers/chatNotificationController/index.ts
```diff
@@ -0,0 +1,46 @@
...
+            const notificationContext = await getNotificationContext(notificationBody);
+            let userToNotify: Pupil | Student;
+            if (recipientUser.pupilId) {
+                userToNotify = await getPupil(recipientUser);
+            } else {
+                userToNotify = await getStudent(recipientUser);
+            }
+
+            const notificationAction = chatType === ChatType.ONE_ON_ONE ? 'missed_one_on_one_chat_message' : 'missed_course_chat_message';
+            await Notification.actionTaken(userToNotify, notificationAction, notificationContext);
```
LomyW, line 40: we dont need to iterate over each participant of a group conversation, because TalkJS itselfs triggers notifications for each participant of the group chat.

common/notification/actions.ts
```diff
@@ -412,6 +412,31 @@ const _notificationActions = {
             appointment: sampleAppointment,
         },
     },
+    missed_one_on_one_chat_message: {
+        description: 'Missed message in 1:1 chat',
+        sampleContext: {
+            sender: sampleUser,
+            recipient: sampleUser,
```
Jonasdoubleyou, line 419: Why do we need the recipient if we implicitly have "user" in the context?

common/notification/actions.ts
```diff
@@ -412,6 +412,31 @@ const _notificationActions = {
...
+        description: 'Missed message in group chat',
+        sampleContext: {
+            sender: sampleUser,
+            recipient: sampleUser,
+            conversationId: sampleAppointment,
+            message: '',
+            totalUnread: '',
+            courseId: '',
+            courseName: '',
+            loginToken: '',
```
Jonasdoubleyou, line 437: This does not need to be in the sampleContext for this action, this is implicitly passed in (like the user)

web/controllers/chatNotificationController/index.ts
```diff
@@ -0,0 +1,46 @@
...
+        }
+        res.status(200).send({ status: 'ok' });
+    } catch (e) {
```
Jonasdoubleyou, line 43: We should log e :)

web/controllers/chatNotificationController/index.ts
```diff
@@ -0,0 +1,46 @@
...
+import { ChatType, getChatType, getNotificationContext, verifyChatUser } from './util';
+
+const logger = getLogger('ChatNotification');
+
+export const chatNotificationRouter = Router();
+
+chatNotificationRouter.post('/chat-notification', handleChatNotification);
+
+async function handleChatNotification(req: Request, res: Response): Promise<void> {
+    try {
```
Jonasdoubleyou, line 17: missing authentication of the other side!

web/controllers/chatNotificationController/index.ts
Jonasdoubleyou, line 40: Can't we just pass recipient in here instead of userToNotify? If not this can stay as is and I have another TODO :)

web/controllers/chatNotificationController/util.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (chatType === ChatType.ONE_ON_ONE) {
+        let courseId: number;
+        if (subcourseConversation?.length === 1) {
+            const subcourse = await getSubcourse(subcourseConversation[0]);
+            courseId = subcourse.courseId;
+        }
+
+        notificationContext = {
+            sender: { firstname: firstnameSender },
+            recipient: { firstname: firstnameRecipient },
```
Jonasdoubleyou, line 51: not needed, one can just use user.firstname

web/controllers/chatNotificationController/util.ts
```diff
@@ -0,0 +1,77 @@
...
+            conversationId: conversation.id,
+            message: messages[0].text,
+            totalUnread: messages.length.toString(),
+            courseId: subcourse.courseId.toString(),
+            courseName: course.name,
+        };
+    }
+
+    logger.info(`CONTEXT: ${JSON.stringify(notificationContext)}`);
```
Jonasdoubleyou, line 74: Maybe a bit more descriptive `logger.info("Created Notification context for chat message", notificationContext);`

web/server.ts
```diff
@@ -93,6 +93,7 @@ export const server = (async function setupWebserver() {
     app.use('/api/attachments', attachmentRouter);
     app.use('/api/certificate', certificateRouter);
     app.use('/api/files', fileRouter);
+    app.use('/api', chatNotificationRouter);
```
Jonasdoubleyou, line 96: Could we move the router to it's own path, like '/api/chat/'? Feels a bit weird to delegate the whole api route to the router.

Jonasdoubleyou (commented):
very nice overall!

web/controllers/chatNotificationController/index.ts
LomyW: Can you tell me more specifically what you mean by that?

web/controllers/chatNotificationController/index.ts
LomyW: Has to be from type `Person`. Thats why I pass `userToNotify``.

web/controllers/chatNotificationController/util.ts
LomyW: the recipient is from talkjs and only has the fullname.

web/controllers/chatNotificationController/index.ts
Jonasdoubleyou: :/ Todo for me

web/controllers/chatNotificationController/util.ts
Jonasdoubleyou: But the recipient is always the user?

web/controllers/chatNotificationController/index.ts
Jonasdoubleyou: How do you know the request comes from TalkJS?

Jonasdoubleyou: https://talkjs.com/docs/Features/Security_Settings/Security_Recommendations/#webhook-integrity

web/controllers/chatNotificationController/util.ts
LomyW: yes, the notification that talkjs triggers always set the `recipient` as the `user` who missed the message. 🙂

Jonasdoubleyou: Yeah but we dont need it twice in the notification context, user is enough.

LomyW: I already removed `recipient` from the context

web/controllers/chatNotificationController/index.ts
```diff
@@ -0,0 +1,65 @@
...
+            }
+
+            const notificationAction = chatType === ChatType.ONE_ON_ONE ? 'missed_one_on_one_chat_message' : 'missed_course_chat_message';
+            await Notification.actionTaken(userToNotify, notificationAction, notificationContext);
+        }
+        res.status(200).send({ status: 'ok' });
+    } catch (error) {
+        if (error instanceof InvalidSignatureError) {
+            logger.info('Invalid Signature');
+            res.status(401).send({ error: 'Unauthorized' });
```
Jonasdoubleyou, line 60: missing return / else

web/server.ts
```diff
@@ -42,7 +44,13 @@ export const server = (async function setupWebserver() {
     });
 
     // Parse Cookies and JSON Bodies:
-    app.use(bodyParser.json());
+    app.use(
+        bodyParser.json({
+            verify: (req: WithRawBody<IncomingMessage>, res, buf) => {
+                req.rawBody = buf;
```
Jonasdoubleyou, line 50: A comment please why this is needed, implementing "verify" seems strange here on a first sight (but it makes sense to keep the raw buffer for the request)

Jonasdoubleyou (approved):

Jonasdoubleyou:
merging this as potentially conflicting with https://github.com/corona-school/backend/pull/702, won't be deployed so just create a followup PR if necessary :)

PR #756, opened 2023-07-28
feat: update zoom meeting on edit appointment

LomyW:
### Description
We should update the zoom meeting if the appointment is edited.
Also we should create the zoom meeting with the duration from the appointment.

### What was done?
- [x] added update zoom meeting function
- [x] adjusted create zoom meeting function (added prop for duration)


common/appointment/update.ts
```diff
@@ -14,31 +15,41 @@ export async function updateAppointment(
...
     const updatedAppointment = await prisma.lecture.update({
-        where: { id: appointment.id },
+        where: { id: id },
         data: { ...appointmentUpdate },
     });
 
     logger.info(`User(${user.userID}) updated Appointment(${appointment.id})`, { appointment, appointmentUpdate });
 
-    const sameDate = !appointmentUpdate.start || appointment.start.toISOString() === appointmentUpdate.start.toISOString();
+    const sameDate = !newStart || start.toISOString() === newStart.toISOString();
```
Jonasdoubleyou, line 35: && (!newDuration || duration === newDuration)

common/appointment/update.ts
```diff
@@ -14,31 +15,41 @@ export async function updateAppointment(
...
     const student = await getStudent(user);
-    switch (appointment.appointmentType) {
+    switch (appointmentType) {
         case AppointmentType.group:
             const subcourse = await prisma.subcourse.findUniqueOrThrow({ where: { id: updatedAppointment.subcourseId }, include: { course: true } });
             const participants = await prisma.subcourse_participants_pupil.findMany({ where: { subcourseId: subcourse.id }, include: { pupil: true } });
+            const subcourseAppointments = await prisma.lecture.findMany({ where: { subcourseId: appointment.subcourseId } });
+            lastDate = subcourseAppointments[subcourseAppointments.length - 1].start;
+
+            logger.info(`Last Date: ${lastDate}`);
```
Jonasdoubleyou, line 50: a bit more descriptive please (or remove)

common/zoom/zoom-scheduled-meeting.ts
```diff
@@ -204,4 +204,47 @@ const getZoomMeetingReport = async (meetingId: string) => {
...
+                    Authorization: `Bearer ${access_token}`,
+                    'Content-Type': 'application/json',
+                },
+                body: body,
+            }),
+        3,
+        1000
+    );
+
+    if (!response.ok && response.status !== 404) {
```
Jonasdoubleyou, line 241: Do we want to ignore updates if the meeting does not exist? Isn't that an error?

common/zoom/zoom-scheduled-meeting.ts
```diff
@@ -204,4 +204,47 @@ const getZoomMeetingReport = async (meetingId: string) => {
...
+        throw new Error(`Zoom - failed to update meeting with ${response.status} ${await response.text()}`);
+    }
+
+    if (response.status === 204) {
```
Jonasdoubleyou, line 245: Why do we conditionally log? What if the response is ok but not 204? Why wouldn't we log in that case?

common/zoom/zoom-scheduled-meeting.ts
LomyW: What should happen if the meeting does not exist?

Jonasdoubleyou: Well should that happen? If not, throw an error

Jonasdoubleyou (approved):

PR #757, opened 2023-07-29
Fix weird 'for await' with early exit in Websockets

Jonasdoubleyou:
Probably caused no real world issues as users are connected with one Websocket only anyways.


timidak (approved):
LGTM 👍🏻 🙂

PR #759, opened 2023-07-31
feat: remove students zoom license

LomyW:
### Description
We have to check whether a student with no active match, is an instructor and has future subcourse appointments and we should then not remove the license.

### Done
- [x] add function to check if zoom license can be removed


common/match/util.ts
```diff
@@ -30,3 +33,34 @@ export function getOverlappingSubjects(pupil: Pupil, student: Student) {
...
+            studentId: studentId,
+            dissolved: false,
+        },
+    });
+
+    const subcourses = await prisma.subcourse.findMany({
+        where: {
+            subcourse_instructors_student: { some: { studentId: studentId } },
+            OR: [
+                { cancelled: true },
```
Jonasdoubleyou, line 52: Why cancelled: true?

common/match/util.ts
```diff
@@ -30,3 +33,34 @@ export function getOverlappingSubjects(pupil: Pupil, student: Student) {
...
+                {
+                    cancelled: false,
+                    lecture: { some: { start: { gte: prevDay } } },
+                },
+            ],
+        },
+    });
+
+    if (Object.keys(subcourses).length === 0 && matchesCount === 0) {
```
Jonasdoubleyou, line 61: subcourses.length

common/match/dissolve.ts
```diff
@@ -49,6 +50,9 @@ export async function dissolveMatch(match: Match, dissolveReason: number, dissol
     const matchDate = '' + +match.createdAt;
     const uniqueId = '' + match.id;
 
+    if (await canRemoveZoomLicense(match.studentId)) {
```
Jonasdoubleyou, line 53: && student.zoomUserId

Jonasdoubleyou (approved):

PR #760, opened 2023-07-31
feat: automatic course promotion on publishing

Leo310:
Resolves https://github.com/corona-school/project-user/issues/741

**Open Questions:**
What differentiates a subcourse from a focus course technically? Because Tobias doesn't want to automatically promote focus courses so we would need to filter them out.

What do you think of this:
> The only thing you could consider for manual course promotions would be to write to people who have not yet been contacted instead of the same or random. But that's more of a nice-to-have, not a hard requirement.

This would require a separate database table keeping track of which pupils were already contacted about a newly published course, right?

### DIFF
--- a/graphql/subcourse/mutations.ts
+++ b/graphql/subcourse/mutations.ts
@@ -167,7 +167,8 @@ export class MutateSubcourseResolver {
         await hasAccess(context, 'Subcourse', subcourse);
 
         await publishSubcourse(subcourse);
-        await sendPupilCoursePromotion(subcourse);
+        const course = await getCourse(subcourse.courseId);
+        if (course.category !== 'focus') await sendPupilCoursePromotion(subcourse);
         return true;
     }
 
--- a/graphql/subcourse/mutations.ts
+++ b/graphql/subcourse/mutations.ts
@@ -168,7 +168,10 @@ export class MutateSubcourseResolver {
 
         await publishSubcourse(subcourse);
         const course = await getCourse(subcourse.courseId);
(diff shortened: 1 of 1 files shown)


Jonasdoubleyou:
The course has a field category with the value 'focus'

Jonasdoubleyou:
"What do you think of this" ... I don't see a benefit, with random selection we also eventually notify anyone.

Jonasdoubleyou:
We could already prevent users from getting the same course promotion by making sure that the context.uniqueId of the notification is unique per course

Leo310:
> The course has a field category with the value 'focus'

Ah damn, how could I miss that

> We could already prevent users from getting the same course promotion by making sure that the context.uniqueId of the notification is unique per course

Ok I will look into it!

Jonasdoubleyou:
> "Isn't the uniqueId unique by default (through subcourse id and Date)?"

Yes it is at the moment, but if we would remove the date then we could filter out. But true, for the bulk notifications we do not have a mechanism at the moment that filters it. We do so in `actionTaken` though. I guess in the bulk notification case we could query `where: { notificationId: id, userId: { in: listOfUsers }, contextId: uniqueId }`, and then filter the list of users by that. But either way, I guess we can postpone that topic for now

Leo310:
@Jonasdoubleyou and should I add some logging to this endpoint and the others that don't have it?

Jonasdoubleyou:
sure, more logging is always better :)

Jonasdoubleyou (approved):

PR #762, opened 2023-07-31
Migrations with Prisma

Jonasdoubleyou:
Use Prisma to generate and run migrations instead of TypeORM.


common/migration/1585628278000-InitialMigration.ts
dhenkel92, line 1: I'm thinking about if we should keep the migrations for documentation purposes

release.sh
```diff
@@ -1,9 +1,23 @@
...
+  npm run db:seed
+  echo "=========== DONE ============================="
+fi
+
+# ----------- PRODUCTION -------------------------
 if [ "$ENV" = 'production' ]; then
-    echo 'Production Deployment, running migrations'
-    npx typeorm migration:run
-else 
-    echo 'Non-Productive Deployment, setting up DB with seeded data'
```
dhenkel92, line 5: Why not if else anymore?

dhenkel92 (approved):

common/migration/1585628278000-InitialMigration.ts
Jonasdoubleyou: I think for the average visitor of the codebase this would rather be confusing, and to travel back in time one can always use version control ...

release.sh
Jonasdoubleyou: I wanted to explicitly check the value of ENV, and then the branches are mutually exclusive anyways

PR #770, opened 2023-08-02
feature(monitoring): add datadog agent & enable tracing

dhenkel92:
There are a couple of environment variables that have to be set in staging/production:

* Everything set in the app.json file
* `DD_API_KEY` & `DD_ENV`

We have to remove the host/service from the heroku log drains


Jonasdoubleyou (approved):
nice!

PR #772, opened 2023-08-06
refactor: Move Notification System to Prisma

Jonasdoubleyou:


common/notification/index.ts
```diff
@@ -82,30 +82,28 @@ const getNotificationChannelPreferences = async (user: User, concreteNotificatio
     return Object.assign({}, channelsBasePreference, channelsUserPreference);
 };
 
-export function getContext(notificationContext: NotificationContext, legacyUser: Person): Context {
+export function getContext(notificationContext: NotificationContext, user: User): Context {
     return {
         ...notificationContext,
-        user: { ...legacyUser, fullName: getFullName(legacyUser) },
+        user: { ...user, fullName: getFullName(user) },
```
Jonasdoubleyou, line 88: This is a potentially breaking change, previously we dropped the whole pupil / student into Mailjet, not sure if there are templates that rely on this ... Either way, I guess we then should rather adapt the templates than always send over the whole pupil / student object.

dhenkel92 (approved):

PR #774, opened 2023-08-07
feat: ad-hoc meeting for matches

LomyW:
refers to: https://github.com/corona-school/user-app/pull/320

### Description
Instructors should be able to start an ad-hoc meeting that lasts 30 minutes.

### What was done?
- [x] add mutation to create an ad-hoc meeting
- [x] add notification for pupil


common/match/create.ts
```diff
@@ -91,3 +94,33 @@ export async function createMatch(pupil: Pupil, student: Student, pool: Concrete
 
     logger.info(`Created Match(${match.uuid}) for Student(${student.id}) and Pupil(${pupil.id})`);
 }
+
+export async function createAdHocMeeting(matchId: number, user: User) {
```
timidak, line 98: The only thing I'm not sure of, is where this function is placed. Maybe appointment create could be a better fit. @Jonasdoubleyou any thoughts on this?

common/match/create.ts
```diff
@@ -91,3 +94,33 @@ export async function createMatch(pupil: Pupil, student: Student, pool: Concrete
...
+    const { pupilId, studentId } = match;
+
+    const pupil = await getPupil(pupilId);
+    const student = await getStudent(studentId);
+
+    const appointment: AppointmentCreateMatchInput[] = [
+        {
+            title: `Sofortbesprechung - ${pupil.firstname} und ${student.firstname} `,
+            matchId: matchId,
+            start: new Date(),
```
timidak, line 109: This could cause time issues again, because we're creating a date in the backend which runs in UTC.

graphql/match/mutations.ts
```diff
@@ -52,4 +53,17 @@ export class MutateMatchResolver {
...
+
+    @Mutation((returns) => JSONResolver, { nullable: true })
+    @AuthorizedDeferred(Role.ADMIN, Role.OWNER)
+    async matchCreateAdHocMeeting(@Ctx() context: GraphQLContext, @Arg('matchId', (type) => Int) matchId: number) {
+        const { user } = context;
+        const match = await getMatch(matchId);
+        await hasAccess(context, 'Match', match);
+        if (user.pupilId) {
+            return;
+        }
```
timidak, line 65: We should throw an AuthorizationError here, so that we get some context in the backend, if nothing happens

common/appointment/util.ts
```diff
@@ -25,3 +25,5 @@ export function getAppointmentForNotification(appointment: Appointment, original
         original: original ? getAppointmentForNotification(original) : undefined,
     };
 }
+
+export class AuthorizationError extends Error {}
```
Jonasdoubleyou, line 29: Not needed, you could use the AuthenticationError class exposed by Apollo

Jonasdoubleyou (approved):

PR #783, opened 2023-08-09
HTOFIX Ignore Message Rendering Errors

Jonasdoubleyou:
When changing backend logic, sometimes the available variables change, and we only validate the current backend logic against the currently created notifications. Thus concrete notifications that were sent from the old backend logic and persisted in the database might not match the current templates. To avoid that this messes up the user-app, we silently ignore these instead (and just log backend errors for the tech-team to handle).

### DIFF
--- a/common/notification/messages.ts
+++ b/common/notification/messages.ts
@@ -130,7 +130,7 @@ export async function getMessage(
     let navigateTo: string | undefined;
     try {
         if (template.navigateTo) {
-            navigateTo = template.headline(context);
+            navigateTo = template.navigateTo(context);
         }
     } catch (error) {
         logger.error(
--- a/common/notification/messages.ts
+++ b/common/notification/messages.ts
@@ -101,12 +101,68 @@ export async function getMessage(
     const user = await getUser(concreteNotification.userId);
     const context = getContext(concreteNotification.context as NotificationContext, user);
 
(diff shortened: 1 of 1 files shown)


common/notification/messages.ts
```diff
@@ -101,12 +101,68 @@ export async function getMessage(
     const user = await getUser(concreteNotification.userId);
     const context = getContext(concreteNotification.context as NotificationContext, user);
 
+    let headline = '';
+    try {
+        headline = template.headline(context);
+    } catch (error) {
+        logger.error(
+            `Failed to render Headline Template of Notification(${concreteNotification.notificationID}) for ConcreteNotification(${concreteNotification.id})`,
```
dhenkel92, line 109: 🙃 It's much easier to filter messages without variables

dhenkel92 (approved):

common/notification/messages.ts
Jonasdoubleyou: Bad Habits ...

dhenkel92: 😆 but you didn't change it?

Jonasdoubleyou: Yeah there are many other logs as well that need to be adapted, and had to do a few other fixes ...

PR #784, opened 2023-08-09
HOTFIX: Make Rate Enforcement Errors well known

Jonasdoubleyou:
Currently Rate Limiting is reported as an error

### DIFF
--- a/graphql/rate-limit.ts
+++ b/graphql/rate-limit.ts
@@ -2,6 +2,7 @@ import { getLogger } from '../common/logger/logger';
 import { createMethodDecorator } from 'type-graphql';
 import { Role } from './authorizations';
 import { GraphQLContext } from './context';
+import { ForbiddenError } from 'apollo-server-express';
 
 const log = getLogger('GraphQL Rate Limiting');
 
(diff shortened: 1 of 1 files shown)


graphql/rate-limit.ts
```diff
@@ -34,7 +35,7 @@ export function RateLimit(name: string, max: number, interval: number /* in ms *
 
         if (count > max) {
             log.warn(`Blocked ${context.ip} from accessing ${name} as maximum of ${max} was reached`);
-            throw new Error(`RateLimit Enforcement`);
+            throw new ForbiddenError(`RateLimit Enforcement`);
```
dhenkel92, line 38: Btw. I always wanted to ask why we have a rate limiter. Have we had any problems with too many requests in the past?

dhenkel92 (approved):

graphql/rate-limit.ts
Jonasdoubleyou: @dhenkel92 no, mainly to prevent email / password enumeration, spam on the registration interface (hasn't happened yet, but better safe than sorry). Also it is abused in case the user app gets stuck in a retry loop and starts sending hundreds of backend requests. We also only Rate Limit a few specific endpoints for this (as rejecting unauthenticated users on authenticated queries is fast enough)

PR #791, opened 2023-08-11
fix: datadog configuration

dhenkel92:


utils/environment.ts
```diff
@@ -10,6 +10,10 @@ export function getEnvironment(): string {
     return process.env.ENV || 'dev';
 }
 
+export function getDDEnvironment(): string {
+    return process.env.DD_ENV || 'dev';
```
dhenkel92, line 14: Will be used to distinguish between "staging" and review apps

Jonasdoubleyou (approved):

PR #796, opened 2023-08-18
feat: update meta data of finished conversations

LomyW:
- [x] added update function to cronjob to flag old conversations
- [x] added update for reactivation of a conversation
- [x] added enum for finished reasons

### DIFF
--- a/jobs/periodic/flag-old-conversations/index.ts
+++ b/jobs/periodic/flag-old-conversations/index.ts
@@ -48,7 +48,7 @@ export default async function flagInactiveConversationsAsReadonly() {
     const conversationsToFlag: conversationsToDeactivate[] = [];
 
     for (const conversation of conversations.data) {
-        let shouldMarkAsReadonly: boolean;
+        let shouldMarkAsReadonly: boolean = true;
 
         // to prevent to flag already deactivated chats we check if the conversation is already readonly (only readMembers)
         const countParticipants = countChatParticipants(conversation);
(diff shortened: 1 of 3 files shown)


jobs/periodic/flag-old-conversations/index.ts
```diff
@@ -55,42 +55,58 @@ export default async function flagInactiveConversationsAsReadonly() {
...
-            const subcourseIds: number[] = JSON.parse(conversation.custom.subcourse);
-            const allSubcoursesActive = await Promise.all(subcourseIds.map((id) => isActiveSubcourse(id)));
-            shouldMarkAsReadonly = allSubcoursesActive.every((active) => active === false);
-        }
-        if (conversation.custom.prospectSubcourse) {
-            const prospectSubcourses: number[] = conversation.custom.prospectSubcourse;
-            const allProspectSubcoursesActive = await Promise.all(prospectSubcourses.map((id) => isActiveSubcourse(id)));
-            shouldMarkAsReadonly = allProspectSubcoursesActive.every((active) => active === false);
-        }
+        if (!isChatReadOnly) {
```
Jonasdoubleyou, line 61: I'd prefer the short curcuit with return instead of an indirect if else

jobs/periodic/flag-old-conversations/index.ts
```diff
@@ -55,42 +55,58 @@ export default async function flagInactiveConversationsAsReadonly() {
...
+            if (conversation.custom.subcourse) {
+                const subcourseIds: number[] = JSON.parse(conversation.custom.subcourse);
+                const allSubcoursesActive = await Promise.all(subcourseIds.map((id) => isActiveSubcourse(id)));
+                shouldMarkAsReadonly = allSubcoursesActive.every((active) => active === false);
+            }
+            if (conversation.custom.prospectSubcourse) {
+                const prospectSubcourses: number[] = conversation.custom.prospectSubcourse;
+                const allProspectSubcoursesActive = await Promise.all(prospectSubcourses.map((id) => isActiveSubcourse(id)));
+                shouldMarkAsReadonly = allProspectSubcoursesActive.every((active) => active === false);
```
Jonasdoubleyou, line 70: This potentially overrides shouldMarkAsReadonly=false with true

LomyW: why? Any other suggestions?

jobs/periodic/flag-old-conversations/index.ts
LomyW: However, this has prevented further conversations from being checked once a conversation is already readonly. Do you know why?

jobs/periodic/flag-old-conversations/index.ts
Jonasdoubleyou: shouldMarkAsReadonly = shouldMarkAsReadonly && ...

jobs/periodic/flag-old-conversations/index.ts
Jonasdoubleyou: Ah true, could use `continue` instead of `return`, or an if/else

Jonasdoubleyou (approved):

PR #797, opened 2023-08-18
Generalize actionTaken to actionTakenAt

Jonasdoubleyou:
This way we can schedule actions in the past (i.e. when new notifications get added we can schedule them once for previous actions) and in the future (i.e. if we know that an appointment will happen at a future point in time). Additionally we now also support negative delays (so if we schedule an action in the future with a negative delay, it will be sent at some point in the future but before the event). 

Additionally:
- Adds a parameter `noDuplicates` to actions that prevents Notifications to be sent out if the same action was already triggered. Previously this was only the case for Notifications without a delay where a uniqueId was set, in which case an error was thrown and logged - which with the generalization creates some inconsistencies (i.e. actionTakenAt(oneHourAgo, ...) with a delay one hour would run into the assertion while actionTakenAt(now, ...) would not)
- Adds a parameter `dryRun` with which one can calculate the notifications to send without actually sending them - This is used in the bulkRun feature
- Add some integration tests for the notification system itself (we already have some implicit coverage through other tests, but they mainly cover immeadiate notifications and do not cover reminders)
- Restructures the integration_test folder, separating utilities and test files, and renaming test files to be sequential


realmayus (approved):
Very nice!

LomyW (approved):
Nice, well done! 🎉

Jonasdoubleyou:
I'll bring this in on the weekend, this change has the potential to mess some things up (and I'm busy the next days)

common/notification/index.ts
```diff
@@ -456,29 +454,102 @@ export async function actionTaken<ID extends ActionID>(
...
+                //                   (negative delay)      (positive delay)
+                sendAt += notification.delay * HOURS_TO_MS;
+            }
 
-        // Trigger notifications that are supposed to be directly sent on this action
-        for (const directSend of directSends) {
-            const concreteNotification = await createConcreteNotification(directSend, user, notificationContext, attachments);
-            await deliverNotification(concreteNotification, directSend, user, notificationContext, attachments);
+            if (sendAt < now && notification.interval) {
+                //   X ----X-| (NOW) |-> X
```
timidak, line 503: These "graphs" helped me a lot to wrap my head around this stuff! Nice one!

common/notification/index.ts
```diff
@@ -456,29 +454,102 @@ export async function actionTaken<ID extends ActionID>(
...
+        // --------------- Determine which Notifications to send directly and which to schedule ---------
 
-        const reminders = relevantNotifications.toSend.filter((it) => it.delay);
-        const directSends = relevantNotifications.toSend.filter((it) => !it.delay);
+        for (const notification of relevantNotifications.toSend) {
+            if (noDuplicates) {
+                assert(notificationContext.uniqueId, 'If noDuplicates is set, a uniqueId shall be set');
 
-        logger.debug(`Notification.actionTaken found reminders ${reminders.map((it) => it.id)} and directSends ${directSends.map((it) => it.id)}`);
+                const existingNotifications = await prisma.concrete_notification.count({
```
timidak, line 483: The name `existingNotifications` is misleading imho. Maybe `existingDuplicates` or `sameNotifications`?

timidak: or we cast it to a boolean and call it `hasExistingDuplicates`.

timidak (approved):
Nice work @Jonasdoubleyou 🎉

PR #807, opened 2023-08-22
HOTFIX: Subcourse isCancelled predicate

Jonasdoubleyou:
isCancelled can also be null, which is equivalent to false. This removed some older courses from the public subcourse page


dhenkel92 (approved):

Jonasdoubleyou:
jumped to conclusions, isCanceled is false ...

PR #808, opened 2023-08-22
fix: Subcourse cancellation

Jonasdoubleyou:
Every subcourse has one last appointment, thus cancelling a subcourse will always fail.


LomyW (approved):
👍🏼

PR #810, opened 2023-08-22
fix(automatic-promotion): moved promotion logic into common

Leo310:


Jonasdoubleyou (approved):
please also adapt the text of the approve button in Retool, to make support aware that the button also publishes & promotes

Leo310:
image
Why are there even two buttons for this? Atm both would do the same thing no?

Jonasdoubleyou:
yeah they would 😅

Jonasdoubleyou:
yes makes sense

Leo310:
OK, I removed it in v1.12.0. If this is fine, I will send a message to tech-updates @Jonasdoubleyou

PR #811, opened 2023-08-24
ci(linting): remove (nearly) all linting warnings

Leo310:
still need to resolve dependency-cycle warnings.


Leo310:
The remaining import/no-cycle warnings would require quite some restructuring.
@Jonasdoubleyou @dhenkel92 should I tackle them or do you think we should ignore them for now?

.eslintrc
dhenkel92, line 1: Could you try to add the recomented rules and see what happens? 
I think we are missing some important rules, like `prefer-const`.

```
extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended'],
```

https://typescript-eslint.io/getting-started/#step-2-configuration

package.json
dhenkel92, line 1: We should also think about upgrading the eslint related dependencies.

dhenkel92 (commented):
> The remaining import/no-cycle warnings would require quite some restructuring. @Jonasdoubleyou @dhenkel92 should I tackle them or do you think we should ignore them for now?

Maybe we should add the "/* eslint-disable import/no-cycle */" comment to these files as well. This would remove the warnings, but makes sure that we don't introduce new cyclic dependencies.

package.json
```diff
@@ -43,6 +43,7 @@
     "@aws-sdk/client-s3": "^3.369.0",
     "@aws-sdk/s3-request-presigner": "^3.137.0",
     "@prisma/client": "4.15.0",
+    "@types/qrcode": "^1.5.1",
```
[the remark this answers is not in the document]
Leo310: my linter was complaining that the declaration file was missing. so installed the types of the qrcode package

.eslintrc
Leo310: After running `eslint --fix` there still are 968 problems haha.  There are a lot of "no-unused-vars" warnings caused mainly by this:
image
and this:
image

package.json
Leo310: what do you mean by upgrading it? Most of the packages dont have a fixed version:
image

.eslintrc
dhenkel92: 😂
What if we remove the no unused car rule?

package.json
dhenkel92: But it doesn't apply for the major version

package.json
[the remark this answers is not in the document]
dhenkel92: Ahh nice 😄

.eslintrc
Jonasdoubleyou: Hm we could remove the "unused function parameter" but I think it actually increases readability here (especially @FieldResolver(returns => Type))

Jonasdoubleyou: But I would generally try to first get to 0 warnings, then enforce no linter warnings in CI. Then we can still decide to turn on further rules

dhenkel92: What is the advantage of this?

package.json
Leo310: Ah true

.eslintrc
Leo310: i turned the no-unused-var rule off and now there are only 156 problems left.
Most of them are unexpected-any's ("no-explicit-any") errors

dhenkel92: What if you remove this one as well?
We could create a follow up ticket to fix it as it's only metadata.

Leo310:
Any idea why the integration tests fail now?

Leo310:
Only 3 more errors to resolve 🎉

graphql/me/mutation.ts
```diff
@@ -196,11 +184,7 @@ export class MutateMeResolver {
     @Mutation((returns) => Student)
     @Authorized(Role.UNAUTHENTICATED, Role.ADMIN)
     @RateLimit('RegisterStudent', 10 /* requests per */, 5 * 60 * 60 * 1000 /* 5 hours */)
-    async meRegisterStudent(
-        @Ctx() context: GraphQLContext,
-        @Arg('data') data: RegisterStudentInput,
-        @Arg('noEmail', { nullable: true }) noEmail: boolean = false
-    ) {
+    async meRegisterStudent(@Ctx() context: GraphQLContext, @Arg('data') data: RegisterStudentInput, @Arg('noEmail', { nullable: true }) noEmail = false) {
```
Jonasdoubleyou, line 187: The :boolean is needed here for TypeGraphQLs reflection (unlile Typescript itself which infers the type from the default parameter) ...

dhenkel92: Then let's suppress this one as well 🙂

Leo310:
@Jonasdoubleyou not sure why the integration test failed at the pupilCreatesChat mutation. any idea? Maybe has something to do with the merge

common/chat/helper.ts
```diff
@@ -166,7 +168,7 @@ const convertConversationInfosToString = (conversationInfos: ConversationInfos):
     };
 
     for (const key in conversationInfos.custom) {
-        if (conversationInfos.custom.hasOwnProperty(key)) {
+        if (Object.prototype.hasOwnProperty.call(key)) {
```
Jonasdoubleyou, line 171: WTF, `.call(` is missing the first argument, should be `.call(conversationInfos, key)` ...

Leo310: ahh lost, overlooked this when fixing all the linting stuff. mb

Leo310: didnt seem to fix it :/

Jonasdoubleyou (dismissed):
looks good, haven't yet checked the changes to package.json. Was it just reordered?

Leo310:
> looks good, haven't yet checked the changes to package.json. Was it just reordered?

The only thing that I changed in the package.json was increasing the eslint package version numbers and including the qrcode/types because they were missing. Not sure what all the other changes are

dhenkel92 (commented):
Amazing work! 😍 
It was much more effort than expected, but I'm confident that the linter & tsc will help us to find problems before they go to production.

Nevertheless, I'm also not sure what happened to the package.json. From the diff it seems like some parts got copied.
Besides that, I have the feeling that we should reserve some time to at least click a couple of times through staging and test different scenarios. I'm especially afraid of the Regex changes 😄

Jonasdoubleyou (approved):
let's go!

PR #815, opened 2023-08-27
fix: db:setup command correctly loads .env file

Jonasdoubleyou:


package.json
```diff
@@ -29,7 +29,8 @@
     "prettier": "./prettier-commits.sh",
     "prettier-on": "prettier --write",
     "test": "jest",
-    "db:setup": "psql \"$DATABASE_URL?sslmode=prefer\" -c 'CREATE EXTENSION IF NOT EXISTS \"uuid-ossp\";'",
+    "db:setup": "dotenv npm run db:setup:run",
+    "db:setup:run": "psql \"$DATABASE_URL?sslmode=prefer\" -c 'CREATE EXTENSION IF NOT EXISTS \"uuid-ossp\";'",
```
dhenkel92, line 33: @Maascha had the problem, that the database did not exist yet. Should we add a command to create it as well?

Jonasdoubleyou: Hm, one anyways has to create a user and grant them permissions (either to create a db or on the created db)

dhenkel92: hmm... that's true.
Let's see if anyone is running into a similar issue

dhenkel92 (approved):

PR #818, opened 2023-08-27
fix: Move addGroupAppointmentsParticipant and addParticipant to commons

Jonasdoubleyou:
Otherwise code that doesn't go through the GraphQL layer, i.e. filling a course, does not properly add the pupils to chat & appointments.


timidak (approved):
Nice clean-up! Thanks @Jonasdoubleyou 👍🏻

PR #819, opened 2023-08-27
feat: Track Screeners involved in Pupil Screening

Jonasdoubleyou:


prisma/migrations/20230827190647_pupil_screening_screeners/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "pupil_screening" ADD COLUMN     "screenerIds" INTEGER[] DEFAULT ARRAY[]::INTEGER[];
```
dhenkel92, line 2: The migrations are not executed automatically yet, right?

dhenkel92 (approved):

prisma/migrations/20230827190647_pupil_screening_screeners/migration.sql
Jonasdoubleyou: right, I'll do it manually with this one, and enable it if that worked well

PR #822, opened 2023-08-28
doc: Document local setup, adapt .env.example to work out of the box

Jonasdoubleyou:
AWS_REGION=eu is needed as otherwise the backend fails to start


dhenkel92 (approved):
🙇

PR #824, opened 2023-08-29
fix: Subcourse Appointment participant Management

Jonasdoubleyou:


timidak (approved):
Nice work! The logs are pretty good to follow 💯

PR #825, opened 2023-08-30
fix: scheduled appointment meetings

LomyW:
- Create one Zoom Meeting for each Appointment (instead of one per Course), this makes lifecycle management easier
- Changing Subcourse Instructors adapts Zoom Meeting Hosts
- Cache Zoom Access Tokens per Scope, synchronize their creation
- Add Mutation for manually creating Zoom Meetings for an Appointment

### DIFF
--- a/common/appointment/participants.ts
+++ b/common/appointment/participants.ts
@@ -1,7 +1,8 @@
 import { getLogger } from '../logger/logger';
 import { Lecture } from '../entity/Lecture';
 import { prisma } from '../prisma';
-import { User } from '../user';
+import { User, userForStudent } from '../user';
+import { addOrganizerToZoomMeeting } from '../zoom/scheduled-meeting';
 
 const logger = getLogger('Appointment Participants');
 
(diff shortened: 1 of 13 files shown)


Jonasdoubleyou (approved):
nice!

Jonasdoubleyou:
I think the integration test just showed that `getAccessToken` is not properly synchronized and multiple access tokens might be created in parallel. I'd suggest to use a Promise for synchronization:

```ts
const tokenFetcher: Promise<string> | null = null;

function getAccessToken(): Promise<string> {
  if (tokenFetcher) return tokenFetcher;

  return tokenFetcher = (async function fetchToken() {
    // async logic
  })();
}
```

common/appointment/participants.ts
```diff
@@ -65,12 +66,13 @@ export async function addGroupAppointmentsOrganizer(subcourseId: number, organiz
             continue;
         }
 
-        await prisma.lecture.update({ where: { id: lecture.id }, data: { participantIds: { push: organizerId } } });
+        await prisma.lecture.update({ where: { id: lecture.id }, data: { organizerIds: { push: organizerId } } });
```
Jonasdoubleyou, line 69: Whoops, that should not have happened ...

Jonasdoubleyou (approved):

PR #827, opened 2023-09-07
feat: appointment reminder

LomyW:
- scheduled appointment reminders 12h before the appointment start 
https://github.com/corona-school/project-user/issues/851

### DIFF
--- a/common/appointment/create.ts
+++ b/common/appointment/create.ts
@@ -14,7 +14,7 @@ import { User, getStudentsFromList, userForPupil, userForStudent } from '../user
 import { getLecture, getMatch, getPupil, getStudent } from '../../graphql/util';
 import { PrerequisiteError, RedundantError } from '../../common/util/error';
 import { AppointmentType } from '../../common/entity/Lecture';
-import { getAppointmentForNotification } from './util';
+import { getContextForGroupAppointmentReminder, getContextForMatchAppointmentReminder } from './util';
 
 const logger = getLogger();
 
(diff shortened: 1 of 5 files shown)


common/appointment/create.ts
```diff
@@ -90,6 +91,21 @@ export const createMatchAppointments = async (matchId: number, appointmentsToBeC
             student,
             matchId: matchId.toString(),
         });
+
+        for (const appointment of createdMatchAppointments) {
+            await Notification.actionTakenAt(new Date(appointment.start), userForPupil(pupil), 'pupil_match_appointment_reminder', {
+                uniqueId: appointment.id.toString(),
+                matchId: appointment.matchId.toString(),
+                appointment: await getAppointmentForNotification(appointment),
```
Jonasdoubleyou, line 99: Could be moved to a variable, to avoid repeated calls

common/notification/actions.ts
```diff
@@ -522,7 +522,42 @@ const _notificationActions = {
         description: 'User added a match ad-hoc meeting',
         sampleContext: { student: sampleUser, appointmentId: '1', appointment: { url: '/video-chat/1/match' } },
     },
-
+    pupil_match_appointment_reminder: {
+        description: 'Remind pupil of upcoming match appointment',
```
Jonasdoubleyou, line 526: I would prefer to semantically decouple the reminder (the notification) from the action, so rather name it "appointment_starts" as that's what the action signifies. Then it also makes more sense to add an action "1 hour before 'appontment starts'"

Jonasdoubleyou (approved):
looks good!

Jonasdoubleyou (approved):

PR #835, opened 2023-09-12
fix: ignore 404 for zoom report

LomyW:


common/zoom/scheduled-meeting.ts
```diff
@@ -196,6 +196,9 @@ const getZoomMeetingReport = async (meetingId: string) => {
 
     if (response.status === 200) {
         logger.info(`Zoom - The Zoom Meeting ${meetingId} report was received.`);
+    } else if (response.status === 404) {
+        logger.info(`Zoom - Meeting report could not be received because only one participant was in the meeting.`);
```
timidak, line 200: I'd rather log the response.text() 

```
logger.info(`Zoom - Meeting report could not be received: ${await response.text()}`);
```

common/zoom/scheduled-meeting.ts
```diff
@@ -196,6 +196,9 @@ const getZoomMeetingReport = async (meetingId: string) => {
...
+        return null;
```
dhenkel92, line 201: Is the function that is calling `getZoomMeetingReport` ready to receive null?

common/appointment/create.ts
```diff
@@ -208,11 +208,14 @@ const createZoomMeetingForAppointmentWithHosts = async (
 export const saveZoomMeetingReport = async (appointment: Lecture) => {
     const result = await getZoomMeetingReport(appointment.zoomMeetingId);
 
+    if (!result) {
+        logger.info(`Meeting report could not be saved for appointment (${appointment.id})`);
```
dhenkel92, line 212: What about an early return here or does it make sense to still try to save the null value in the DB?

PR #836, opened 2023-09-16
fix: Silence Notification System during Seeding

Jonasdoubleyou:
Otherwise we get mails for every deploy


common/notification/index.ts
```diff
@@ -406,6 +410,11 @@ export async function actionTakenAt<ID extends ActionID>(
     noDuplicates = false,
     attachments?: AttachmentGroup
 ) {
+    if (SILENCE_NOTIFICATION_SYTEM) {
```
dhenkel92, line 413: maybe just to be sure ?
```suggestion
    if (SILENCE_NOTIFICATION_SYTEM && isNotProdEnv) {
```

Jonasdoubleyou: good idea

PR #856, opened 2023-10-07
feature: use prisma migrate reset instead of prisma db push

dhenkel92:
Instead of pushing the schema to local/review dbs, we will use the migrations generated by prisma. The adventage is that we don't need the db:setup scripts anymore as the uuid extenstion is already enabled in the first init script.
This makes local setups much easier to run.


package.json
```diff
@@ -29,12 +29,8 @@
...
         "integration-tests:debug": "node -r source-map-support/register -r dotenv/config ./built/integration-tests/index.js dotenv_config_path=.env.integration-tests.debug",
         "test": "jest",
-        "db:setup": "dotenv npm run db:setup:run",
-        "db:setup:run": "psql \"$DATABASE_URL?sslmode=prefer\" -c 'CREATE EXTENSION IF NOT EXISTS \"uuid-ossp\";'",
-        "db:setup:heroku": "psql \"$DATABASE_URL?sslmode=prefer\" -c 'CREATE EXTENSION IF NOT EXISTS \"uuid-ossp\" WITH SCHEMA heroku_ext;'",
-        "db:reset": "prisma db push --skip-generate",
-        "db:reset-hard": "prisma db push --skip-generate --force-reset",
-        "db:reset-for-migration": "prisma migrate reset",
+        "db:reset": "prisma migrate reset --skip-generate",
+        "db:reset-hard": "prisma migrate reset -f --skip-generate",
```
dhenkel92, line 33: @Jonasdoubleyou Was there a special reason to use db push instead of migrate reset? 
When using the reset function we don't need the psql commands anymore. Review environments / integration tests and local setups seem to work.

Jonasdoubleyou: I think push takes the schema, so can be used to quickly check local changes to the schema. But I'm fine to change it and always require a migration to be generated first.

Jonasdoubleyou (approved):

package.json
dhenkel92: We can add it back when we see that the new workflow is much more effort.

PR #857, opened 2023-10-10
feature: replace statsd with prometheus endpoint

dhenkel92:
This PR will replace the statsd client with a prometheus `/metrics` endpoint.
It basically has two advantages:

1) Statsd is pushing metrics, which means that if something happens, we'll get a point in time. The prometheus endpoint on the other side is fetching all metrics every 30s. This helps to get a constant stream, so that the information can be properly graphed.
2) The prometheus client is type safe. So we are sure that all labels are always present.


dhenkel92:
From single values (left) to a constant stream (right)

image

Jonasdoubleyou (dismissed):

PR #859, opened 2023-10-14
Add TRUSTED_SCREENER role 

realmayus:
Add TRUSTED_SCREENER role which grants privileges to log in with users' accounts

Closes https://github.com/corona-school/project-user/issues/928


Jonasdoubleyou (approved):
nice!

PR #864, opened 2023-10-16
fix: invalidate confirmation interest on match request

dhenkel92:


common/match/pool.ts
```diff
@@ -253,7 +253,7 @@ export const TEST_POOL = {
 const _pools = [
     {
         name: 'lern-fair-now',
-        confirmInterest: false,
+        confirmInterest: true,
```
Jonasdoubleyou, line 256: This needs to stay false

common/match/pool.ts
```diff
@@ -386,9 +386,6 @@ export async function runMatching(poolName: string, apply: boolean, _toggles: st
     if (apply) {
         const startCommit = Date.now();
         for (const match of matches) {
-            if (pool.confirmInterest && match.pupil.openMatchRequestCount <= 1) {
-                await removeInterest(match.pupil);
-            }
```
Jonasdoubleyou, line 391: This can stay

common/match/create.ts
```diff
@@ -50,6 +51,7 @@ export async function createMatch(pupil: Pupil, student: Student, pool: Concrete
     });
 
     await invalidateAllScreeningsOfPupil(pupil.id);
+    await removeInterest(pupil);
```
Jonasdoubleyou, line 54: This is the actual fix

common/match/pool.ts
dhenkel92: Is there a case where it's needed?

common/match/pool.ts
dhenkel92: Then I misunderstood our conversation 😅 
I'll revert the change.

common/match/pool.ts
Jonasdoubleyou: Ah nevermind, we call createMatch right after which now calls it anyways

common/match/pool.ts
Jonasdoubleyou: Yeah we do not want to invite new users for interest confirmation, which this toggle controls

Jonasdoubleyou (approved):

PR #866, opened 2023-10-20
feat: Achievements

LomyW:
# 🏆 Achievement feature branch 
do not merge yet...


common/achievement/index.ts
```diff
@@ -0,0 +1,123 @@
...
+        user: user,
+        context,
+    };
+    await trackEvent(actionEvent);
+
+    for (const [, group] of templatesByGroups) {
+        let achievementToCheck: AchievementToCheck;
+        for (const template of group) {
+            const userAchievement = await getOrCreateUserAchievement(template, user.userID, context);
+            if (userAchievement.achievedAt === null || userAchievement.recordValue) {
```
dhenkel92, line 39: I think `userAchievement` can be undefined if it's the last in group.

common/achievement/create.ts
```diff
@@ -0,0 +1,96 @@
...
+async function createNextUserAchievement<ID extends ActionID>(
+    templatesForGroup: Achievement_template[],
+    nextStepIndex: number,
+    userId: string,
+    context: SpecificNotificationContext<ID>
+) {
+    const nextStepTemplate = templatesForGroup[nextStepIndex];
+    // Here a user template is created for the next template in the group. This is done to always have the data availible for the next step.
+    // This could mean to, for example, have the name of a match partner that is not yet availible due to a unfinished matching process.
+    if (nextStepTemplate && nextStepTemplate.isActive) {
```
dhenkel92, line 78: Shouldn't we check if a template is inactive while fetching them from the DB?

common/achievement/create.ts
```diff
@@ -0,0 +1,96 @@
...
+                groupOrder: nextStepTemplate.groupOrder,
+                context: context ? context : Prisma.JsonNull,
+                template: { connect: { id: nextStepTemplate.id } },
+                recordValue: nextStepTemplate.type === 'STREAK' ? 0 : null,
+            },
+            select: { id: true, userId: true, context: true, template: true, achievedAt: true, recordValue: true },
+        });
+        return createdUserAchievement;
+    }
+    const nextUserAchievement = await createNextUserAchievement(templatesForGroup, nextStepIndex + 1, userId, context);
```
dhenkel92, line 92: If we keep it like this, we need an abort check like this in the beginning. Otherwise, we might crash the server.
```
if (templatesForGroup.length <= nextStepIndex){
    return;
}
```

common/achievement/create.ts
```diff
@@ -0,0 +1,96 @@
...
+    context: SpecificNotificationContext<ID>
+) {
+    const nextStepTemplate = templatesForGroup[nextStepIndex];
+    // Here a user template is created for the next template in the group. This is done to always have the data availible for the next step.
+    // This could mean to, for example, have the name of a match partner that is not yet availible due to a unfinished matching process.
+    if (nextStepTemplate && nextStepTemplate.isActive) {
+        const createdUserAchievement = await prisma.user_achievement.create({
+            data: {
+                userId: userId,
+                group: nextStepTemplate.group,
```
dhenkel92, line 82: How do we distinguish two match groups? Do we need to use handlebars for the group name as well?

common/achievement/create.ts
```diff
@@ -0,0 +1,96 @@
...
+                groupOrder: nextStepTemplate.groupOrder,
+                context: context ? context : Prisma.JsonNull,
```
dhenkel92, line 84: I think we should call some transformation function here. Just to make sure that the context is always of the same format

common/achievement/index.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+async function isAchievementConditionMet(achievement: UserAchievementTemplate) {
+    const {
+        recordValue,
+        template: { condition, conditionDataAggregations, metrics },
+    } = achievement;
+    if (!condition) {
+        return;
+    }
```
dhenkel92, line 102: We should log an error in this case. Would be quite bad to have active achievements without condition.

common/achievement/index.ts
```diff
@@ -0,0 +1,123 @@
...
+async function checkUserAchievement<ID extends ActionID>(userAchievement: UserAchievementTemplate, event: ActionEvent<ID>) {
+    const evaluationResult = await isAchievementConditionMet(userAchievement);
+
+    if (evaluationResult.conditionIsMet) {
+        const conditionDataAggregations = userAchievement?.template.conditionDataAggregations as ConditionDataAggregations;
+        const dataAggregationKey = Object.keys(conditionDataAggregations)[0];
+        const evaluationResultValue =
+            typeof evaluationResult.resultObject[dataAggregationKey] === 'number' ? Number(evaluationResult.resultObject[dataAggregationKey]) : null;
+        const awardedAchievement = await rewardUser(evaluationResultValue, userAchievement, event);
+        const userAchievementContext: UserAchievementContext = {};
```
dhenkel92, line 85: Should it always be empty? 🤔

common/achievement/evaluate.ts
```diff
@@ -0,0 +1,114 @@
...
+import { getBucketContext } from './util';
+const logger = getLogger('Achievement');
+
+export async function evaluateAchievement(
+    condition: string,
+    dataAggregation: ConditionDataAggregations,
+    metrics: string[],
+    recordValue: number
+): Promise<EvaluationResult> {
+    const achievementEvents = await prisma.achievement_event.findMany({ where: { metric: { in: metrics } }, orderBy: { createdAt: 'desc' } });
```
dhenkel92, line 17: We need to add the user id

common/achievement/evaluate.ts
```diff
@@ -0,0 +1,114 @@
...
+
+        const bucketCreator = dataAggregationObject.createBuckets || 'default';
+        const bucketAggregator = dataAggregationObject.bucketAggregator || 'count';
+
+        const aggregator = dataAggregationObject.aggregator;
+
+        const eventsForMetric = eventsByMetric[metricName];
+        if (!eventsForMetric) {
+            continue;
+        }
```
dhenkel92, line 45: I think I would remove this. It doesn't hurt to evaluate an empty array, and maybe we have reasons to do it.

common/achievement/evaluate.ts
```diff
@@ -0,0 +1,114 @@
...
+    };
+}
+
+export function createBucketEvents(events: Achievement_event[], bucketConfig: BucketConfig): BucketEvents[] {
+    switch (bucketConfig.bucketKind) {
+        case 'default':
+            return createDefaultBuckets(events, bucketConfig);
+        case 'time':
+            return createTimeBuckets(events, bucketConfig);
+    }
```
dhenkel92, line 90: please throw an error if none was found.

common/achievement/template.ts
```diff
@@ -0,0 +1,84 @@
+import 'reflect-metadata';
```
dhenkel92, line 1: Please add a comment about why we need this line

common/achievement/get.ts
```diff
@@ -0,0 +1,135 @@
...
+import { User } from '../user';
+import { ConditionDataAggregations } from './types';
+import { getAchievementState, getCurrentAchievementTemplateWithContext, transformPrismaJson } from './util';
+import { evaluateAchievement } from './evaluate';
+
+// TODO: getAchievementById -> passed user and achievementId to return a single achievement
+// TODO: resolver for nextSteps -> get active sequential achievements and important information
+const getUserAchievements = async (user: User): Promise<Achievement[]> => {
+    const userAchievements = await prisma.user_achievement.findMany({
+        where: { userId: user.userID },
```
dhenkel92, line 13: Should we add `isActive: true`?

common/achievement/get.ts
```diff
@@ -0,0 +1,135 @@
...
+                        const achievement: Achievement = await assembleAchievementData([groupAchievement], user);
+                        return achievement;
+                    })
+                );
+            }
+            const groupAchievement: Achievement = await assembleAchievementData(sortedGroupAchievements, user);
+            return [groupAchievement];
+        })
+    );
+    return Promise.all(achievements.reduce((a, b) => a.concat(b), []));
```
dhenkel92, line 45:
```suggestion
    return achievements.reduce((a, b) => a.concat(b), []);
```

common/achievement/get.ts
```diff
@@ -0,0 +1,135 @@
...
+    return achievements;
+};
+
+const generateReorderedAchievementData = async (groups: { [group: string]: User_achievement[] }, user: User): Promise<Achievement[]> => {
+    const groupKeys = Object.keys(groups);
+    const achievements = await Promise.all(
+        groupKeys.map(async (key) => {
+            const group = groups[key];
+            const sortedGroupAchievements = group.sort((a, b) => a.groupOrder - b.groupOrder);
+            if (sortedGroupAchievements[0].template.type === achievement_type_enum.TIERED) {
```
dhenkel92, line 33: Why do we do this for `TIERED` achievements only?

graphql/achievement/mutations.ts
```diff
@@ -0,0 +1,45 @@
...
+            await Notification.actionTaken(user, 'pupil_joined_match_meeting', {
+                matchId: matchId.toString(),
+                relation: `match/${matchId}`,
+            });
+        }
+
+        return true;
+    }
+
+    // ! - Just for testing
```
dhenkel92, line 34: Remove

seed-db.ts
```diff
@@ -738,6 +740,558 @@ void (async function setupDevDB() {
         },
     });
 
+    /* Achievements */
+    // STUDENT ONBOARDING
+    await prisma.achievement_template.create({
```
dhenkel92, line 745: ❤️ Thank you for all the test data

graphql/user/mutations.ts
```diff
@@ -72,4 +73,13 @@ export class MutateUserResolver {
...
     }
+    @Mutation(() => Boolean)
+    @Authorized(Role.USER)
+    async markAchievementAsSeen(@Arg('achievementId') achievementId: number) {
+        await prisma.user_achievement.update({
+            where: { id: achievementId },
+            data: { isSeen: true },
+        });
+        return true;
+    }
```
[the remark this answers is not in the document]
LucasFalkowsky: 😅 Yes

common/achievement/create.ts
```diff
@@ -0,0 +1,96 @@
...
+    });
+    return userAchievement;
+}
+
+async function getOrCreateUserAchievement<ID extends ActionID>(template: Achievement_template, userId: string, context?: SpecificNotificationContext<ID>) {
+    const isGlobal =
+        template.templateFor === achievement_template_for_enum.Global ||
+        template.templateFor === achievement_template_for_enum.Global_Courses ||
+        template.templateFor === achievement_template_for_enum.Global_Matches;
+    const existingUserAchievement = await findUserAchievement(template.id, userId, !isGlobal && context);
```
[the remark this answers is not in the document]
LucasFalkowsky: Was refactored to make this clearer.

common/achievement/create.ts
```diff
@@ -0,0 +1,96 @@
...
+        select: { id: true, userId: true, context: true, template: true, achievedAt: true, recordValue: true },
+    });
+    return userAchievement;
+}
+
+async function getOrCreateUserAchievement<ID extends ActionID>(template: Achievement_template, userId: string, context?: SpecificNotificationContext<ID>) {
+    const isGlobal =
+        template.templateFor === achievement_template_for_enum.Global ||
+        template.templateFor === achievement_template_for_enum.Global_Courses ||
+        template.templateFor === achievement_template_for_enum.Global_Matches;
```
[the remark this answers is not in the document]
LucasFalkowsky: We should be able to distinguish between global matches or global courses as well as all templates (global) in order to be able to specificly get the lectures for all matches, all subcourses, a single match or a single subcourse. Using null instead a enum value for global itself does not give us any benefit but makes it harder to understand what is happening in the code.

Correct me if I am completely off here.

common/achievement/create.ts
LucasFalkowsky: was done like that anyway. The if statement was removed.

common/achievement/create.ts
LucasFalkowsky: currently the match id is saved in the context, so it does not need to be included in the group name. Currently the group is just there to distinguish all achievements that belong to one tiered list / sequence.

common/achievement/create.ts
LucasFalkowsky: context has the type SpecificNotificationContext and therefore needs no transformation (right format). Prisma.JsonNull is used to pass an empty object to prisma db because prisma can't handle {}. The prisma documentation recomends the usage of Prisma.JsonNull.

common/achievement/create.ts
dhenkel92: In this case we have to make sure that it's taking into account everywhere.
https://github.com/corona-school/backend/blob/c42ff66a901b889b5a443b659dcaf4f6653b1ad5/common/achievement/get.ts#L16-L22

If I understand the code correctly, it would stack user_achievements of the same group, but different matches together.

common/achievement/create.ts
dhenkel92: Can you guarantee that there is no way to pass other data than `SpecificNotificationContext`?

common/achievement/evaluate.ts
LucasFalkowsky: The case could not be nothing, following the code before. But, if it was, we return the default bucket events now.

common/achievement/index.ts
LucasFalkowsky: Nope. should be event.context.

common/achievement/evaluate.ts
dhenkel92: My idea was that it would save us from typo's when adding new buckets in the future.
I just checked, and the typesystem would catch it. Good job 👍

common/achievement/index.ts
LucasFalkowsky: why could a userAchievement be undefined. Can not follow.

common/achievement/template.ts
```diff
@@ -0,0 +1,84 @@
...
+export enum TemplateSelectEnum {
+    BY_GROUP = 'group',
+    BY_METRIC = 'metrics',
+}
+
+// string == metricId, group
+const achievementTemplates: Map<TemplateSelectEnum, Map<string, Achievement_template[]>> = new Map();
+
+async function getAchievementTemplates(select: TemplateSelectEnum): Promise<Map<string, Achievement_template[]>> {
+    if (!achievementTemplates.has(select)) {
```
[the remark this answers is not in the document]
LucasFalkowsky: @LomyW will create a ticket for that.

common/notification/actions.ts
```diff
@@ -614,6 +614,87 @@ const _notificationActions = {
...
+        sampleContext: {
+            achievement: {
+                id: '0',
+                name: 'achievement',
+            },
+        },
+    },
+
+    /* MEETINGS */
+    student_joined_meeting: {
```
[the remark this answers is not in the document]
LucasFalkowsky: Used to be for testing. Was removed now.

common/notification/types.ts
```diff
@@ -34,6 +34,8 @@ export interface NotificationContextExtensions {
     campaign?: string;
     // For Campaigns, support sending custom Mailjet Notifications:
     overrideMailjetTemplateID?: string;
+    // For Achievements, the match or subcourse is needed as a relation to allocate events to a specific user achievement
+    relation?: string;
```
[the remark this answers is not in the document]
LucasFalkowsky: If we have no access to the relation, we would not be able to access it at all. Having a nullable relation field is the best solution we had. Do you have any other suggestion how we could make this work otherwise?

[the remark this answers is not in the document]
LucasFalkowsky: If we did not include the relation to this NotificationContext the Error Property 'relation' does not exist on type 'SpecificNotificationContext<ID>' pops up.

common/achievement/template.ts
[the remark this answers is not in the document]
LomyW: @dhenkel92 added the issue to the `Potentielle Tickets` doc 👍🏼

common/achievement/get.ts
Jonasdoubleyou: .flat() ;)

common/achievement/index.ts
```diff
@@ -0,0 +1,123 @@
...
+        const formula = metric.formula;
+        const value = formula(event.context);
+
+        await prisma.achievement_event.create({
+            data: {
+                metric: metric.metricName,
+                value: value,
+                action: event.actionId,
+                userId: event.user.userID,
+                relation: event.context.relation ?? '',
```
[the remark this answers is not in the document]
Jonasdoubleyou: Speaking if cycles ... We should ensure that no achievement is attached to actions triggered by the achievement system :)

common/achievement/metrics.ts
```diff
@@ -0,0 +1,35 @@
+import 'reflect-metadata';
```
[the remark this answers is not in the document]
Jonasdoubleyou: And more importantly, why is this needed in this file? We have that at multiple places already, this suggests the startup order was changed?

common/achievement/create.ts
LucasFalkowsky: you are right. This function was updated to meet the requirements:

    const userAchievementGroups: { [groupRelation: string]: User_achievement[] } = {};
    userAchievements.forEach((ua) => {
        const relation = ua.context['relation'] || null;
        const key = relation ? `${ua.template.group}/${relation}` : ua.template.group;
        if (!userAchievementGroups[key]) {
            userAchievementGroups[key] = [];
        }
        userAchievementGroups[key].push(ua);
    });

common/achievement/get.ts
LucasFalkowsky: Tiered achievements are the only achievements that are not put together as a group to one achievement for the frontend.
Example:
Sequential achievements use every userAchievement as a step and therefore get processed together to one achievement seen in the frontend.
Streaks just exist as single achievement for a group and do not need this single processing either.
Tiered are steps on a road to higher scores. First you reach 1 then 3 then 5 then 10 ... these achievements all get displayed as a milestone so they are processt one by one.

common/achievement/get.ts
LucasFalkowsky: Crazy. Did not know .flat is a thing!

common/achievement/index.ts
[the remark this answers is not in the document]
LucasFalkowsky: what actions are you talking about? Could you describe this further, so we can adjust.

graphql/user/mutations.ts
```diff
@@ -72,4 +73,14 @@ export class MutateUserResolver {
...
         return true;
     }
+    @Mutation(() => Boolean)
+    @AuthorizedDeferred(Role.ADMIN, Role.OWNER)
+    async markAchievementAsSeen(@Ctx() context: GraphQLContext, @Arg('achievementId') achievementId: number) {
+        const acheivement = await prisma.user_achievement.update({
+            where: { id: achievementId },
+            data: { isSeen: true },
+        });
+        await hasAccess(context, 'User_achievement', acheivement);
```
Jonasdoubleyou, line 83: the authorization check should be done before the update?

common/achievement/evaluate.ts
```diff
@@ -0,0 +1,126 @@
...
+
+async function _evaluateAchievement(
+    userId: string,
+    condition: string,
+    dataAggregation: ConditionDataAggregations,
+    recordValue: number,
+    relation?: string
+): Promise<EvaluationResult | undefined> {
+    // We only care about metrics that are used for the data aggregation
+    const metrics = Object.values(dataAggregation).map((entry) => entry.metric);
```
dhenkel92, line 23: Why is there achievementTemplate.metrics at all then?

common/achievement/get.ts
```diff
@@ -0,0 +1,214 @@
...
+    const state: achievement_state = getAchievementState(userAchievements, currentAchievementIndex);
+
+    const isNewAchievement = state === achievement_state.COMPLETED && !userAchievements[currentAchievementIndex].isSeen;
+
+    const condition = currentAchievementTemplate.condition.includes('recordValue')
+        ? currentAchievementTemplate.condition.replace('recordValue', (userAchievements[currentAchievementIndex].recordValue + 1).toString())
+        : currentAchievementTemplate.condition;
+
+    let maxValue;
+    let currentValue;
```
dhenkel92, line 144: Please add types.

common/achievement/get.ts
dhenkel92: Ahhhh, now I get it.
Could you please add a comment with these details?

common/achievement/util.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+    const transformedJson: AchievementContextType = {
+        user: user,
+        match: json['match'] ? json['match'] : undefined,
+        subcourse: json['subcourse'] ? json['subcourse'] : undefined,
+    };
+    return transformedJson;
+}
+
+export function getCurrentAchievementTemplateWithContext(userAchievement: User_achievement, achievementContext: AchievementContextType): Achievement_template {
```
dhenkel92, line 108:
```suggestion
export function renderAchievementWithContext(userAchievement: User_achievement, achievementContext: AchievementContextType): Achievement_template {
```

common/achievement/get.ts
```diff
@@ -0,0 +1,214 @@
...
+        if (!userAchievementGroups[ua.template.group]) {
+            userAchievementGroups[ua.template.group] = [];
+        }
+        userAchievementGroups[ua.template.group].push(ua);
+    });
+    const achievements: Achievement[] = await generateReorderedAchievementData(userAchievementGroups, user);
+    return achievements;
+};
+
+const generateReorderedAchievementData = async (groups: { [group: string]: User_achievement[] }, user: User): Promise<Achievement[]> => {
```
dhenkel92, line 102: I think the functions of this file should use the prisma instead of the graphql type as all the data is coming from the DB.

common/achievement/util.ts
dhenkel92, line 108: The `userAchievement` argument should use the prisma type I guess

common/achievement/util.ts
```diff
@@ -0,0 +1,142 @@
...
+        const updatedElement =
+            currentAchievementContext[key] && typeof currentAchievementContext[key] === 'string'
+                ? renderTemplate(currentAchievementContext[key], achievementContext)
+                : currentAchievementContext[key];
+        currentAchievementContext[key] = updatedElement;
+    });
+    return currentAchievementContext;
+}
+
+export function getAchievementState(userAchievements: User_achievement[], currentAchievementIndex: number) {
```
dhenkel92, line 121: Should use the prisma type

common/achievement/get.ts
```diff
@@ -0,0 +1,214 @@
...
+
+const assembleAchievementData = async (userAchievements: User_achievement[], user: User): Promise<Achievement> => {
+    let currentAchievementIndex = userAchievements.findIndex((ua) => !ua.achievedAt);
+    currentAchievementIndex = currentAchievementIndex >= 0 ? currentAchievementIndex : userAchievements.length - 1;
+
+    const achievementContext = transformPrismaJson(user, userAchievements[currentAchievementIndex].context);
+    const currentAchievementTemplate = getCurrentAchievementTemplateWithContext(userAchievements[currentAchievementIndex], achievementContext);
+
+    const achievementTemplates = await prisma.achievement_template.findMany({
+        where: { group: currentAchievementTemplate.group },
```
dhenkel92, line 131:
```suggestion
        where: { group: currentAchievementTemplate.group, isActive: true },
```

common/achievement/get.ts
```diff
@@ -0,0 +1,214 @@
...
+        orderBy: { groupOrder: 'asc' },
+    });
+
+    const state: achievement_state = getAchievementState(userAchievements, currentAchievementIndex);
+
+    const isNewAchievement = state === achievement_state.COMPLETED && !userAchievements[currentAchievementIndex].isSeen;
+
+    const condition = currentAchievementTemplate.condition.includes('recordValue')
+        ? currentAchievementTemplate.condition.replace('recordValue', (userAchievements[currentAchievementIndex].recordValue + 1).toString())
+        : currentAchievementTemplate.condition;
```
dhenkel92, line 141: This is not needed

common/achievement/get.ts
```diff
@@ -0,0 +1,214 @@
...
+        const dataAggregationKeys = Object.keys(currentAchievementTemplate.conditionDataAggregations);
+        const relation = userAchievements[currentAchievementIndex].context['relation'] || null;
+        const evaluationResult = await evaluateAchievement(
+            user.userID,
+            condition,
+            currentAchievementTemplate.conditionDataAggregations as ConditionDataAggregations,
+            userAchievements[currentAchievementIndex].recordValue,
+            relation
+        );
+        currentValue = dataAggregationKeys.map((key) => evaluationResult.resultObject[key]).reduce((a, b) => a + b, 0);
```
dhenkel92, line 155: I think this doesn't work as the resultObject is including the recordValue as well.

common/achievement/get.ts
```diff
@@ -0,0 +1,214 @@
...
+                : dataAggregationKeys
+                      .map((key) => {
+                          return Number(currentAchievementTemplate.conditionDataAggregations[key].valueToAchieve);
+                      })
+                      .reduce((a, b) => a + b, 0);
+    } else {
+        const achievementTemplates = await prisma.achievement_template.findMany({
+            where: { group: currentAchievementTemplate.group, isActive: true },
+            orderBy: { groupOrder: 'asc' },
+        });
```
dhenkel92, line 170: We are fetching them in the beginning of the function already

common/achievement/index.ts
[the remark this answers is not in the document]
dhenkel92: He is talking about this one
https://github.com/corona-school/backend/blob/03ce2c39b1c338e356069b98592bc69a3442fdb7/common/achievement/index.ts#L177-L179

common/achievement/types.ts
```diff
@@ -1,64 +1,119 @@
...
-        metricId: number;
+        metric: string;
         aggregator: string;
         // These two are used to first create all the needed buckets and then aggregate the events that fall into these
         // Default: count
         bucketAggregator?: string;
         // Default: one bucket / event
         createBuckets?: string;
+        // For tiered achievements we need the number (max value) that can be achieved (for the resolver)
+        valueToAchieve?: number;
```
dhenkel92, line 67: 🤔 What is this about?

common/achievement/metrics.ts
```diff
@@ -0,0 +1,34 @@
...
+// ↑ Needed by typegraphql: https://typegraphql.com/docs/installation.html
+import { ActionID } from '../notification/actions';
+import { Metric } from './types';
+
+export const metricByName: Map<string, Metric> = new Map();
+const metricsByAction: Map<ActionID, Metric[]> = new Map();
```
dhenkel92, line 7: Why is there a getter for only one of them?

common/achievement/template.ts
```diff
@@ -0,0 +1,65 @@
...
+        logger.info(`Loaded ${templatesFromDB.length} achievement templates into the cache`);
+    }
+    return achievementTemplates.get(select);
+}
+
+async function getTemplatesByMetrics(metricsForAction: Metric[]) {
+    const templatesByMetric = await getAchievementTemplates(TemplateSelectEnum.BY_METRIC);
+    const templateKeys = templatesByMetric ? Array.from(templatesByMetric.keys()) : [];
+    if (templateKeys.length === 0) {
+        logger.debug(`No achievement templates were found in the database for the metrics: ${metricsForAction.map((m) => `${m.metricName}, `)}`);
```
dhenkel92, line 52: 🤔 If there are templateKeys, it doesn't automatically mean that there are templates for the metrics passed.

graphql/achievement/mutations.ts
```diff
@@ -0,0 +1,33 @@
...
+import * as Notification from '../../common/notification';
+import { GraphQLContext } from '../context';
+import { AuthorizedDeferred, hasAccess } from '../authorizations';
+import { prisma } from '../../common/prisma';
+
+@Resolver(() => Achievement)
+export class MutateAchievementResolver {
+    @Mutation((returns) => Boolean)
+    @AuthorizedDeferred(Role.ADMIN, Role.OWNER)
+    async matchMeetingJoin(@Ctx() context: GraphQLContext, @Arg('matchId') matchId: number) {
```
dhenkel92, line 13: I think this should be in the match/mutations.ts file

graphql/achievement/mutations.ts
```diff
@@ -0,0 +1,33 @@
...
+        await hasAccess(context, 'Match', match);
+
+        if (user.studentId) {
+            await Notification.actionTaken(user, 'student_joined_match_meeting', {
+                matchId: matchId.toString(),
+                pupil: match.pupil,
+                relation: `match/${matchId}`,
+            });
+        } else if (user.pupilId) {
+            await Notification.actionTaken(user, 'pupil_joined_match_meeting', {
```
dhenkel92, line 25: Do we need the student in the context?

graphql/achievement/mutations.ts
```diff
@@ -0,0 +1,33 @@
...
+                matchId: matchId.toString(),
+                relation: `match/${matchId}`,
```
dhenkel92, line 27: Could you please create a helper function next to this function, where the split is happening

https://github.com/corona-school/backend/pull/866/files#diff-98b7fdaff6dad941d9abe0a45b8b1e2a43ff4bacc1acd99bd3c5a1479356a325R26-R33

And replace it everywhere

common/achievement/evaluate.ts
LucasFalkowsky: This is a very good question, which I am not able to answer. @LomyW do you know why this was done in the concept as we have it here?

common/achievement/create.ts
```diff
@@ -0,0 +1,111 @@
...
+    userId: string,
+    context?: SpecificNotificationContext<ID>
+): Promise<AchievementToCheck> {
+    const isGlobal =
+        template.templateFor === achievement_template_for_enum.Global ||
+        template.templateFor === achievement_template_for_enum.Global_Courses ||
+        template.templateFor === achievement_template_for_enum.Global_Matches;
+    const existingUserAchievement: AchievementToCheck = await findUserAchievement(template.id, userId, !isGlobal ? context : undefined);
+    if (!existingUserAchievement) {
+        return await createAchievement(template, userId, context);
```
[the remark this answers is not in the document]
LucasFalkowsky: I changed the context param, so it can not be undefined. This context being able to be undefined was unnecessary anyway.

common/achievement/get.ts
LucasFalkowsky: ok everything was retyped but it now was necessary including a new type in the file to secure right type of the achievement group including templates.

common/achievement/get.ts
LucasFalkowsky: I think it is, why should it not be?

common/achievement/get.ts
LucasFalkowsky: this is why data aggregation keys are use of which none is revordValue, therefore having record value being ignored here.

common/achievement/metrics.ts
LucasFalkowsky: metricByName is not necessary to get something. MetricByName is used to register all metrics found in the metric.ts. It is checked, if the metric was already registered or not.

common/achievement/template.ts
LucasFalkowsky: This statement should work?

`if (templateKeys.map((key) => templatesByMetric[key]).reduce((all, temp) => [...all, ...temp], 0).length === 0)`

graphql/achievement/mutations.ts
LucasFalkowsky: i dont think so. We don't even need the pupil there.

common/achievement/template.ts
LucasFalkowsky: We now use all.concat inside the reduce function.

`templateKeys.map((key) => templatesByMetric[key]).reduce((all, temp) => all.concat(temp), []).length === 0`

common/achievement/get.ts
dhenkel92: Because the condition is only used for `evaluateAchievement`, where we have some logic to be able to use the `recordValue` without replacing it.
Nevertheless, I just noticed that we are  doing 
```
userAchievements[currentAchievementIndex].recordValue + 1
```
here. Why is that?

common/achievement/evaluate.ts
dhenkel92: Let's try to remove it and see if something breaks. I think it's a leftover from one of our first iterations.

graphql/subcourse/mutations.ts
```diff
@@ -99,6 +100,11 @@ export class MutateSubcourseResolver {
         const student = await getSessionStudent(context, studentId);
         await prisma.subcourse_instructors_student.create({ data: { subcourseId: result.id, studentId: student.id } });
 
+        await Notification.actionTaken(userForStudent(student), 'instructor_course_created', {
+            courseName: course.name,
+            subcourseId: result.id.toString(),
```
dhenkel92, line 105: Why do we need this?

graphql/course/mutations.ts
```diff
@@ -106,6 +109,30 @@ export class MutateCourseResolver {
...
+            where: { context: { path: ['relation'], equals: `subcourse/${subcourse.id}` }, userId: user.userID },
+            include: { template: true },
+        });
+        const subcourseAchievements = await Promise.all(
+            usersSubcourseAchievements.map(async (usersSubcourseAchievement) => {
+                return await prisma.user_achievement.findMany({
+                    where: { context: { path: ['relation'], equals: `subcourse/${subcourse.id}` }, templateId: usersSubcourseAchievement.template.id },
+                });
+            })
+        );
```
dhenkel92, line 126: Isn't that the same as 
```
        const subcourses = await prisma.user_achievement.findMany({
            where: { context: { path: ['relation'], equals: `subcourse/${subcourse.id}` } },
            include: { template: true },
        });
```
?

graphql/course/mutations.ts
```diff
@@ -106,6 +109,30 @@ export class MutateCourseResolver {
...
+        subcourseAchievements.flat().forEach(async (achievement) => {
```
dhenkel92, line 127: This should be awaited

graphql/course/mutations.ts
```diff
@@ -193,6 +220,20 @@ export class MutateCourseResolver {
...
+
+        const subcourse = await prisma.subcourse.findFirst({
+            where: { courseId: courseId },
+            include: { subcourse_instructors_student: { select: { student: true } } },
+        });
+        if (subcourse && subcourse.subcourse_instructors_student) {
+            subcourse.subcourse_instructors_student.forEach(async (instructor) => {
+                await Notification.actionTaken(userForStudent(instructor.student), 'instructor_course_submitted', {
+                    courseName: course.name,
+                    subcourseId: subcourse.id.toString(),
```
dhenkel92, line 232: Why do we need this?

graphql/achievement/mutations.ts
```diff
@@ -0,0 +1,34 @@
...
+import * as Notification from '../../common/notification';
+import { GraphQLContext } from '../context';
+import { AuthorizedDeferred, hasAccess } from '../authorizations';
+import { prisma } from '../../common/prisma';
+
+@Resolver(() => Achievement)
+export class MutateAchievementResolver {
+    @Mutation((returns) => Boolean)
+    @AuthorizedDeferred(Role.ADMIN, Role.OWNER)
+    async subcourseMeetingJoin(@Ctx() context: GraphQLContext, @Arg('subcourseId') subcourseId: number) {
```
dhenkel92, line 13: as the `matchMeetingJoin` is inside the match controller, maybe this function should be inside the sucourse controller.

graphql/achievement/mutations.ts
```diff
@@ -0,0 +1,34 @@
...
+        const { user } = context;
+        const subcourse = await prisma.subcourse.findUnique({ where: { id: subcourseId }, include: { course: true, lecture: true } });
```
dhenkel92, line 15: What if there is no subcourse found?

common/notification/actions.ts
```diff
@@ -210,10 +210,25 @@ const _notificationActions = {
     instructor_course_created: {
         description: 'Instructor / Course created (not yet published)',
         sampleContext: {
-            course: {
-                name: 'Hallo Welt',
-                description: 'Ein Kurs',
-            },
+            courseName: 'Beispielkurs',
+            subcourseId: '1',
```
dhenkel92, line 214: Why do we need this? We should always use the relation field in such cases.

common/notification/actions.ts
```diff
@@ -210,10 +210,25 @@ const _notificationActions = {
...
+            relation: 'subcourse/1',
+        },
+    },
+    instructor_course_submitted: {
+        description: 'Instructor / Course submitted for review',
+        sampleContext: {
+            courseName: 'Beispielkurs',
+            subcourseId: '1',
```
dhenkel92, line 222: Why do we need this? We should always use the relation field in such cases.

common/notification/actions.ts
```diff
@@ -210,10 +210,25 @@ const _notificationActions = {
...
+            relation: 'subcourse/1',
+        },
+    },
+    instructor_course_approved: {
+        description: 'Instructor / Course approved',
+        sampleContext: {
+            courseName: 'Beispielkurs',
+            subcourseId: '1',
```
dhenkel92, line 230: Why do we need this? We should always use the relation field in such cases.

common/notification/actions.ts
```diff
@@ -618,6 +633,74 @@ const _notificationActions = {
...
+    pupil_calendly_appointment_booked: {
+        description: 'Pupil booked appointment via calendly ',
+        sampleContext: {},
+    },
+
+    /* MEETINGS */
+    student_joined_match_meeting: {
+        description: 'Student joined a match meeting',
+        sampleContext: {
+            matchId: '1',
```
dhenkel92, line 680: Why do we need this? We should always use the relation field in such cases.

common/notification/actions.ts
```diff
@@ -618,6 +633,74 @@ const _notificationActions = {
...
+        description: 'Student joined subcourse meeting',
+        sampleContext: {
+            relation: 'subcourse/1',
+            subcourseLecturesCount: '5',
+        },
+    },
+    pupil_joined_match_meeting: {
+        description: 'Pupil joined a match meeting',
+        sampleContext: {
+            matchId: '1',
```
dhenkel92, line 694: Why do we need this? We should always use the relation field in such cases.

common/achievement/create.ts
```diff
@@ -0,0 +1,112 @@
...
+    templateId: number,
+    userId: string,
+    context: SpecificNotificationContext<ID>
+): Promise<AchievementToCheck> {
+    const keys = context ? Object.keys(context) : [];
+    const userAchievement = await prisma.user_achievement.findFirst({
+        where: {
+            templateId,
+            userId,
+            AND: keys.map((key) => {
```
dhenkel92, line 19: Do we actually need to search for the whole context here? Shouldn't it be enough to search by templateID + relation?

common/achievement/create.ts
```diff
@@ -0,0 +1,112 @@
...
+    const templatesForGroup = templatesByGroup.get(currentTemplate.group).sort((a, b) => a.groupOrder - b.groupOrder);
+
+    const keys = Object.keys(context);
+    const userAchievementsByGroup = await prisma.user_achievement.findMany({
+        where: {
+            template: {
+                group: currentTemplate.group,
+            },
+            userId,
+            AND: keys.map((key) => {
```
dhenkel92, line 61: Do we actually need to search for the whole context here? Shouldn't it be enough to search by templateID + relation?

common/achievement/create.ts
```diff
@@ -0,0 +1,112 @@
...
+    templatesForGroup: Achievement_template[],
+    nextStepIndex: number,
+    userId: string,
+    context: SpecificNotificationContext<ID>
+) {
+    if (templatesForGroup.length <= nextStepIndex) {
+        return;
+    }
+    const nextStepTemplate = templatesForGroup[nextStepIndex];
+    const achievedAt =
```
dhenkel92, line 91: Does this mean that the last SEQUENTIAL achievement will be always automatically achieved?

common/courses/states.ts
```diff
@@ -70,12 +71,24 @@ export async function allowCourse(course: Course, screeningComment: string | nul
...
+        include: { subcourse_instructors_student: { select: { student: true } } },
+    });
     for (const subcourse of subcourses) {
         if (await canPublish(subcourse)) {
             await publishSubcourse(subcourse);
         }
     }
+
+    const [subcourse] = subcourses;
+    subcourse.subcourse_instructors_student.forEach(async (instructor) => {
```
dhenkel92, line 85: I think it doesn't hurt to iterate over `subcourses`. This would at least ensure that it's not breaking when we want to support multiple subcourses for a course.

common/achievement/util.ts
```diff
@@ -0,0 +1,153 @@
+import 'reflect-metadata';
+// ↑ Needed by typegraphql: https://typegraphql.com/docs/installation.html
+import { AchievementContextType, RelationTypes } from './types';
+import { join } from 'path';
+import { prisma } from '../prisma';
+import { Prisma, achievement_template, user_achievement } from '@prisma/client';
+import { accessURLForKey } from '../file-bucket';
+import { achievement_state } from '../../graphql/types/achievement';
+import { User, getUserTypeAndIdForUserId } from '../user';
+import { Achievement_template, User_achievement } from '../../graphql/generated';
```
dhenkel92, line 10: Graphql types should not be used here.

common/achievement/bucket.ts
```diff
@@ -1,14 +1,23 @@
...
         return [];
     }
+    data.lecture.sort((a, b) => a.start.getTime() - b.start.getTime());
+    const filteredLectures: ContextLecture[] = data.lecture.filter((lecture, index, array) => {
+        if (index === 0) {
+            return true;
+        }
+        const previousEndTime = new Date(array[index - 1].start.getTime() + array[index - 1].duration * 60000);
+        return lecture.start >= previousEndTime;
+    });
```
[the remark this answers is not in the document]
LucasFalkowsky: This function is taking out lectures that overlap in a unwanted way. Having 2 lectures with the same relation would result in 2 buckets taking the same events, therefore rewarding the user with a value of 2 instead of 1.

I had a talk with AK and we came up with a different solution, perhibitting the creation of 2 lectures, overlapping in their timeframe.

Whats your opinion?

common/courses/states.ts
```diff
@@ -85,6 +85,7 @@ export async function allowCourse(course: Course, screeningComment: string | nul
     subcourse.subcourse_instructors_student.forEach(async (instructor) => {
         await Notification.actionTaken(userForStudent(instructor.student), 'instructor_course_approved', {
             courseName: course.name,
+            subcourseId: subcourse.id.toString(),
```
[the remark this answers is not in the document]
LucasFalkowsky: Technically this subcourseId is not for the evaluation but to replace a variable inside the actionRedirectLink of Cours submit, therefore making this single value without relation viable. But where you are right, the action is linked to a template that does not require the id inside another string, so it was removed for this action.

common/achievement/index.ts
```diff
@@ -171,8 +171,21 @@ async function rewardUser<ID extends ActionID>(evaluationResult: number, userAch
...
 
-    await actionTakenAt(new Date(event.at), event.user, 'user_achievement_reward_issued', {
-        achievement: { name: updatedAchievement.template.name, id: updatedAchievement.id.toString() },
-    });
+    const { type, group, groupOrder } = updatedAchievement.template;
+
+    if (type === achievement_type_enum.SEQUENTIAL) {
+        const templatesByGroup = await getAchievementTemplates(TemplateSelectEnum.BY_GROUP);
+        const groupTemplates = templatesByGroup.get(group);
+        const lastTemplate = groupTemplates[groupTemplates.length - 2];
```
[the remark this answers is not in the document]
LucasFalkowsky: So these are sequential achievements. The last template that actually needs evaluation is the second last in the group. The real last template needs no evaluation, since it is just there to display the information of a sequential achievement being rewarded. So we check, whether the last achievement to be evaluated is achieved, and if so, the next (reward achievement) is also generated for this sequential achievement.

I highlighted this inside the code now by mentioning it in a comment.

graphql/course/mutations.ts
LucasFalkowsky: you are right, the userId is unnecessary here, having the id of a user_achievement

graphql/course/mutations.ts
LucasFalkowsky: This was fixed by using a for ... of loop awaiting the prisma operations.

graphql/course/mutations.ts
LucasFalkowsky: This is the id, replacing a value inside the actionRedirectLink of the related template.

graphql/achievement/mutations.ts
LucasFalkowsky: The findUnique was replaced with findUniqueOrThrow.

common/achievement/bucket.ts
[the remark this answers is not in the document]
dhenkel92: 👍 I like the new idea.

common/courses/states.ts
```diff
@@ -257,6 +257,7 @@ export async function addSubcourseInstructor(user: User | null, subcourse: Subco
     const { name } = await prisma.course.findUnique({ where: { id: subcourse.courseId }, select: { name: true } });
     await Notification.actionTaken(userForStudent(newInstructor), 'instructor_course_created', {
         courseName: name,
+        subcourseId: subcourse.id.toString(),
```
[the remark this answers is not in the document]
dhenkel92: Can't we make it available in another way?

common/achievement/create.ts
LucasFalkowsky: yes, thats what happens.

common/achievement/index.ts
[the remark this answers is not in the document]
dhenkel92: 🤔 So it's a hack to show a last state?

common/courses/states.ts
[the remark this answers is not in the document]
LucasFalkowsky: Soooooo. How to fix this?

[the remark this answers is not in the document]
dhenkel92: What if we do something like this  (pdseudocode)
```
if (json['relation']) {
  const [relationType, relationId] = getRelationTypeAndId(json['relation']);
  transformedJson[`${relationType}ID`] = relationId;
}
```

here
https://github.com/corona-school/backend/pull/866/files#diff-98b7fdaff6dad941d9abe0a45b8b1e2a43ff4bacc1acd99bd3c5a1479356a325R96-R106

common/courses/states.ts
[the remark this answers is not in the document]
LucasFalkowsky: I don't think so. But we do this with other values too. like matchName or sth. We need to provide the information for templating in the context.

common/courses/states.ts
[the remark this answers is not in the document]
LucasFalkowsky: Yes we can do that there too. But it is the same as before but later. We add values to the context: AchievementContextType.

The code was adjusted!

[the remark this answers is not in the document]
dhenkel92: Ahh, I confused the two threads. Actually the answer ⬆️  is for the question ⬇️ 

For the type difference, we'll probably have to use a different type.

common/courses/states.ts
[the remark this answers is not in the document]
dhenkel92: I confused the threads. This belongs here.

https://github.com/corona-school/backend/pull/866#discussion_r1466253488

common/achievement/aggregator.ts
```diff
@@ -4,4 +4,36 @@ type Aggregator = Record<string, AggregatorFunction>;
...
-export const aggregators: Aggregator = {};
+export const aggregators: Aggregator = {
+    sum: {
+        function: (elements): number => {
+            return elements.reduce((total, num) => total + num, 0);
+        },
+    },
+    count: {
+        function: (elements): number => {
+            // TODO: evaluate if this assumption is correct
```
Jonasdoubleyou, line 15: Todo?

common/achievement/get.ts
```diff
@@ -0,0 +1,234 @@
...
+    });
+    Object.keys(userAchievementGroups).forEach((groupName) => {
+        const group = userAchievementGroups[groupName].sort((a, b) => a.groupOrder - b.groupOrder);
+        group[group.length - 1].achievedAt && delete userAchievementGroups[groupName];
+    });
+    const achievements: Achievement[] = await generateReorderedAchievementData(userAchievementGroups, user);
+    return achievements;
+};
+
+// Inactive achievements are acheievements that are not yet existing but could be achieved in the future.
```
Jonasdoubleyou, line 53: achievements

common/achievement/evaluate.ts
```diff
@@ -0,0 +1,134 @@
...
+async function _evaluateAchievement(
+    userId: string,
+    condition: string,
+    dataAggregation: ConditionDataAggregations,
+    recordValue?: number,
+    relation?: string
+): Promise<EvaluationResult | undefined> {
+    // We only care about metrics that are used for the data aggregation
+    const metrics = Object.values(dataAggregation).map((entry) => entry.metric);
+    // filter: wenn wir eine richtige relation haben -> filtern nach relation
```
Jonasdoubleyou, line 24: deutsch?

common/achievement/get.ts
```diff
@@ -0,0 +1,234 @@
...
+// They are created for every template in a Tiered achievements group that is not yet used as a achievement for a specific user.
+const getFurtherAchievements = async (user: User): Promise<Achievement[]> => {
+    const userAchievements = await prisma.user_achievement.findMany({
```
Jonasdoubleyou, line 56: getUserAchievementsWithTemplates

common/achievement/types.ts
```diff
@@ -1,64 +1,129 @@
...
-            start: Date;
-        }[];
-    };
-    appointment?: {
-        id: number;
-        duration?: number;
-        match?: number;
-        subcourse?: number;
-    };
+// Used to destinguish between different types of buckets
```
Jonasdoubleyou, line 24: distinguish

common/achievement/util.ts
```diff
@@ -0,0 +1,151 @@
...
+        templatesByGroups.get(template.group)!.push(template);
+    }
+    templatesByGroups.forEach((group, key) => {
+        group.sort((a, b) => a.groupOrder - b.groupOrder);
+        templatesByGroups.set(key, group);
+    });
+    return templatesByGroups;
+}
+
+export function isDefined<T>(arugment: T | undefined | null): arugment is T {
```
Jonasdoubleyou, line 149: arugment

common/courses/states.ts
```diff
@@ -70,12 +71,27 @@ export async function allowCourse(course: Course, screeningComment: string | nul
...
+        where: { courseId: course.id },
+        include: { subcourse_instructors_student: { select: { student: true, subcourseId: true } } },
+    });
     for (const subcourse of subcourses) {
         if (await canPublish(subcourse)) {
             await publishSubcourse(subcourse);
         }
     }
+
+    await Promise.all(
```
Jonasdoubleyou, line 84: This is at the wrong place, canPublish might be false. This should be inside of publishSubcourse

common/courses/states.ts
```diff
@@ -152,6 +168,16 @@ export async function cancelSubcourse(user: User, subcourse: Subcourse) {
     }
     await sendSubcourseCancelNotifications(course, subcourse);
     logger.info(`Subcourse (${subcourse.id}) was cancelled`);
+
+    const courseInstructors = await prisma.subcourse_instructors_student.findMany({ where: { subcourseId: subcourse.id }, select: { student: true } });
+    await prisma.user_achievement.deleteMany({
```
Jonasdoubleyou, line 173: Move to /common/achievement?

graphql/authorizations.ts
```diff
@@ -590,24 +591,26 @@ export const authorizationModelEnhanceMap: ModelsEnhanceMap = {
...
-                matchId: participantOrOwnerOrAdmin,
-                participantIds: adminOrOwner,
-                organizerIds: adminOrOwner,
-                declinedBy: participantOrOwnerOrAdmin,
-                zoomMeetingId: participantOrOwnerOrAdmin,
-                zoomMeetingReport: adminOrOwner,
-                override_meeting_link: participantOrOwnerOrAdmin,
-            }
-        ),
+        // TODO: This is a workaround for presentation purposes
```
Jonasdoubleyou, line 594: its okay to make these fields public

graphql/course/mutations.ts
```diff
@@ -106,6 +108,26 @@ export class MutateCourseResolver {
         }
         const result = await prisma.course.update({ data, where: { id: courseId } });
         logger.info(`Course (${result.id}) updated by Student (${context.user.studentId})`);
+
+        const subcourses = await prisma.subcourse.findMany({
+            where: { courseId: courseId },
+        });
+
+        for (const subcourse of subcourses) {
+            const usersSubcourseAchievements = await prisma.user_achievement.findMany({
```
Jonasdoubleyou, line 117: Move to common/achievement?

graphql/subcourse/mutations.ts
```diff
@@ -138,6 +141,14 @@ export class MutateSubcourseResolver {
             await removeParticipantFromCourseChat(instructorUser, subcourse.conversationId);
         }
         logger.info(`Student(${studentId}) was deleted from Subcourse(${subcourseId}) by User(${context.user.userID})`);
+
+        await prisma.user_achievement.deleteMany({
```
Jonasdoubleyou, line 145: common/achievement

graphql/subcourse/mutations.ts
```diff
@@ -371,4 +382,28 @@ export class MutateSubcourseResolver {
         logger.info(`Subcourse(${subcourseId}) was manually promoted by instructor(${context.user.userID})`);
         return true;
     }
+
+    @Mutation((returns) => Boolean)
+    @AuthorizedDeferred(Role.ADMIN, Role.OWNER)
+    async subcourseMeetingJoin(@Ctx() context: GraphQLContext, @Arg('subcourseId') subcourseId: number) {
```
Jonasdoubleyou, line 388: Why do we have different mutations for subcourse and match meetings? Why not just appointmentMeetingJoin(...)?

graphql/authorizations.ts
dhenkel92: Thanks for confirming

graphql/subcourse/mutations.ts
dhenkel92: Even though the use cases are quite similar at the moment, I recommend keeping them separate since they represent two distinct business scenarios.

Jonasdoubleyou (approved):

graphql/subcourse/mutations.ts
Jonasdoubleyou: hm, but in the UI they are all "appointments" so we now anyways need to branch off (currently in the frontend)

dhenkel92: Interesting. 
What's the expected setup?

Jonasdoubleyou: Well I'd prefer to have appointmentJoinMeeting(...) and then branch off in the backend. But this does not block this PR

common/achievement/types.ts
dhenkel92: https://github.com/corona-school/project-user/issues/1107

common/achievement/bucket.ts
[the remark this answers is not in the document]
dhenkel92: https://github.com/corona-school/project-user/issues/1108

common/achievement/index.ts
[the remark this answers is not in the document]
dhenkel92: https://github.com/corona-school/project-user/issues/1109

PR #869, opened 2023-10-21
cleanup: Remove Mentoring and Project Matching

Jonasdoubleyou:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1440


prisma/migrations/20231021213254_remove_project_matching_and_experts/migration.sql
```diff
@@ -0,0 +1,122 @@
+/* MANUAL */
+DELETE "log" WHERE 
+ "logtype" IN ('projectMatchDissolve', 'updateProjectFields', 'contactExpert');
+
+/* GENERATED */
+/*
+  Warnings:
```
Jonasdoubleyou, line 7: Hi @Maascha, could you clarify whether we (a) still need to keep this data somewhere (maybe as an export) for legal reasons and (b) whether there is any chance that this might be reactivated someday? If neither, I'd go ahead and drop all of this.

graphql/roles.ts
```diff
@@ -0,0 +1,85 @@
+import { pupil as Pupil, student as Student } from '@prisma/client';
```
Jonasdoubleyou, line 1: Wrong merge, I think this file moved to common/user/roles.ts in the meantime

JeangelLF: Oh, 🤦‍♂️ Got it, I'll remove this one then.

JeangelLF:
@Jonasdoubleyou Do you see something else I should consider here? :)

Jonasdoubleyou:
No, LGTM :)

JeangelLF (approved):

PR #879, opened 2023-10-28
HOTFIX: Migration: Only try to derive info about account deactivation if dissolvedAt date of match is set

realmayus:
Apparently some dissolved matches don't have a dissolvedAt date set


Jonasdoubleyou (approved):
Ah yes, we have reactivated matches

PR #883, opened 2023-11-01
feat: Declining the only lecture of a Subcourse leaves the course

Jonasdoubleyou:
resolves https://github.com/corona-school/project-user/issues/976

### DIFF
--- a/common/appointment/decline.ts
+++ b/common/appointment/decline.ts
@@ -6,6 +6,7 @@ import { getLogger } from '../logger/logger';
 import { getAppointmentForNotification } from './util';
 import { RedundantError } from '../util/error';
 import { getNotificationContextForSubcourse } from '../courses/notifications';
+import { leaveSubcourse } from '../courses/participants';
 
 const logger = getLogger('Appointment');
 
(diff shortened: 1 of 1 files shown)


common/appointment/decline.ts
```diff
@@ -52,4 +53,12 @@ export async function declineAppointment(user: User, appointment: Appointment) {
     }
 
     logger.info(`User(${user.userID}) declined Appointment(${appointment.id})`);
+
+    if (appointment.subcourseId) {
+        const isOnlyAppointment = (await prisma.lecture.count({ where: { subcourseId: appointment.subcourseId, isCanceled: false } })) === 1;
+        if (isOnlyAppointment) {
+            await leaveSubcourse(await prisma.subcourse.findUniqueOrThrow({ where: { id: appointment.subcourseId } }), pupil);
```
realmayus, line 60: What happens if `pupil` is null? Can't `declineAppointment` also be called with a student user?

realmayus (changes requested):

common/appointment/decline.ts
Jonasdoubleyou: Good point, I think for now students are always organizers and thus see the "cancel appointment" button instead of the "decline appointment". But in fact the GraphQL Mutation is also released to OWNER, so better to support this also here (it wasn't really supported before)

realmayus (approved):

PR #887, opened 2023-11-06
feat: achievement.actionTaken()

LomyW:
Closes https://github.com/corona-school/project-user/issues/881

## ✅ What was done
- [x] add `Achievement.actionTaken()


common/achievement/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+// import * as Achievement from '../../common/achievement';
+
+export type ActionEvent = {
+    action: string;
+    at: Date;
+    user: User;
+};
+
+export function actionTaken<ID extends ActionID>(user: User, actionId: ID, context: Context) {
```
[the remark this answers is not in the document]
LomyW: So if we add new actions for the achievements, should we extend the existing actions object? What do you think @Jonasdoubleyou? Then therefore it makes sense to use the `SpecificNotificationContext` or?
Should we than rename the context? Whats your suggestion?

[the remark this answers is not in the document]
Jonasdoubleyou: I'd suggest to use "actions" as a generic interface between code and business logic:

code -> actions -> notifications
                -> achievements

I would extend actions whenever something is generic enough, for specific things I'd directly call into the achievement logic (as we do in the Notification System)

common/achievement/index.ts
```diff
@@ -0,0 +1,25 @@
+import { ActionID, SpecificNotificationContext } from '../notification/actions';
+import { User } from '../user';
+import { Context } from './types';
+
+// import * as Achievement from '../../common/achievement';
```
Jonasdoubleyou, line 5: I think this is rather confusing

common/achievement/types.ts
```diff
@@ -7,7 +7,8 @@ export type Metric = {
 
 export type EventValue = number | string | boolean;
 
-export type FormulaContext = {
+// match, pool, openMatchRequestCount, screening, firstMatchRequest
+export type Context = {
```
Jonasdoubleyou, line 11: If you want to reuse the existing actions, why do we need a "Context" specific to achievements? Can't we use the notification context ("action context")?

common/achievement/types.ts
```diff
@@ -41,10 +27,10 @@ export interface BucketEventsWithAggr extends BucketEvents {
     aggregation: EventValue;
 }
 
-type BucketFormulaFunction = (context: FormulaContext) => Bucket[];
+type BucketFormulaFunction<ID extends ActionID> = (context: SpecificNotificationContext<ID>) => Bucket[];
```
LomyW, line 30: @Jonasdoubleyou do we also need the builder for buckets to work with the `SpecificNotificationContext`?

Jonasdoubleyou: Does the bucket formula function need the full context? Or would the "relationID" be enough?

common/achievement/helper.ts
```diff
@@ -0,0 +1,22 @@
...
+
+function isMetricExistingForActionId<ID extends ActionID>(actionId: ID): boolean {
+    return metrics.some((m) => m.onActions.includes(actionId));
+}
+
+async function isAchievementExistingForAction<ID extends ActionID>(actionId: ID): Promise<boolean> {
+    const metricsByAction = metrics.filter((m) => m.onActions.includes(actionId));
+
+    for (const metric of metricsByAction) {
+        const templates = await prisma.achievement_template.findMany({ where: { metrics: { has: metric.metricName } } });
```
Jonasdoubleyou, line 13: It might be worth caching this in a Map per metric (like the Notification System does)

common/achievement/helper.ts
```diff
@@ -0,0 +1,22 @@
+import { ActionID } from '../notification/actions';
+import { prisma } from '../prisma';
+import { metrics } from './metrics';
+
+function isMetricExistingForActionId<ID extends ActionID>(actionId: ID): boolean {
+    return metrics.some((m) => m.onActions.includes(actionId));
```
Jonasdoubleyou, line 6: might be worth storing in a map to avoid the m * n comparisons

common/achievement/metric.ts
```diff
@@ -1,4 +1,36 @@
...
+
+// This function utilizes generics to ensure flexibility in the ActionIDs and their respective contexts, allowing for dynamic metric creation.
+function createMetric<T extends ActionID[], K extends ContextForActions<T>>(
+    id: number,
+    metricName: string,
+    onActions: T,
+    formula: (context: K) => number
+): Metric {
+    return {
+        id,
```
Jonasdoubleyou, line 23: What is the id used for?

common/achievement/metrics.ts
```diff
@@ -0,0 +1,26 @@
+import { Metric } from './types';
+
+export const metrics: Metric[] = [];
```
Jonasdoubleyou, line 3: Could be a Map<string, Metric> as it is always accessed by metric name?

common/achievement/metrics.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function registerMetric(metric: Metric) {
+    const { metricName } = metric;
+    if (metricExists(metricName)) {
+        throw new Error(`Metric may only be registered once`);
+    }
+    metrics.push(metric);
+}
+
+export function registerMetrics(metricBatch: Metric[]) {
```
Jonasdoubleyou, line 15: Needed? Can't one just do metrics.forEach(registerMetric)

common/achievement/metrics.ts
```diff
@@ -0,0 +1,26 @@
...
+    metricBatch.forEach((metric) => {
+        const { metricName } = metric;
+        if (metricExists(metricName)) {
+            throw new Error(`Metric may only be registered once`);
+        }
+        if (metricName in metrics) {
```
Jonasdoubleyou, line 21: string key will never exist in an array

common/achievement/metrics.ts
```diff
@@ -0,0 +1,28 @@
+import { Metric } from './types';
+
+export const metrics: Map<string, Metric>[] = [];
```
Jonasdoubleyou, line 3: Why Map Array?

Jonasdoubleyou: I think it rather makes sense to have

metricByName: Map<string, Metric>
metricsByAction: Map<ActionID, Metric[]>

Jonasdoubleyou (approved):

PR #890, opened 2023-11-10
Pre-fill the admin auth token in dev environments to speed up testing

realmayus:
Pre-fill the admin auth token in dev environments to speed up testing

### DIFF
--- a/graphql/playground.ts
+++ b/graphql/playground.ts
@@ -2,7 +2,7 @@ import { ApolloServerPluginLandingPageGraphQLPlayground } from 'apollo-server-co
 import { isDev } from '../common/util/environment';
 
 const userHeader = { authorization: 'Bearer <somerandombearer>' };
-const adminHeader = { authorization: `Basic ${btoa('admin:' + process.env.ADMIN_AUTH_TOKEN)}` };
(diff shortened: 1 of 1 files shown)


graphql/playground.ts
```diff
@@ -2,17 +2,17 @@ import { ApolloServerPluginLandingPageGraphQLPlayground } from 'apollo-server-co
 import { isDev } from '../common/util/environment';
 
 const userHeader = { authorization: 'Bearer <somerandombearer>' };
-const adminHeader = { authorization: "Basic <btoa('admin' + password)>" };
+const adminHeader = { authorization: `Basic ${btoa('admin:' + process.env.ADMIN_AUTH_TOKEN)}` };
```
Jonasdoubleyou, line 5: Mmm can we at least exclude productive envs here? With this we might leak the admin token in the future

realmayus: The admin headers are only attached to dev environments, no?

Jonasdoubleyou: For now yes, but someone might accidentally change it in the future

Jonasdoubleyou (approved):

PR #893, opened 2023-11-15
feat: achievement standard aggregators

LomyW:
- Closes https://github.com/corona-school/project-user/issues/920

# ✅ What was done?
- [ ] add three standard aggregators

### DIFF
--- a/common/achievement/aggregator.ts
+++ b/common/achievement/aggregator.ts
@@ -1,7 +1,19 @@
+import { Achievement_event } from '../../graphql/generated';
 import { AggregatorFunction } from './types';
 
 type Aggregator = Record<string, AggregatorFunction>;
 
 // Aggregators are needed to aggregate event values (achievement_event.value) or buckets for evaluation (like sum, count, max, min, avg)
 
-export const aggregators: Aggregator = {};
+export const aggregators: Aggregator = {
+    sum: {
+        function: (values: number[]): number => {
+            return values.reduce((total, num) => total + num, 0);
+        },
+    },
+    count: {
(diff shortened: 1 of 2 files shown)


common/achievement/aggregator.ts
```diff
@@ -1,7 +1,19 @@
+import { Achievement_event } from '../../graphql/generated';
 import { AggregatorFunction } from './types';
 
 type Aggregator = Record<string, AggregatorFunction>;
 
 // Aggregators are needed to aggregate event values (achievement_event.value) or buckets for evaluation (like sum, count, max, min, avg)
 
-export const aggregators: Aggregator = {};
+export const aggregators: Aggregator = {
+    sum: {
```
Jonasdoubleyou, line 9: Why can't we do these in the DB? `SELECT SUM() FROM achievement_event WHERE userID = ? AND metricId = ?` ?

Jonasdoubleyou (approved):

common/achievement/aggregator.ts
LomyW: @Jonasdoubleyou Can you explain this in more detail? I don't quite understand it. 🙂

Jonasdoubleyou: Yeah maybe I don't understand what Aggregator is used for, but it seems a bit counterintuitive to do aggregations on the application side if the DB can do this for us - but I'm okay with that if it is hard to push down the bucketing logic to the DB :) Maybe clarify this with @dhenkel92 , he was more involved in the aggregation topic :)

PR #903, opened 2023-11-23
feat: achievements resolver

LucasFalkowsky:
What was done:

- getter function to select the necessary data and shape it into the right format
- achievement enums and types to return to the frontend
- field resolver to get user achievements


common/achievement/get.ts
```diff
@@ -0,0 +1,71 @@
+import { prisma } from '../prisma';
+import { Achievement_template, User_achievement, achievement_action_type_enum, achievement_type_enum } from '../../graphql/generated';
+import { Step, Achievement, achievement_state } from '../../graphql/types/achievement';
+
+const getUserAchievements = async (userId: string): Promise<Achievement[]> => {
```
Jonasdoubleyou, line 5: Why not (user: User)? Unlike a string, this can not be misused, and if we have a userId we have a user anyways ;)

common/achievement/get.ts
```diff
@@ -0,0 +1,71 @@
...
+    });
+    const userAchievementGroups: { [group: string]: User_achievement[] } = {};
+    userAchievements.forEach((ua) => {
+        if (!userAchievementGroups[ua.group]) {
+            userAchievementGroups[ua.group] = [];
+        }
+        userAchievementGroups[ua.group].push(ua);
+    });
+    const templates = await prisma.achievement_template.findMany({
+        where: { group: { in: userAchievements.map((ua) => ua.group) } },
```
Jonasdoubleyou, line 17: I don't think this is correct, according to https://github.com/corona-school/backend/blob/master/prisma/schema.prisma achievement_template.group is a template whereas user_achievement.group is instantiated. Why not use `user_achievement.templateId` or directly `{ include: { template: true } }` in the user_achievement query?

common/achievement/get.ts
```diff
@@ -0,0 +1,71 @@
...
+                return null;
+            }
+            const step: Step = {
+                description: achievement.stepName,
+                isActive: index === userAchievementGroups[group].length - 1,
+            };
+            return step;
+        });
+        const groupAchievement: Achievement = {
+            name: sortedGroupAchievements[0].name,
```
Jonasdoubleyou, line 50: I find this code hard to follow. Maybe it could be split up into multiple functions? And also it would benefit from some more variables:
```
const currentAchievementOfGroup = sortedGroupAchievements[0];
```

Jonasdoubleyou:
(also if you'd introduce `groupAchievements( achievements: UserAchievement[] ): GroupedAchievement[]` this could be unit tested ;)

common/achievement/get.ts
LucasFalkowsky: Good idea to include the template in the user_achievement already

common/achievement/get.ts
Jonasdoubleyou: https://softwareengineering.stackexchange.com/a/80092 - Prefer code quality over performance unless you can really prove a measurable difference (which I doubt exist in this case)

Jonasdoubleyou:
looks good from my side, but I guess @LomyW should als take a look :)

common/achievement/get.ts
```diff
@@ -0,0 +1,67 @@
...
+const createFrontendAchievements = (groups: { [group: string]: User_achievement[] }): Achievement[] => {
+    const achievements: Achievement[] = [];
+    for (const group in groups) {
+        const sortedGroupAchievements = groups[group].sort((a, b) => a.groupOrder - b.groupOrder);
+        const state: achievement_state =
+            groups[group].length === 0
+                ? achievement_state.INACTIVE
+                : groups[group].length === groups[group].length
+                ? achievement_state.COMPLETED
+                : achievement_state.ACTIVE;
```
dhenkel92, line 31: What is happening here? 😅 
Isn't `groups[group].length === groups[group].length` always true? Besides that, `groups[group].length === 0` can never happen right now.

common/achievement/get.ts
```diff
@@ -0,0 +1,67 @@
...
+        const newAchievement = groups[group].length === sortedGroupAchievements.length && !groups[group][groups[group].length - 1].isSeen;
```
dhenkel92, line 32: In what case is `groups[group].length === sortedGroupAchievements.length` not true?

common/achievement/get.ts
```diff
@@ -0,0 +1,67 @@
...
+
+        const steps: Step[] = sortedGroupAchievements.map((achievement, index) => {
+            if (!achievement.template.stepName) {
+                return null;
```
dhenkel92, line 36: Should these nulls be filtered out afterwards?

common/achievement/get.ts
```diff
@@ -0,0 +1,67 @@
...
+            }
+            const step: Step = {
+                description: achievement.template.stepName,
+                isActive: index === groups[group].length - 1,
```
dhenkel92, line 40: Can you please add a comment with a short explanation?

dhenkel92 (commented):
I have the feeling that I'm missing some information 😅

common/achievement/get.ts
LucasFalkowsky: the first thing you are right, is completely rubbish. Don't know what got into me there...
for the length === 0 we have this since the frontend accepts inactive achievements and this is only possible if there is no user achievement initiated to a template. Not possible yes but Its part of the design, so the resolver allows it.

LucasFalkowsky: the check for completed state checks if there is a achievement in the group that is not yet finished. else it is active state.

common/achievement/get.ts
LucasFalkowsky: // if a achievementTemplate has a stepName, it means that it must have multiple steps as well as being a sequential achievement
// for every achievement in the sortedGroupAchievements, we create a step object with the stepName (descirption) and isActive property for the achievement step currently active but unachieved

common/achievement/get.ts
```diff
@@ -0,0 +1,67 @@
...
+    });
+    const achievements: Achievement[] = createFrontendAchievements(userAchievementGroups);
+    return achievements;
+};
+
+const createFrontendAchievements = (groups: { [group: string]: User_achievement[] }): Achievement[] => {
+    const achievements: Achievement[] = [];
+    for (const group in groups) {
+        const sortedGroupAchievements = groups[group].sort((a, b) => a.groupOrder - b.groupOrder);
+        const currentAchievementIndex = sortedGroupAchievements.findIndex((ua) => !ua.achievedAt) - 1;
```
dhenkel92, line 26: Can't this cause an out of index error if the first achievement in the list (index: 0) isn't achieved yet?

common/achievement/get.ts
```diff
@@ -0,0 +1,67 @@
...
+        const state: achievement_state =
+            sortedGroupAchievements.length === 0
```
dhenkel92, line 28: I still don't get how `sortedGroupAchievements.length === 0` can happen 😅 

In lines 12 - 17, we only create a group if there is an achievement, and it's pushed to the array immediately. This means that if we have a group, it has at least one achievement in any case.

It would only make sense if we somehow try to compare templates with user_achievements.

common/achievement/get.ts
```diff
@@ -0,0 +1,67 @@
...
+                ? achievement_state.INACTIVE
+                : typeof currentAchievementIndex !== 'number'
```
dhenkel92, line 30: In what case is `currentAchievementIndex` not a number? 😅

common/achievement/get.ts
LucasFalkowsky: in this line we are searching for the first index to not have an achievedAt value. So this will not cause an error. But we would run in a different issue. If there was found no element the result would be -1. Also we have to return the index without substracting 1 so we get the real index value.

I added a resultIndex variable that points to null if index < 0 and replaced all currentAchievementIndexes with this variable.

common/achievement/get.ts
LucasFalkowsky: Ok this is a correct thought. To make this reasonable we'll have to make this userAchievementGroups map to be created from templates and not already existing achievements. I'll think about how to fix this!

common/achievement/get.ts
LucasFalkowsky: if no element was found at all.

dhenkel92: Got it, and it makes sense with the new implementation 🙂

graphql/types/achievement.ts
```diff
@@ -0,0 +1,71 @@
...
+    subtitle: string;
+
+    @Field()
+    description: string;
+
+    @Field()
+    image: string;
+
+    @Field()
+    alternativeText: string;
```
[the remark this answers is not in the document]
LucasFalkowsky: This is the alternative text for images.

common/achievement/get.ts
```diff
@@ -0,0 +1,99 @@
...
+        if (!userAchievementGroups[ua.template.group]) {
+            userAchievementGroups[ua.template.group] = [];
+        }
+        userAchievementGroups[ua.template.group].push(ua);
+    });
+    const achievements: Achievement[] = await createFrontendAchievements(userAchievementGroups, user);
+    return achievements;
+};
+
+const createFrontendAchievements = async (groups: { [group: string]: User_achievement[] }, user: User): Promise<Achievement[]> => {
```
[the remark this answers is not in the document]
LucasFalkowsky: I called it generateReorderedAchievementData. It does generate Achievements by reordering information from the database.

common/achievement/get.ts
```diff
@@ -0,0 +1,99 @@
...
+        achievements.push(groupAchievement);
+    }
+    return achievements;
+};
+
+const putTogetherAchievement = async (userAchievements: User_achievement[], user: User): Promise<Achievement> => {
+    let currentAchievementIndex = userAchievements.findIndex((ua) => !ua.achievedAt);
+    currentAchievementIndex = currentAchievementIndex >= 0 ? currentAchievementIndex : userAchievements.length - 1;
+    const contextKeys = Object.keys(userAchievements[currentAchievementIndex].context);
+    const achievementContext: Partial<Context> = {
```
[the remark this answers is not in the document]
LucasFalkowsky: If I understood you right, you want the context to contain any data, that is passed with the action. For example if the action taken has a matchId/1 and pupil/4 we want the context to contain match: <match for matchId 1>, pupil: <pupil for pupilId 4. From this context we get the template information like {{pupil.firstName}}.

If this is the case, that is exactly what is happening from line 38 until line 61.

common/achievement/get.ts
```diff
@@ -0,0 +1,99 @@
...
+        const sortedGroupAchievements = groups[group].sort((a, b) => a.groupOrder - b.groupOrder);
+        const groupAchievement: Achievement = await putTogetherAchievement(sortedGroupAchievements, user);
+        achievements.push(groupAchievement);
+    }
+    return achievements;
+};
+
+const putTogetherAchievement = async (userAchievements: User_achievement[], user: User): Promise<Achievement> => {
+    let currentAchievementIndex = userAchievements.findIndex((ua) => !ua.achievedAt);
+    currentAchievementIndex = currentAchievementIndex >= 0 ? currentAchievementIndex : userAchievements.length - 1;
```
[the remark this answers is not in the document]
LucasFalkowsky: It does. If we get the index of a userAchievement that has no value in isAchieved we take that index. If we have no index of an achievement since all of them are achieved, we take the index of the last achievementTemplate.

common/achievement/get.ts
```diff
@@ -0,0 +1,99 @@
...
+                  return {
+                      description: achievement.template.stepName,
+                      isActive: index === resultIndex,
+                  };
+              })
+            : null,
+        maxSteps: userAchievements.length,
+        currentStep: userAchievements.length,
+        newAchievement: newAchievement,
+        progressDescription: `Noch ${userAchievements.length - userAchievements.length} Schritte bis zum Abschluss`,
```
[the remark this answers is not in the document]
LucasFalkowsky: This was my first thought too but the problem here is, that we do not know the progressDescription in the frontend. It could be "Noch X Schritte" or "Noch X Meetings" or anything else. The unit is never certain. I completely forgot about this issue but I think we need an extra field for the progress bar description that could handle this. Maybe I am wrong, let me know your thoughts on this!

common/achievement/get.ts
```diff
@@ -0,0 +1,99 @@
...
+            ? userAchievements.map((achievement, index) => {
+                  // if a achievementTemplate has a stepName, it means that it must have multiple steps as well as being a sequential achievement
+                  // for every achievement in the sortedGroupAchievements, we create a step object with the stepName (descirption) and isActive property for the achievement step currently active but unachieved
+                  return {
+                      description: achievement.template.stepName,
+                      isActive: index === resultIndex,
+                  };
+              })
+            : null,
+        maxSteps: userAchievements.length,
```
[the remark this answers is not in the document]
LucasFalkowsky: You are right! And I see a different issue on this. The maxSteps of Tiered Achievements is also not the length but the condition of 15 or 20 or what ever events. We do not have this information here. We should have a talk about how to solve this!!!

common/achievement/get.ts
```diff
@@ -0,0 +1,99 @@
...
+const assembleAchievementData = async (userAchievements: User_achievement[], user: User): Promise<Achievement> => {
+    let currentAchievementIndex = userAchievements.findIndex((ua) => !ua.achievedAt);
+    currentAchievementIndex = currentAchievementIndex >= 0 ? currentAchievementIndex : userAchievements.length - 1;
+    const contextKeys = Object.keys(userAchievements[currentAchievementIndex].context);
+    const achievementContext: Partial<Context> = {
+        user: { ...user, fullName: `${user.firstname} ${user.lastname}` },
+    };
+    const newContextValue = await Promise.all(
+        contextKeys.map(async (key) => {
+            const [type, id] = userAchievements[currentAchievementIndex].context[key].split('/');
```
LomyW, line 43: I think this will not work, because we store not the relationId `match/1` in the user achievement context, we store other data like 
```
{
    matchId?: number
    subcourseId?: number
    matchPartner?: string
    ....
}
```

LomyW (commented):
some questions 👀

common/achievement/helper.ts
```diff
@@ -0,0 +1,88 @@
...
+        );
+        newContextValue.forEach((context) => {
+            achievementContext[context.key] = context.value;
+        });
+    }
+
+    return achievementContext;
+}
+
+export function getCurrentAchievementContext(userAchievement: User_achievement, achievementContext: Partial<Context>): Achievement_template {
```
dhenkel92, line 46: The name of the function is wrong. It's not returning a context, but a rendered template.

common/achievement/create.ts
```diff
@@ -0,0 +1,78 @@
+import { Prisma } from '@prisma/client';
+import { Achievement_template, JsonFilter } from '../../graphql/generated';
+import { ActionID, SpecificNotificationContext } from '../notification/actions';
+import { prisma } from '../prisma';
+import { TemplateSelectEnum, getAchievementTemplates } from './template';
+
+async function doesUserAchievementAlreadyExist<ID extends ActionID>(templateId: number, userId: string, context: SpecificNotificationContext<ID>) {
```
dhenkel92, line 7: Please update naming. `doesXXX` implies that it returns a boolean. You could use something like `findUserAchievementIfExists`.

common/achievement/create.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+async function createAchievement<ID extends ActionID>(templateToCreate: Achievement_template, userId: string, context: SpecificNotificationContext<ID>) {
+    const templatesByGroup = await getAchievementTemplates(TemplateSelectEnum.BY_GROUP);
+    const userAchievementsByGroup = await prisma.user_achievement.findMany({
+        where: { template: { group: templateToCreate.group } },
+        orderBy: { template: { groupOrder: 'asc' } },
+    });
+
+    const nextStepIndex = userAchievementsByGroup.length > 0 ? templateToCreate.groupOrder + 1 : 1;
```
dhenkel92, line 42: Can you please add a comment that this is used as a groupOrder and not array index? Besides that, please also say that it's not starting at 0 but 1.

dhenkel92, line 42: I don't feel comfortable in relying on the groupOrder field here. What if there is a gap?
We should find the array index of the last `userAchievements` within the `templatesByGroup` array, and then just pass the next one to `createNextUserAchievement`

common/achievement/create.ts
```diff
@@ -0,0 +1,78 @@
...
+    context: SpecificNotificationContext<ID>
+) {
+    const nextStepTemplate = templatesForGroup.find((template) => template.groupOrder === nextStepIndex);
+    // Here a user template is created for the next template in the group. This is done to always have the data availible for the next step.
+    // This could mean to, for example, have the name of a match partner that is not yet availible due to a unfinished matching process.
+    if (nextStepTemplate && nextStepTemplate.isActive) {
+        const createdUserAchievement = await prisma.user_achievement.create({
+            data: {
+                userId: userId,
+                group: nextStepTemplate.group,
```
dhenkel92, line 64: I think this should be rendered as well, right?

common/achievement/index.ts
```diff
@@ -1,50 +1,57 @@
...
 
-    const tracked = await trackEvent(event, context);
-    // TODO: check if user achievement already exists
-    // TODO: track event(event)
-    // TODO: checkAwardAchievement
+    for (const [key, group] of templatesByGroups) {
+        let achievementToCheck: AchievementToCheck;
+        for (const template of group) {
+            const userAchievement = await getOrCreateUserAchievement(template, user.userID, context);
+            if (userAchievement.achievedAt === null || userAchievement.recordValue) {
```
dhenkel92, line 41: Why are we using the recordValue to identify the type of an achievement if we have a achievement type field?
```suggestion
            if (userAchievement.achievedAt === null || userAchievement.type === 'STREAK') {
```

common/achievement/create.ts
dhenkel92: I just saw that `createNextUserAchievement` is recursive to get the same result.
I'm a little bit afraid what happens if the last achievement in the group is not active. This could crash the application 😅

common/achievement/evaluate.ts
```diff
@@ -0,0 +1,117 @@
...
+            eventsByMetric[event.metric] = [];
+        }
+        eventsByMetric[event.metric].push(event);
+    }
+
+    const resultObject: Record<string, number | string | boolean> = {};
+
+    for (const key in dataAggregation) {
+        if (!dataAggregation[key]) {
+            return;
```
dhenkel92, line 23: shouldn't this be a continue?

common/achievement/types.ts
```diff
@@ -47,3 +67,58 @@ export type ConditionDataAggregations = {
         createBuckets?: string;
```
dhenkel92, line 67:
```suggestion
        bucketCreator?: string;
```

common/achievement/evaluate.ts
```diff
@@ -0,0 +1,117 @@
...
+        const metricName = dataAggregationObject.metric;
+
+        const bucketCreator = dataAggregationObject.createBuckets || 'default';
+        const bucketAggregator = dataAggregationObject.bucketAggregator || 'count';
+
+        const aggregator = dataAggregationObject.aggregator;
+
+        const eventsForMetric = eventsByMetric[metricName];
+        if (!eventsForMetric) {
+            return;
```
dhenkel92, line 35: continue?

common/achievement/evaluate.ts
```diff
@@ -0,0 +1,117 @@
...
+        }
+        // we take the relation from the first event, that posesses one, in order to create buckets from it, if needed
+        const relation = eventsForMetric.find((event) => event.relation)?.relation;
+
+        const bucketCreatorFunction = bucketCreatorDefs[bucketCreator].function;
+        const bucketAggregatorFunction = aggregators[bucketAggregator].function;
+        const aggFunction = aggregators[aggregator].function;
+
+        if (!bucketCreatorFunction || !bucketAggregatorFunction || !aggFunction) {
+            return;
```
dhenkel92, line 45: I would either throw an error in this case or stay with the other return format where `conditionIsMet: false`. In any case I would suggest to log an error with all the information needed as this is a misconfiguration on our side.

common/achievement/evaluate.ts
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        const buckets = await bucketCreatorFunction(relation);
+        const bucketEvents = createBucketEvents(eventsForMetric, buckets);
+
+        const bucketAggr = bucketEvents.map(
+            (bucketEvent): BucketEventsWithAggr => ({
+                ...bucketEvent,
+                aggregation: bucketAggregatorFunction(bucketEvent.events.map((event) => event.value)),
+            })
```
dhenkel92, line 55: I would just return the aggregation value. We don't care about the rest later on. Then we can also remove the map on the next line.

common/achievement/evaluate.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+export function createBucketEvents(events: Achievement_event[], bucketConfig: BucketConfig): BucketEvents[] {
+    switch (bucketConfig.bucketKind) {
+        case 'default':
+            return createDefaultBuckets(events, bucketConfig);
+        case 'time':
+            return createTimeBuckets(events, bucketConfig);
+        case 'filter':
+            return createFilterBuckets(events, bucketConfig);
```
dhenkel92, line 80: I still don't get the idea of filter buckets. Could you please explain them to me again? 
I mean I get what they do, but I'm missing a good example to understand their need.

common/achievement/index.ts
```diff
@@ -1,50 +1,57 @@
...
-    user: User;
-};
-export type Achievement_Event = {
-    userId?: string;
-    metric: string;
-    value: EventValue;
-    action?: string;
-    relation?: string; // e.g. "user/10", "subcourse/15", "match/20"
-};
 export async function actionTaken<ID extends ActionID>(user: User, actionId: ID, context: SpecificNotificationContext<ID>) {
```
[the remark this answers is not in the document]
LucasFalkowsky: I'd go for rewardActionTaken, since the function is there to reward people whereas gamification would be the system itself. I'd also be fine with that. Tell me your thaughts.

common/achievement/helper.ts
```diff
@@ -0,0 +1,88 @@
+import { Prisma } from '@prisma/client';
+import { AchievementContextType } from './types';
+import { achievement_state } from '../../graphql/types/achievement';
+import { Context } from '../notification/types';
+import { User } from '../user';
+import { prisma } from '../prisma';
+import { Achievement_template, User_achievement } from '../../graphql/generated';
+import { renderTemplate } from '../../utils/helpers';
+
+export function transformPrismaJson(json: Prisma.JsonValue): AchievementContextType | null {
```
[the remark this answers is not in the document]
LucasFalkowsky: Yes. The content in the DB is a json. but Prisma has a own type, Prisma.JsonValue, which also can have a Prisma.JsonNull or other possible objects. Typescript does not accept this Prisma JsonValue as a valid object so we transform it beforehand.

common/achievement/helper.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export async function getAchievementContext(user: User, userAchievementContext: AchievementContextType): Promise<Partial<Context>> {
+    const achievementContext: Partial<Context> = {
+        user: { ...user, fullName: `${user.firstname} ${user.lastname}` },
+    };
+
+    if (userAchievementContext) {
+        const contextKeys = Object.keys(userAchievementContext);
+        const newContextValue = await Promise.all(
+            contextKeys.map(async (key) => {
```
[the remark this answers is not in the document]
LucasFalkowsky: I removed this whole function. I wanted to fit the AchievementContetType into Context. but now I extended the renderTemplate helper to also take achievementContextType.

common/achievement/get.ts
```diff
@@ -0,0 +1,78 @@
...
+
+const assembleAchievementData = async (userAchievements: User_achievement[], user: User): Promise<Achievement> => {
+    let currentAchievementIndex = userAchievements.findIndex((ua) => !ua.achievedAt);
+    currentAchievementIndex = currentAchievementIndex >= 0 ? currentAchievementIndex : userAchievements.length - 1;
+
+    const userAchievementContext = transformPrismaJson(userAchievements[currentAchievementIndex].context);
+    const achievementContext = await getAchievementContext(user, userAchievementContext);
+    const currentAchievementTemplate = getCurrentAchievementContext(userAchievements[currentAchievementIndex], achievementContext);
+
+    const resultIndex = currentAchievementIndex < 0 ? null : currentAchievementIndex;
```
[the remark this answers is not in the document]
LucasFalkowsky: I thought I already fixed this in an other branch, but yes thats absolutely right.

common/achievement/helper.ts
```diff
@@ -0,0 +1,88 @@
...
+        type: json['match'] ? 'match' : 'subcourse',
+        match: json['match'] ? json['match'] : undefined,
+        subcourse: json['subcourse'] ? json['subcourse'] : undefined,
+    };
+    return transformedJson;
+}
+
+export async function getAchievementContext(user: User, userAchievementContext: AchievementContextType): Promise<Partial<Context>> {
+    const achievementContext: Partial<Context> = {
+        user: { ...user, fullName: `${user.firstname} ${user.lastname}` },
```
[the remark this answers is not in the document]
LucasFalkowsky: also removed with the update of renderTemplate

common/achievement/get.ts
```diff
@@ -0,0 +1,78 @@
...
+    const state: achievement_state = getAchievementState(userAchievements, currentAchievementIndex);
+
+    const newAchievement = state === achievement_state.COMPLETED && !userAchievements[resultIndex].isSeen;
```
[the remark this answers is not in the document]
LucasFalkowsky: this should be isNewAchievement. I will also have to update the name in the frontend.

common/achievement/get.ts
```diff
@@ -0,0 +1,78 @@
...
+        alternativeText: 'alternativeText',
+        actionType: currentAchievementTemplate.actionType as achievement_action_type_enum,
+        achievementType: currentAchievementTemplate.type as achievement_type_enum,
+        achievementState: state,
+        steps: currentAchievementTemplate.stepName
+            ? userAchievements.map((achievement, index) => {
+                  // if a achievementTemplate has a stepName, it means that it must have multiple steps as well as being a sequential achievement
+                  // for every achievement in the sortedGroupAchievements, we create a step object with the stepName (descirption) and isActive property for the achievement step currently active but unachieved
+                  return {
+                      description: achievement.template.stepName,
```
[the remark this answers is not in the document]
LucasFalkowsky: the description is the step name.

common/achievement/create.ts
LucasFalkowsky: I think findUserAchievement would do the trick since a find function also could return undefined if nothing is found.

common/achievement/create.ts
LucasFalkowsky: This was already changed in a feature branch and also merged now.

LucasFalkowsky: we are using the indexes not the groups

LucasFalkowsky: as mentioned in the comment before, we had this issue during implementation and changed it in a feature branch that is now merged. But your concerns are completely justified.

common/achievement/types.ts
LucasFalkowsky: I think the previous made more sense. The createBuckets variable is pointing to the values after which the buckets will be created. The bucketCreator will then be selected using these values.

common/achievement/evaluate.ts
LucasFalkowsky: I logg an error message containing the values that caused the error and also where the error is originating from.

LucasFalkowsky: I let the function continue anyways, the evaluation will turn out false anyhow.

common/achievement/evaluate.ts
LucasFalkowsky: we removed the filter bucket.

common/achievement/index.ts
```diff
@@ -54,18 +61,62 @@ async function trackEvent(event: ActionEvent, context: NotificationContext) {
...
+async function isAchievementConditionMet(achievement: UserAchievementTemplate) {
+    const {
+        userId,
+        template: { condition, conditionDataAggregations, metrics },
+    } = achievement;
+    if (!condition) {
+        return;
+    }
+
+    const updatedCondition = injectRecordValue(condition, achievement.recordValue);
```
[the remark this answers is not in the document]
LucasFalkowsky: this is not possible. The dataAggregationValues belong to the template not the userAchievement.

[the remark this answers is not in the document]
dhenkel92: My explanation was probably very bad 😅 
Why not something like this? https://github.com/corona-school/backend/pull/929

common/achievement/evaluate.ts
```diff
@@ -48,22 +49,18 @@ export async function evaluateAchievement(
         const aggFunction = aggregators[aggregator].function;
 
         if (!bucketCreatorFunction || !bucketAggregatorFunction || !aggFunction) {
-            return;
+            logger.error(
+                `No bucket creator or aggregator function found for ${bucketCreator}, ${aggregator} or ${bucketAggregator} during the evaluation of achievement`
+            );
+            continue;
```
dhenkel92, line 55: I would do a return in this case. It doesn't make sense to evaluate if we cannot get all the data needed.

common/achievement/get.ts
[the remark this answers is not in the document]
dhenkel92: But why is it called name on the one side and description on the other 😂

common/achievement/create.ts
dhenkel92: I don't remember if it changed with recent iterations, but I thought that the group name might include a handlebars template, so that we can distinguish two groups of the same template.

common/achievement/evaluate.ts
dhenkel92: Why not returning? It would make sure that we don't run into weird side-effects. Furthermore, it doesn't make sense to spend more cpu time when we already know that we don't like the result.

common/achievement/index.ts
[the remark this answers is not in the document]
LucasFalkowsky: ah Ok. insert the record value in the resultObject for evaluation inside the evaluation function. Got it! Good idea!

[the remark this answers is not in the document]
LucasFalkowsky: I took the liberty of merging it directly.

common/achievement/evaluate.ts
LucasFalkowsky: good point.

common/achievement/index.ts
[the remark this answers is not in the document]
dhenkel92: !Attention! this might break the way we figure out the new record value

dhenkel92 (approved):

PR #911, opened 2023-11-30
feat: relation context

LucasFalkowsky:
closes: https://github.com/corona-school/project-user/issues/1030

What was changed:

- userAchievements are now fetched by context
- when adding a userAchievements the context is passed directly or set to null

### DIFF
--- a/common/achievement/create.ts
+++ b/common/achievement/create.ts
@@ -1,29 +1,37 @@
-import { Achievement_template } from '../../graphql/generated';
+import { Achievement_template, JsonFilter } from '../../graphql/generated';
+import { ActionID, SpecificNotificationContext } from '../notification/actions';
 import { prisma } from '../prisma';
 import { TemplateSelectEnum, getAchievementTemplates } from './template';
-import { UserAchievementContext } from './types';
 
-async function doesUserAchievementAlreadyExist(templateId: number, userId: string) {
-    // TODO - check if user achievement exist for one match or one subcourse
+async function doesUserAchievementAlreadyExist<ID extends ActionID>(templateId: number, userId: string, context: SpecificNotificationContext<ID>) {
+    const keys = Object.keys(context);
     const userAchievement = await prisma.user_achievement.findFirst({
         where: {
             templateId,
             userId,
+            AND: keys.map((key) => {
+                return {
+                    context: {
+                        path: key,
+                        equals: context[key],
+                    },
+                };
+            }),
         },
         select: { id: true, userId: true, context: true, template: true, achievedAt: true, recordValue: true },
     });
     return userAchievement;
 }
 
-async function getOrCreateUserAchievement(template: Achievement_template, userId: string, context?: UserAchievementContext) {
-    const existingUserAchievement = await doesUserAchievementAlreadyExist(template.id, userId);
+async function getOrCreateUserAchievement<ID extends ActionID>(template: Achievement_template, userId: string, context?: SpecificNotificationContext<ID>) {
+    const existingUserAchievement = await doesUserAchievementAlreadyExist(template.id, userId, context);
     if (!existingUserAchievement) {
         return await createAchievement(template, userId, context);
     }
     return existingUserAchievement;
 }
 
-async function createAchievement(templateToCreate: Achievement_template, userId: string, context: UserAchievementContext) {
+async function createAchievement<ID extends ActionID>(templateToCreate: Achievement_template, userId: string, context: SpecificNotificationContext<ID>) {
     const templatesByGroup = await getAchievementTemplates(TemplateSelectEnum.BY_GROUP);
     const userAchievementsByGroup = await prisma.user_achievement.findMany({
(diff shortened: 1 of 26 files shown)


common/achievement/create.ts
```diff
@@ -39,7 +47,12 @@ async function createAchievement(templateToCreate: Achievement_template, userId:
...
 }
 
-async function createNextUserAchievement(templatesForGroup: Achievement_template[], nextStepIndex: number, userId: string, context: UserAchievementContext) {
+async function createNextUserAchievement<ID extends ActionID>(
+    templatesForGroup: Achievement_template[],
+    nextStepIndex: number,
+    userId: string,
+    context: SpecificNotificationContext<ID>
+) {
     const nextStepTemplate = templatesForGroup.find((template) => template.groupOrder === nextStepIndex);
```
LomyW, line 57: For validation reasons: we could check here whether the correct context for the template exists using the `templateFor` field, as we know here if `templateFor === Match` that a `matchId` / or `relation: "match/1"` must exist, the same applies to subcourse

common/notification/actions.ts
```diff
@@ -638,15 +638,11 @@ const _notificationActions = {
     },
     joined_match_meeting: {
         description: 'User joined a match meeting',
-        sampleContext: {
```
LomyW, line 641: I think I would still collect further data in context such as the `matchId / subcourseId  /matchpartner` etc.

common/achievement/util.ts
```diff
@@ -26,10 +26,10 @@ export function getMetricsByAction<ID extends ActionID>(actionId: ID): Metric[]
     return metricsByAction.get(actionId) || [];
 }
 
-type RelationTypes = 'match' | 'subcourse';
+type RelationTypes = 'match' | 'subcourse' | 'achievementName';
```
LomyW, line 29: what is the `achievementName` for?

common/achievement/create.ts
```diff
@@ -1,29 +1,37 @@
...
 
-async function doesUserAchievementAlreadyExist(templateId: number, userId: string) {
-    // TODO - check if user achievement exist for one match or one subcourse
+async function doesUserAchievementAlreadyExist<ID extends ActionID>(templateId: number, userId: string, context: SpecificNotificationContext<ID>) {
+    const keys = Object.keys(context);
     const userAchievement = await prisma.user_achievement.findFirst({
         where: {
             templateId,
             userId,
+            AND: keys.map((key) => {
```
LomyW, line 12: So with this check we can guarantee that the user_achievement for a user exists for a specific match / subcourse?

LomyW (changes requested):
Nice ✨

common/notification/actions.ts
LucasFalkowsky: and we do in the mutations file. As I see it, this is just a sample and not responsible for fetching all necessary data, as this is done by the mutation. Am I wrong?

common/achievement/create.ts
LucasFalkowsky: but we do pass in the context from the previous userAchievement. Isn't that what we want here? like:
{ matchId: match/1 } from previous userAchievement gets also written in this achievements context.

common/notification/actions.ts
LucasFalkowsky: I talked to our beloved boss and he explained to me that the sample context is just for visualization and is never used. I adjusted my code accordingly.

Jonasdoubleyou: "and he explained to me that the sample context is just for visualization and is never used" - This is not correct. The sample context is used to validate Notifications created for a certain action to ensure all templating done is correct. Also it is used to infer the type of `SpecificNotificationContext`. So please maintain this.

common/achievement/util.ts
```diff
@@ -39,9 +39,9 @@ export function getRelationTypeAndId(relation: string): [relationType: RelationT
...
 }
 
-export async function getRelationContext(relation: string): Promise<RelationContextType> {
+export async function getAchievementContext(relation: string): Promise<AchievementContextType> {
     const [type, id] = getRelationTypeAndId(relation);
-    const relationContext: RelationContextType = {
+    const achievementContext: AchievementContextType = {
         match:
             type === 'match'
                 ? await prisma.match.findFirst({ where: { id }, select: { id: true, lecture: { select: { start: true, duration: true } } } })[0]
```
LomyW, line 47: Is this working `[0]`?

common/achievement/types.ts
```diff
@@ -111,7 +111,7 @@ export type EvaluationResult = {
     resultObject: Record<string, string | number | boolean>;
 };
 
-export type RelationContextType = {
+export type AchievementContextType = {
     match?: {
         id: number;
         lecture: {
```
LomyW, line 117: Here I think we need an array of lectures or? So that we can create a bucket for each match appointment or?

LomyW (changes requested):

PR #914, opened 2023-12-02
feat: add shared attribute, add markshared mutation, add templateCour…

paul-jaworski:


graphql/course/fields.ts
```diff
@@ -71,4 +71,41 @@ export class ExtendedFieldsCourseResolver {
...
+                        AND: [
+                            {
+                                course_instructors_student: {
+                                    some: {
+                                        studentId: studentId,
+                                    },
+                                },
+                            },
+                            {
+                                OR: [{ name: { contains: search } }, { description: { contains: search } }],
```
Jonasdoubleyou, line 96: Instead of rolling another version of search we should just use `courseSearch(search)` here (c.f. https://github.com/corona-school/backend/pull/870)

graphql/course/mutations.ts
```diff
@@ -82,6 +82,19 @@ export class MutateCourseResolver {
         return result;
     }
 
+    @Mutation((returns) => GraphQLModel.Course)
+    @Authorized(Role.ADMIN, Role.INSTRUCTOR)
```
Jonasdoubleyou, line 86: This should be `@AuthorizedDeferred` as we do not evaluate the roles before the method, but inside of it (with `hasAccess`). Roles should be `Role.ADMIN` and `Role.OWNER` - We want only the owner of a course be able to edit their own course, not every course instructor should be able to edit every course.

graphql/course/mutations.ts
```diff
@@ -82,6 +82,19 @@ export class MutateCourseResolver {
...
+    async courseMarkShared(@Ctx() context: GraphQLContext, @Arg('courseId') courseId: number, @Arg('shared') shared: boolean): Promise<GraphQLModel.Course> {
+        const course = await getCourse(courseId);
+        await hasAccess(context, 'Course', course);
+
+        const updatedCourse = await prisma.course.update({
+            where: { id: course.id },
+            data: { shared: shared },
+        });
```
Jonasdoubleyou, line 94: Missing logging, maybe add something like 

```
logger.info(`Course(${course.id} was ${shared ? "shared" : "unshared"} by User(${context.user.userID})`)
```

Jonasdoubleyou (commented):
Looks pretty good already! 
You also need to adapt `authorizations.ts` and mark the `shared` attribute as public, it's not something we need to keep private.

Jonasdoubleyou:
And this could use an integration test where one instructor publishes a course, and then another instructor is able to find it (c.f. https://github.com/corona-school/backend/blob/fef120319c0c099fc9c12bd15227821e2d95da8c/integration-tests/07_course.ts#L240)

paul-jaworski:
> And we should also relax the check in
> 
> https://github.com/corona-school/backend/blob/fef120319c0c099fc9c12bd15227821e2d95da8c/graphql/subcourse/mutations.ts#L82
> - The access check on the course should be checked if the course is shared.

If I make the hasAccess method call conditional, the authorization method must not be @AuthorizedDeferred then right?

graphql/course/fields.ts
Jonasdoubleyou: No it can just be combined `AND: [{ ... }, courseSearch(search)]`

Jonasdoubleyou:
No, just `if (!course.shared) await hasAccess(...);`

Jonasdoubleyou:
Ah, true, we need two checks then. `@Authorized(Role.INSTRUCTOR)` and then manually check inside subcourseAdd whether the course is shared or the user is an instructor

paul-jaworski:
Hey Jonas, this got a bit confusing to me after a long break. May you please sum up which methods and auth checks should be changed?

Thank you very much :)

Jonasdoubleyou:
- `subcourseCreate` should be available to all `Role.INSTRUCTOR`, then it should be checked whether the course is shared or owned by the instructor
- An integration test is missing

paul-jaworski:
I see that there is a method call `"await hasAccess(context, 'Course', course);"`. Should the check whether the course is shared or owned by the instructor be included there?

graphql/subcourse/mutations.ts
```diff
@@ -80,27 +80,39 @@ export class MutateSubcourseResolver {
...
         });
+        const isCourseSharedOrOwned = !!courseInstructorAssociation || course.shared;
+        if (!isCourseSharedOrOwned) {
+            logger.error(`Subcourse(${courseId}) is not shared or Student(${studentId}) is not an instructor of this subcourse`);
+            throw new PrerequisiteError('This subcourse is not shared or you are not the instructor of this subcourse!');
+        }
 
-        const student = await getSessionStudent(context, studentId);
-        await prisma.subcourse_instructors_student.create({ data: { subcourseId: result.id, studentId: student.id } });
+        if (isCourseSharedOrOwned) {
```
Jonasdoubleyou, line 95: unnecessary if

graphql/subcourse/mutations.ts
```diff
@@ -80,27 +80,39 @@ export class MutateSubcourseResolver {
...
-                groupChatType,
+        const courseInstructorAssociation = await prisma.course_instructors_student.findFirst({
+            where: {
+                courseId: courseId,
+                studentId: studentId,
             },
         });
+        const isCourseSharedOrOwned = !!courseInstructorAssociation || course.shared;
+        if (!isCourseSharedOrOwned) {
+            logger.error(`Subcourse(${courseId}) is not shared or Student(${studentId}) is not an instructor of this subcourse`);
```
Jonasdoubleyou, line 91: Subcourse -> Course

Jonasdoubleyou (commented):
looks good, some minor touchups to be made then this is ready :)

graphql/subcourse/mutations.ts
```diff
@@ -71,7 +71,7 @@ class PublicLectureInput {
 @Resolver((of) => GraphQLModel.Subcourse)
 export class MutateSubcourseResolver {
     @Mutation((returns) => GraphQLModel.Subcourse)
-    @AuthorizedDeferred(Role.ADMIN, Role.OWNER)
+    @AuthorizedDeferred(Role.INSTRUCTOR)
```
Jonasdoubleyou, line 74: Ah, and this must be `@Authorized` instead of `@AuthorizedDeferred`

graphql/subcourse/mutations.ts
paul-jaworski: Thanks, you are right since I already throw an error in line 92 when the if condition isnt met :)

graphql/subcourse/mutations.ts
```diff
@@ -80,6 +80,17 @@ export class MutateSubcourseResolver {
     ): Promise<GraphQLModel.Subcourse> {
         const course = await getCourse(courseId);
         await hasAccess(context, 'Course', course);
+        const courseInstructorAssociation = await prisma.course_instructors_student.findFirst({
+            where: {
+                courseId: courseId,
+                studentId: studentId,
```
Jonasdoubleyou, line 86: Ah, this for sure is wrong, we cannot trust our users to pass in their id. Please do

```
const student = await getSessionStudent(context, studentId);

... studentId: student.id
```

Jonasdoubleyou (commented):
sorry, overlooked this earlier

integration-tests/07_course.ts
```diff
@@ -21,6 +22,7 @@ const courseOne = test('Create Course One', async () => {
             description: "Why should I test if my users can do that for me in production?"
             category: club
             allowContact: true
+            shared: true
```
Jonasdoubleyou, line 25: I think you need a separate mutation markShared afterwards

integration-tests/07_course.ts
```diff
@@ -126,7 +129,7 @@ export const subcourseOne = test('Create Subcourse', async () => {
                 allowChatContactProspects: true
                 allowChatContactParticipants: true
                 groupChatType: ${ChatType.NORMAL}
-            }) { id }
+            } studentId: ${student.userID}) { id }
```
Jonasdoubleyou, line 132: and this can be reverted as we now use the session student

Jonasdoubleyou (commented):
apart from fixing the integration tests this should be good now

integration-tests/07_course.ts
paul-jaworski: Can you please rephrase this comment? Do you mean that we need a separate test for shared and nonshared courses? For the current integration test I manually set the shared value to false. See line 414

Jonasdoubleyou: No the courseCreate Mutation does not take a shared attribute, thus the test is failing?

paul-jaworski:
@Jonasdoubleyou or @realmayus Can you please have another look here?

Jonasdoubleyou:
Um, the tests are still failing, have you seen my comment regarding shared: true?

integration-tests/07_course.ts
```diff
@@ -126,7 +128,7 @@ export const subcourseOne = test('Create Subcourse', async () => {
                 allowChatContactProspects: true
                 allowChatContactParticipants: true
                 groupChatType: ${ChatType.NORMAL}
-            }) { id }
+            } { id }
```
Jonasdoubleyou, line 131: ...

graphql/course/fields.ts
```diff
@@ -71,4 +71,40 @@ export class ExtendedFieldsCourseResolver {
         const student = await getSessionStudent(context, studentId);
         return (await prisma.course_instructors_student.count({ where: { courseId: course.id, studentId: student.id } })) > 0;
     }
+
+    @FieldResolver((returns) => [Course])
```
Jonasdoubleyou, line 75: This must be `@Query` not FieldResolver - sorry, overlooked this earlier

graphql/course/fields.ts
```diff
@@ -71,4 +71,39 @@ export class ExtendedFieldsCourseResolver {
...
+                    {
+                        AND: [
+                            {
+                                course_instructors_student: {
+                                    some: {
+                                        studentId: studentId,
+                                    },
+                                },
+                            },
+                            courseSearchFilters,
```
Jonasdoubleyou, line 96: the search will not search shared courses

graphql/subcourse/mutations.ts
```diff
@@ -73,15 +73,28 @@ class PublicLectureInput {
 @Resolver((of) => GraphQLModel.Subcourse)
 export class MutateSubcourseResolver {
     @Mutation((returns) => GraphQLModel.Subcourse)
-    @AuthorizedDeferred(Role.ADMIN, Role.OWNER)
+    @AuthorizedDeferred(Role.INSTRUCTOR, Role.OWNER)
```
Jonasdoubleyou, line 76: INSTRUCTOR is enough, OWNER is useless, can be Authorized

graphql/subcourse/mutations.ts
```diff
@@ -73,15 +73,28 @@ class PublicLectureInput {
...
     async subcourseCreate(
         @Ctx() context: GraphQLContext,
         @Arg('courseId') courseId: number,
         @Arg('subcourse') subcourse: PublicSubcourseCreateInput,
         @Arg('studentId', { nullable: true }) studentId?: number
     ): Promise<GraphQLModel.Subcourse> {
         const course = await getCourse(courseId);
+        const student = await getSessionStudent(context, studentId);
+
         await hasAccess(context, 'Course', course);
```
Jonasdoubleyou, line 86: this hasAccess can be removed, we do the access check afterwards

graphql/subcourse/mutations.ts
Jonasdoubleyou: Add ADMIN

integration-tests/07_course.ts
```diff
@@ -113,21 +114,23 @@ export const subcourseOne = test('Create Subcourse', async () => {
...
     const {
         subcourseCreate: { id: subcourseId },
     } = await client.request(`
         mutation CreateSubcourse {
-            subcourseCreate(courseId: ${courseId} subcourse: {
+            subcourseCreate(courseId: ${courseId}, subcourse: {
                 minGrade: 5
                 maxGrade: 10
-                maxParticipants: 1
+                maxParticipants: 1
```
Jonasdoubleyou, line 126: whitespace?

integration-tests/07_course.ts
```diff
@@ -390,3 +393,44 @@ void test('Add / Remove another instructor', async () => {
...
+              id
+            }
+        }
+    `);
+
+    assert.ok(templateCourses.some((it) => it.id === courseId));
+});
+
+void test('Test template course search', async () => {
+    const { instructor: instructor1 } = await screenedInstructorTwo;
```
Jonasdoubleyou, line 418: instructorTwo -> instructor2

integration-tests/07_course.ts
```diff
@@ -390,3 +393,44 @@ void test('Add / Remove another instructor', async () => {
...
+
+    await client.request(`
+        mutation MarkCourseShared {
+            courseMarkShared(courseId: ${courseId}, shared: true){id}
+        }
+    `);
+
+    const { templateCourses } = await adminClient.request(`
+        query FindSharedCourses {
+            templateCourses(search: "", take: 1) {
```
Jonasdoubleyou, line 408: You might also want to test search with a non empty string

integration-tests/07_course.ts
```diff
@@ -390,3 +393,44 @@ void test('Add / Remove another instructor', async () => {
...
+    `);
+
+    assert.ok(templateCourses.some((it) => it.id === courseId));
+});
+
+void test('Test template course search', async () => {
+    const { instructor: instructor1 } = await screenedInstructorTwo;
+    const { instructor, courseId } = await subcourseOne;
+
+    await prisma.course.update({ where: { id: courseId }, data: { shared: false } });
```
Jonasdoubleyou, line 421: the two tests depend on each other - if they are run in parallel or in a different order, they both modify subcourseOne - either use one large test which is then sequential, or use two different subcourses

Jonasdoubleyou: Also please avoid updating the database in tests - we want to test the GraphQL interface so prefer that

integration-tests/07_course.ts
```diff
@@ -390,3 +393,44 @@ void test('Add / Remove another instructor', async () => {
...
+        }
+    `);
+
+    const { templateCourses } = await adminClient.request(`
+        query FindSharedCourses {
+            templateCourses(search: "", take: 1) {
+              id
+            }
+        }
+    `);
```
Jonasdoubleyou, line 412: There is currently no test that if an instructor shares a course, that another instructor can find it

graphql/course/fields.ts
Jonasdoubleyou: I guess you want AND [ OR [ own, shared ], search ]

paul-jaworski: As @realmayus explained to me, a course shall be returned as a template if:
- shared: true or
- the student is instructor of the course.

This or is handled by the outer OR in line 86. The inner AND in line 88 handles the second case along with the given search parameters. So it looks correct to me, but of course I might be wrong.

Jonasdoubleyou: Yeah well the search will currently only search the own courses, it should also search the shared courses.

integration-tests/07_course.ts
paul-jaworski: > the two tests depend on each other - if they are run in parallel or in a different order, they both modify subcourseOne - either use one large test which is then sequential, or use two different subcourses

Do multiple calls of subcourse one return the same subcourse object?

How can I obtain a second subcourse?

integration-tests/07_course.ts
paul-jaworski: This test case is also added (line 429). This assert also tests if the searchh by course name works.

graphql/course/fields.ts
paul-jaworski: > Yeah well the search will currently only search the own courses, it should also search the shared courses.

The query will also find shared courses since the shared:true condition is part of the outer OR block.

@realmayus What do you think?

realmayus: @paul-jaworski Yes, but it will return ***all*** shared courses, but we want to be able to filter (i.e. search within) the shared courses as well. If there are 1000 shared courses, it would return all 1000 of them, even if we have entered a very restrictive search query.

Currently the query does the following:
```
if course.student == student {
    if searchFilters.match(course) == true {
        return course
    }
} else if course.isShared {
    return course
}
```
what we want is:
```
if searchFilters.match(course) == true {
    if course.student == student {
        return course
    } else if course.isShared {
        return course
    }
}
```

Jonasdoubleyou (approved):
nice!

PR #915, opened 2023-12-03
fix: Prohibit usage of AuthorizedDeferred without context dependent roles

Jonasdoubleyou:
```
This suggests a programming mistake:

@AuthorizedDeferred(Role.INSTRUCTOR)
...
await hasAccess(course1);
```

This suggests the intent was to check access specific to a course, yet the access check is on generic roles. Usually one actually wants to use Role.OWNER instead

### DIFF
--- a/graphql/course/mutations.ts
+++ b/graphql/course/mutations.ts
@@ -175,7 +175,7 @@ export class MutateCourseResolver {
     }
 
     @Mutation((returns) => Boolean)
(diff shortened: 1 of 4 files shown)


Jonasdoubleyou:
It makes me thinking .. Maybe we should rename `Role.INSTRUCTOR` to `Role.ALL_INSTRUCTORS` to avoid this problem in the future?

realmayus (approved):

PR #918, opened 2023-12-04
feat: Synchronize Jobs

Jonasdoubleyou:
Prevents parallel execution of one job across all Dynos (i.e. one cannot manually run `_executeJob` if it is currently running on the Worker Dyno). To synchronize we use a table in the Postgres (as we do not have any other way to communicate between the Dynos).


graphql/authorizations.ts
```diff
@@ -344,6 +344,7 @@ export const authorizationEnhanceMap: Required<ResolversEnhanceMap> = {
     Achievement_template: allAdmin,
     User_achievement: allAdmin, // TODO change
     Achievement_event: allAdmin,
+    Job_run: { _all: nobody },
```
dhenkel92, line 347: why not admin only?

jobs/execute.ts
```diff
@@ -0,0 +1,105 @@
...
+    const server = http.createServer(app);
+
+    // Start listening
+    await new Promise<void>((res) => server.listen(port, res));
+    logger.info(`Server listening on port ${port}`);
+}
+
+if (process.env.METRICS_SERVER_ENABLED === 'true') {
+    startMetricsServer().catch((e) => logger.error('Failed to setup metrics server', e));
+}
```
dhenkel92, line 27: This part should stay in the scheduler. When running jobs from the web, it will use the metrics endpoint [from there](https://github.com/corona-school/backend/blob/master/web/server.ts#L105).

jobs/execute.ts
```diff
@@ -0,0 +1,105 @@
...
+            data: { endedAt: new Date() },
+        });
+
+        logger.info(`Finished Job '${name}'`);
+    } catch (error) {
+        logger.error(`Failure during Job Scheduling - This might leave the system in a locked state requiring manual cleanup! - ${error.message}`, error);
+        success = false;
+        // Eventually we now have a job run in the job_run table that has no endedAt,
+        // but which will never finish. To unlock this again, simply delete this entry
+        // (This should only happen in the rare case that the Dyno is killed (!) during execution)
```
dhenkel92, line 101: We should build some alert around this case. We can easily use the error above if you remove the variable from the message.

prisma/migrations/20231204145218_job_lock/migration.sql
```diff
@@ -0,0 +1,9 @@
+-- CreateTable
+CREATE TABLE "job_run" (
+    "job_name" TEXT NOT NULL,
+    "startedAt" TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "endedAt" TIMESTAMP(3),
```
dhenkel92, line 5: Why is started a `TIMESTAMP(6)` and ended a `TIMESTAMP(3)`?

graphql/authorizations.ts
Jonasdoubleyou: Because this includes the permission for creating new entries in the Job Run Table via GraphQL, which we don't want anyone to do (also all resolvers / mutations are currently not added, so in that sense "nobody can do nothing")

jobs/execute.ts
Jonasdoubleyou: Ah I see. Moved it to jobs/index.ts which is supposed to start that Dyno

jobs/execute.ts
Jonasdoubleyou: good point, splitted up into two log lines

prisma/migrations/20231204145218_job_lock/migration.sql
Jonasdoubleyou: Whoops, missed the annotation

PR #919, opened 2023-12-04
feat: Exclude unscreened helpers from campaigns

Jonasdoubleyou:


tobiasbork (approved):
Could add projectcoaching-screening, but certainly not necessary.

PR #923, opened 2023-12-05
fix: Retry Lock if rolled back by DB

Jonasdoubleyou:
If the DB transaction succeeded we know if another job is running or not and can abort / continue accordingly.

If the transaction was rolled back by the DB due to a deadlock, we do not know if two transactions on the same job name collided or unrelated transactions collided (due to limitations in Postgres deadlock detection). Thus in that case it makes sense to retry, as usually one concurrent transaction should be aborted, so if we try again there should be no concurrent transaction

### DIFF
--- a/jobs/execute.ts
+++ b/jobs/execute.ts
@@ -18,11 +18,11 @@ enum LockStatus {
     AQUIRED = 3,
 }
 
-export async function runJob(name: JobName): Promise<boolean> {
+export async function runJob(jobName: JobName): Promise<boolean> {
     let success = false;
 
     try {
-        logger.info(`Starting to run Job '${name}'`);
+        logger.info(`Starting to run Job '${jobName}'`, { jobName });
 
         // ---------- AQUIRE --------------
         // Prevent Job Runs running concurrently (across dynos), as jobs usually lack synchronization internally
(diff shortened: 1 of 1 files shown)


jobs/execute.ts
```diff
@@ -19,34 +30,61 @@ export async function runJob(name: JobName): Promise<boolean> {
...
+                );
+            } catch (error) {
+                // The transaction was aborted, likely because the DB rolled back the deadlock
+                lockStatus = LockStatus.ROLLBACK;
+                lockRetries -= 1;
+            }
+        } while (lockStatus === LockStatus.ROLLBACK && lockRetries > 0);
+
+        if (lockStatus === LockStatus.CONFLICT) {
+            return false;
```
dhenkel92, line 78: I think we should move the error from below to here
```        
logger.error(`Failure during Job Scheduling - This might leave the system in a locked state requiring manual cleanup!`, error);
```

jobs/execute.ts
```diff
@@ -19,34 +30,61 @@ export async function runJob(name: JobName): Promise<boolean> {
...
+        }
+
+        if (lockStatus === LockStatus.ROLLBACK) {
+            logger.error(`Failed to aquire Lock after 5 retries`);
```
dhenkel92, line 82: Let's add some more information to the log data? Like
```
{ jobName: name }
```

jobs/execute.ts
Jonasdoubleyou: Yes, moved it into the ROLLBACK case, as that is the "not well defined" state. The CONFLICT state can happen if the same job is triggered multiple times - I'll also increase the logging to error so that we notice if there are many conflicts still

dhenkel92 (approved):

PR #924, opened 2023-12-05
feat: achievement notification

LucasFalkowsky:
closes: https://github.com/corona-school/project-user/issues/987

What was done:

- notifications are now created when a user is rewarded with an achievement


common/notification/actions.ts
```diff
@@ -618,6 +618,10 @@ const _notificationActions = {
     // ACHIEVEMENT TEST ACTIONS
 
     /* ONBOARDING */
+    reward_issued: {
+        description: 'Reward issued',
+        sampleContext: {},
```
LomyW, line 623: Maybe we have to define a context here. I would wait for more informations here.
I could imagine that we could use the `achievement name` and `achievedAt`

common/achievement/index.ts
```diff
@@ -127,11 +127,12 @@ function injectRecordValue(condition: string, recordValue: number) {
     return condition;
 }
 
-async function awardUser(evaluationResult: number, userAchievement: UserAchievementTemplate) {
+async function rewardUser<ID extends ActionID>(evaluationResult: number, userAchievement: UserAchievementTemplate, event: ActionEvent<ID>) {
     let newRecordValue = null;
     if (typeof userAchievement.recordValue === 'number' && evaluationResult) {
         newRecordValue = evaluationResult;
     }
+    await actionTakenAt(new Date(event.at), event.user, 'reward_issued', event.context);
```
LomyW, line 135: Can we call the `actionTakenAt()` function after the update? Because if in any case the update does not work, we still send the notification...

LomyW: and also here: maybe the context will change

common/notification/actions.ts
```diff
@@ -618,6 +618,10 @@ const _notificationActions = {
     // ACHIEVEMENT TEST ACTIONS
 
     /* ONBOARDING */
+    reward_issued: {
```
LomyW, line 621: maybe rename to `user_achievement_rewarded` to make it more clear

LomyW (commented):
well done! we have to wait for more informations for the notification context

common/notification/actions.ts
LucasFalkowsky: user_achievemnet_reward_issued

common/notification/actions.ts
```diff
@@ -634,10 +634,26 @@ const _notificationActions = {
         description: 'Student / E-Mail verified',
         sampleContext: {},
     },
+
+    user_achievemnet_reward_issued: {
```
LomyW, line 638: just a spelling mistake, should be
`user_achievement_reward_issued`

PR #926, opened 2023-12-07
feat: refactor of aggregators

LomyW:
# Description
Refactoring the aggregators, that they just get values (number[]).

## ✅ What was done
- [x] adjusted aggregators
- [x] added `presenceOfEvents` aggregator for streaks
- [x] changed and cleaned types
- [x] adjust evaluate process to pass values to the aggregator

## Checklist before requesting a review
- [x] I have performed a self-review of my code

### DIFF
--- a/common/achievement/aggregator.ts
+++ b/common/achievement/aggregator.ts
@@ -6,36 +6,36 @@ type Aggregator = Record<string, AggregatorFunction>;
 
 export const aggregators: Aggregator = {
     sum: {
-        function: (buckets): number => {
-            const values = [];
-            for (const bucket of buckets) {
-                if (bucket.events.length > 0) {
-                    values.push(bucket.events.reduce((total, event) => total + event.value, 0));
-                }
-            }
-            return values.reduce((total, num) => total + num, 0);
+        function: (elements): number => {
+            return elements.reduce((total, num) => total + num, 0);
         },
     },
     count: {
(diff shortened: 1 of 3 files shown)


common/achievement/aggregator.ts
```diff
@@ -6,36 +6,36 @@ type Aggregator = Record<string, AggregatorFunction>;
...
+            let value = 0;
+            let afterNull = false;
+            for (const element of elements) {
+                if (element === 0) {
+                    afterNull = true;
                     break;
                 }
+
+                if (afterNull) {
+                    value += element;
```
dhenkel92, line 35: I think it should be `+= 1` just to make sure we don't have anything else than 1 in a streak event

common/achievement/aggregator.ts
```diff
@@ -6,36 +6,36 @@ type Aggregator = Record<string, AggregatorFunction>;
...
+    presenceOfEvents: {
+        function: (elements): number => {
+            return elements.length > 0 ? 1 : 0;
+        },
+    },
+    streak: {
+        function: (elements): number => {
+            let value = 0;
+            let afterNull = false;
+            for (const element of elements) {
```
dhenkel92, line 28: Is it always given that the newest events are the first in the array?

LomyW: yes u are totally right

PR #931, opened 2023-12-12
fix: global achievements

LucasFalkowsky:
what was changed:

- buckets now take the relation of a userAchievement to compare it with the events' relation
- buckets, if needed, filter events by their relation
- a new resolver, getInactiveAchievements, was introduced to get all inactive tiered achievements
- lectures are just counted for achievements, if they were not declined by the user


common/achievement/util.ts
```diff
@@ -24,26 +24,48 @@ export function getMetricsByAction<ID extends ActionID>(actionId: ID): Metric[]
...
-    return [relationType as RelationTypes, Number(id)];
+    return [relationType as RelationTypes, id];
 }
 
-export async function getBucketContext(relation: string): Promise<AchievementContextType> {
-    const [type, id] = getRelationTypeAndId(relation);
+export async function getBucketContext(relation: string, id: string): Promise<AchievementContextType> {
+    const [userType, userId] = getUserTypeAndIdForUserId(id);
+    const [relationType, relationId] = getRelationTypeAndId(relation);
+    if (relationId === 'all') {
```
LomyW, line 39: I would recommend not to check the relation id. My suggestion is to extend the `RelationTypes` with `match_all` and `subcourse_all` and still provide the id of the match / subcourse. This allows the events to be assigned to the individual matches / subcourses later. We would therefore save `match_all/1` in the user_achievement instead of `match/all`.

common/achievement/index.ts
```diff
@@ -82,8 +82,7 @@ async function checkUserAchievement<ID extends ActionID>(userAchievement: UserAc
         const evaluationResultValue =
             typeof evaluationResult.resultObject[dataAggregationKey] === 'number' ? Number(evaluationResult.resultObject[dataAggregationKey]) : null;
         const awardedAchievement = await rewardUser(evaluationResultValue, userAchievement, event);
-        const userAchievementContext: UserAchievementContext = {};
-        await createAchievement(awardedAchievement.template, userAchievement.userId, userAchievementContext);
+        await createAchievement(awardedAchievement.template, userAchievement.userId, event.context);
```
LomyW, line 85: the only problem I see with using the same context as in the events is that the action `student_joined_match_meeting` / `pupil_joined_match_meeting` triggers more than one achievement. So if we store `match_all/1` and the join meeting action is also used for a specific achievement - for example "Regelmäßiges Lernen", where we have to check the lecture for just one match - than the buckets will not work anymore.

LomyW: so maybe we need a own action for global_match / global_subcourse achievements? Would this solve this?

LomyW: so what I mean is to store more achievement_events, that are more specific for each metric.

common/achievement/index.ts
```diff
@@ -101,7 +100,14 @@ async function isAchievementConditionMet(achievement: UserAchievementTemplate) {
...
     }
 
-    const { conditionIsMet, resultObject } = await evaluateAchievement(condition, conditionDataAggregations as ConditionDataAggregations, metrics, recordValue);
+    const { conditionIsMet, resultObject } = await evaluateAchievement(
+        condition,
+        conditionDataAggregations as ConditionDataAggregations,
+        metrics,
+        recordValue,
+        achievement.userId,
+        achievement.context
```
LomyW, line 109: why do we not destructre the context and userId from the achievement - like the recordValue i.e.

LomyW (changes requested):
some questions 😇

common/achievement/get.ts
```diff
@@ -73,7 +123,9 @@ const assembleAchievementData = async (userAchievements: User_achievement[], use
             condition,
             currentAchievementTemplate.conditionDataAggregations as ConditionDataAggregations,
             currentAchievementTemplate.metrics,
-            userAchievements[currentAchievementIndex].recordValue
+            userAchievements[currentAchievementIndex].recordValue,
+            user.userID,
+            userAchievements[currentAchievementIndex].context
         );
         currentValue = dataAggregationKeys.map((key) => evaluationResult.resultObject[key]).reduce((a, b) => a + b, 0);
```
LomyW, line 130: I think the `currentValue` doesn't work right here for sequential achievements, because we always check the current template
Bildschirmfoto 2023-12-18 um 08 42 55

LomyW (changes requested):
😇

common/achievement/get.ts
LucasFalkowsky: The issue was found and fixed!

dhenkel92 (commented):
I've created an opinionated PR to unify the new global array with the existing one.

https://github.com/corona-school/backend/pull/934

common/achievement/evaluate.ts
```diff
@@ -6,15 +6,26 @@ import swan from '@onlabsorg/swan-js';
...
 const logger = getLogger('Achievement');
 
 export async function evaluateAchievement(
     condition: string,
     dataAggregation: ConditionDataAggregations,
     metrics: string[],
-    recordValue: number
+    recordValue: number,
+    userId: string,
+    context: Prisma.JsonValue
```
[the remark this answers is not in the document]
LucasFalkowsky: We pass the relation directly now

common/achievement/bucket.ts
```diff
@@ -43,42 +54,52 @@ export const bucketCreatorDefs: BucketCreatorDefs = {
...
-                buckets.push({
-                    kind: 'time',
-                    startTime: weeksBefore.startOf('week').toDate(),
-                    endTime: weeksBefore.endOf('week').toDate(),
-                });
-            }
+            context[context.type].forEach((contextType) => {
+                if (!contextType.lecture || contextType.lecture.length === 0) {
+                    return;
+                }
```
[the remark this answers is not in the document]
dhenkel92: Okay, but the main question still remains. Do we need weekly buckets by relation?

common/achievement/get.ts
```diff
@@ -5,9 +5,58 @@ import { User } from '../user';
 import { ConditionDataAggregations } from './types';
 import { getAchievementState, getCurrentAchievementTemplateWithContext, transformPrismaJson } from './util';
 import { evaluateAchievement } from './evaluate';
+import { Prisma } from '@prisma/client';
 
 // TODO: getAchievementById -> passed user and achievementId to return a single achievement
 // TODO: resolver for nextSteps -> get active sequential achievements and important information
+// Inactive achievements are acheievements that are not yet existing but could be achieved in the future.
+// They are created for every template in a Tiered achievements group that is not yet used as a achievement for a specific user.
+const getInactiveAchievements = async (user: User): Promise<Achievement[]> => {
```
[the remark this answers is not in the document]
LucasFalkowsky: was renamed to furtherAchievements

common/achievement/bucket.ts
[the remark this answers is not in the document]
LucasFalkowsky: Oh yeah you are right.

dhenkel92 (approved):

PR #933, opened 2023-12-15
Feat: new resolvers

LucasFalkowsky:
What was done:

- 2 new resolvers were introduced (getAchievementById & getNextStepAchievements)


graphql/user/fields.ts
```diff
@@ -210,6 +210,18 @@ export class UserFieldsResolver {
         return await getAppointmentsForUser(user, take, skip, cursor, direction);
     }
 
+    @FieldResolver((returns) => [Achievement])
+    @Authorized(Role.ADMIN, Role.OWNER)
+    async achievementById(@Ctx() context: GraphQLContext, @Arg('id') id: number): Promise<Achievement> {
```
LomyW, line 215: can we just call it `achievement`

common/achievement/get.ts
```diff
@@ -1,14 +1,37 @@
...
+        include: { template: true },
+    });
+    const achievement = await assembleAchievementData([userAchievement], user);
+    return achievement;
+};
+
+// Next step achievements are sequential achievements that are currently active and not yet completed. They get displayed in the next step card section.
+const getNextStepAchievements = async (user: User): Promise<Achievement[]> => {
+    const userAchievements = await prisma.user_achievement.findMany({
+        where: { userId: user.userID, isSeen: false, template: { type: achievement_type_enum.SEQUENTIAL } },
```
LomyW, line 21: do we also have to check here for `achievedAt != null`?

LomyW (commented):
looks good to me 👍🏼

graphql/user/fields.ts
LucasFalkowsky: done!

common/achievement/get.ts
LucasFalkowsky: No I don't think so. If that was the case we could not get all steps necessary to know the currentStep value. Or the maxSteps value.

graphql/user/fields.ts
```diff
@@ -210,6 +210,18 @@ export class UserFieldsResolver {
         return await getAppointmentsForUser(user, take, skip, cursor, direction);
     }
 
+    @FieldResolver((returns) => [Achievement])
+    @Authorized(Role.ADMIN, Role.OWNER)
+    async achievement(@Ctx() context: GraphQLContext, @Arg('id') id: number): Promise<Achievement> {
```
dhenkel92, line 215:
```suggestion
    async achievements(@Ctx() context: GraphQLContext, @Arg('id') id: number): Promise<Achievement> {
```

graphql/user/fields.ts
```diff
@@ -210,6 +210,18 @@ export class UserFieldsResolver {
...
+        const achievement = await getAchievementById(context.user, id);
+        return achievement;
+    }
+    @FieldResolver((returns) => [Achievement])
+    @Authorized(Role.ADMIN, Role.OWNER)
+    async nextStepAchievements(@Ctx() context: GraphQLContext): Promise<Achievement[]> {
```
dhenkel92, line 221: Maybe rename?
```suggestion
    async achievementsNextSteps(@Ctx() context: GraphQLContext): Promise<Achievement[]> {
```

graphql/user/fields.ts
LucasFalkowsky: no this is achievement. Singular.

graphql/user/fields.ts
```diff
@@ -210,6 +210,18 @@ export class UserFieldsResolver {
         return await getAppointmentsForUser(user, take, skip, cursor, direction);
     }
 
+    @FieldResolver((returns) => [Achievement])
```
dhenkel92, line 213: If it's supposed to return only one achievement, it should be reflected in the annotaion.
```suggestion
    @FieldResolver((returns) => Achievement)
```

graphql/user/fields.ts
LucasFalkowsky: Well this actually is not really the next step of an achievement. The sequential achievements get diesplayed as next step card on the front page.
So if you really would prefer another name we should go for sth like achievementsAsNextSteps. You have any more ideas?

PR #934, opened 2023-12-18
try to unify matches with global_matches

dhenkel92:


common/achievement/types.ts
```diff
@@ -109,26 +109,23 @@ export type EvaluationResult = {
 export type RelationTypes = 'match' | 'subcourse' | 'global_match' | 'global_subcourse'; // match_all, subcourse_all, all
 
 type ContextLecture = Pick<Lecture, 'start' | 'duration'>;
+export type ContextMatch = {
+    id: number;
+    relation: string | null; // will be null if searching for all matches
```
LomyW, line 114: is the relation really null here? What about `global_match` / `global_subcourse`?

`export type RelationTypes = 'match' | 'subcourse' | 'global_match' | 'global_subcourse'; `

LomyW (commented):
just a question 🙌🏼

common/achievement/types.ts
dhenkel92: My idea was that it stores the real relation `match/<id>` in case we need it. In this PR it will be null if we are searching for all matches.

LucasFalkowsky (commented):
I love the idea of putting together the different types. I commented on some stuff that we might have to further think about.

common/achievement/util.ts
```diff
@@ -33,55 +33,57 @@ export function getRelationTypeAndId(relation: string): [type: RelationTypes, id
     return [relationType as RelationTypes, id];
 }
 
-export async function getBucketContext(relation: string, id: string): Promise<AchievementContextType> {
-    const [userType, userId] = getUserTypeAndIdForUserId(id);
-    const [relationType, relationId] = getRelationTypeAndId(relation);
+// TODO: fix naming
+export async function getBucketContext(myUserID: string, relation?: string): Promise<AchievementContextType> {
```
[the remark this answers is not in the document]
dhenkel92: It's possible to have all cases
relation `match/<id>` -> specific match
relation `match` -> all matches
relation `subcourse` -> all subcourses
relation null -> everything

> If this was the case though, then merging would leave us the challenge to us the right type on prisma find many

Could you please elaborate on this a little bit more?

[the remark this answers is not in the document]
LucasFalkowsky: got it!

The last part is a comment on my comment of leaving out any type differentiation when spreading the buckets together later anyways.
prisma[relationType] does not work when relation is null.

dhenkel92:
I'm not sure if we should combine the subcourse with the match information. Theoretically, it would be possible, but it feels somewhat wrong.

LucasFalkowsky (approved):
This code is good anyway, so you can merge it.

common/achievement/util.ts
```diff
@@ -33,55 +33,57 @@ export function getRelationTypeAndId(relation: string): [type: RelationTypes, id
...
+    const [userType, userId] = getUserTypeAndIdForUserId(myUserID);
```
Jonasdoubleyou, line 38: userId and myUserId are confusing.

dhenkel92: yeah, this is why there is an open todo 🙄 
I forgot to rename it...

PR #941, opened 2024-01-03
Multiple match dissolve reasons

realmayus:
Part of https://github.com/corona-school/project-user/issues/494#issuecomment-1791440154

- implement migration from dissolve_reason type to dissolve_reason[] in separate DB column
- add checks


prisma/migrations/20240103181300_add_dissolve_reasons_array/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "match" ADD COLUMN     "dissolveReasons" "dissolve_reason"[] DEFAULT ARRAY[]::"dissolve_reason"[];
```
Jonasdoubleyou, line 2: Would be easier to just UPDATE here and then directly drop the old column, but if you prefer to do it in a separate step thats also fine

Jonasdoubleyou (approved):

PR #943, opened 2024-01-07
Create evaluate unit tests

dhenkel92:
First set of unit tests that are covering achievement evaluation as well as aggregators and bucket creators.

### DIFF
--- a/common/achievement/aggregator.spec.ts
+++ b/common/achievement/aggregator.spec.ts
@@ -1,0 +1,113 @@
+import { aggregators } from './aggregator';
+
+describe('test sum aggregator', () => {
+    const tests: {
+        name: string;
+        elements: number[];
+        expected: number;
+    }[] = [
+        {
+            name: 'should sum all elements in array',
+            elements: [1, 2, 3, 4, 5],
+            expected: 15,
+        },
+        {
+            name: 'should return 0 for empty array',
+            elements: [],
+            expected: 0,
+        },
+    ];
+
+    it.each(tests)('$name', ({ elements, expected }) => {
+        expect(aggregators['sum'].function(elements)).toEqual(expected);
+    });
+});
+
+describe('test count aggregator', () => {
+    const tests: {
+        name: string;
+        elements: number[];
+        expected: number;
+    }[] = [
+        {
+            name: 'should count all elements in array',
+            elements: [1, 2, 3, 4, 5],
+            expected: 5,
+        },
+        {
+            name: 'should return 0 for empty array',
+            elements: [],
+            expected: 0,
+        },
+        {
+            name: 'should skip elements that are zero',
+            elements: [1, 0, 3, 0, 5],
+            expected: 3,
+        },
+    ];
+
+    it.each(tests)('$name', ({ elements, expected }) => {
+        expect(aggregators['count'].function(elements)).toEqual(expected);
+    });
+});
+
+describe('test presenceOfEvents aggregator', () => {
+    const tests: {
+        name: string;
+        elements: number[];
+        expected: number;
+    }[] = [
+        {
+            name: 'should return 1 if at least one element is present',
+            elements: [1, 2, 3, 4, 5],
+            expected: 1,
+        },
+        {
+            name: 'should return 0 for empty array',
+            elements: [],
+            expected: 0,
+        },
+        {
+            name: 'should return 0 if all elements are zero',
+            elements: [0, 0, 0, 0, 0],
+            expected: 0,
+        },
+    ];
+
+    it.each(tests)('$name', ({ elements, expected }) => {
+        expect(aggregators['presenceOfEvents'].function(elements)).toEqual(expected);
+    });
+});
+
+describe('test lastStreakLength aggregator', () => {
+    const tests: {
+        name: string;
+        elements: number[];
+        expected: number;
+    }[] = [
+        {
+            name: 'should get a streak of five',
+            elements: [1, 2, 3, 4, 5],
+            expected: 5,
+        },
+        {
+            name: 'should not have a streak if no elements are present',
+            elements: [],
+            expected: 0,
+        },
+        {
+            name: 'should reset streak if zero in chain',
+            elements: [1, 1, 0, 1, 1],
+            expected: 2,
+        },
+        {
+            name: 'should have a streak of 0 if first element is zero',
+            elements: [0, 1, 1, 1, 1],
+            expected: 0,
+        },
+    ];
+
+    it.each(tests)('$name', ({ elements, expected }) => {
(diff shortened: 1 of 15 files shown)


common/achievement/bucket.spec.ts
```diff
@@ -0,0 +1,280 @@
+import moment from 'moment';
+import { bucketCreatorDefs } from './bucket';
+import { BucketCreatorContext, ContextMatch, ContextSubcourse, TimeBucket } from './types';
+
+describe('test create buckets by_lecture_start', () => {
+    // TODO: think about setting dow globally
+    moment.updateLocale('de', { week: { dow: 1 } });
```
dhenkel92, line 7: I think we should set `dow: 1` globally. Otherwise, the week is starting on sundays.

common/achievement/aggregator.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+            expected: 5,
+        },
+        {
+            name: 'should not have a streak if no elements are present',
+            elements: [],
+            expected: 0,
+        },
+        {
+            name: 'should reset streak if zero in chain',
+            elements: [1, 1, 0, 1, 1],
```
Jonasdoubleyou, line 100: How is the chain sorted? Maybe change to [1, 0, 1, 1] to also check the order?

common/achievement/bucket.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+                    kind: 'time',
+                    startTime: moment('2023-08-14T21:50:00.000Z').toDate(),
+                    endTime: moment('2023-08-14T23:10:00.000Z').toDate(),
+                    relation: 'match',
+                },
+            ],
+            matches: [{ id: 1, relation: 'match', lecture: [{ start: today.toDate(), duration: 60 }] }],
+        },
+        {
+            name: 'should create multiple buckets for a match with multiple lectures',
```
Jonasdoubleyou, line 34: Huh, matches have lectures? Don't we call them appointments elsewhere?

common/achievement/bucket.ts
```diff
@@ -5,12 +5,13 @@ type BucketCreatorDefs = Record<string, BucketFormula>;
 
 function createLectureBuckets<T extends ContextMatch | ContextSubcourse>(data: T): TimeBucket[] | null {
```
Jonasdoubleyou, line 6: null can be removed from return type

common/achievement/evaluate.ts
```diff
@@ -53,21 +54,21 @@ async function _evaluateAchievement(
 
         const aggregator = dataAggregationObject.aggregator;
 
-        const eventsForMetric = eventsByMetric[metricName];
+        const eventsForMetric = eventsByMetric[metricName] || [];
```
Jonasdoubleyou, line 57: I would prefer ?? over the more generic ||

common/achievement/evaluate.ts
```diff
@@ -16,10 +16,11 @@ async function _evaluateAchievement(
...
     condition: string,
     dataAggregation: ConditionDataAggregations,
-    metrics: string[],
     recordValue: number,
-    relation?: string | null
-): Promise<EvaluationResult> {
+    relation?: string
+): Promise<EvaluationResult | undefined> {
+    // We only care about metrics that are used for the data aggregation
+    const metrics = Object.values(dataAggregation).map((entry) => entry.metric);
```
Jonasdoubleyou, line 23: Off topic to this PR: Why is there achievementTemplate.metrics at all then? Also wouldn't it be cleaner to just pass in the achievement template in the parameter?

jest/singletons.ts
```diff
@@ -0,0 +1,20 @@
...
+}));
+
+beforeEach(() => {
+    mockReset(prismaMock);
+});
+
+export const prismaMock = mockDeep<PrismaClient>() as unknown as DeepMockProxy<{
+    // this is needed to resolve the issue with circular types definition
+    // https://github.com/prisma/prisma/issues/10203
+    [K in keyof PrismaClient]: Omit<PrismaClient[K], 'groupBy'>;
```
Jonasdoubleyou, line 19: 😅

Jonasdoubleyou (approved):

common/achievement/evaluate.ts
dhenkel92: > Why is there achievementTemplate.metrics at all then?

That's a good point. I think the main idea was to have a field, which can be quried, like
```
SELECT * FROM achievement_templates WEHRE metric in (....)
```
Nevertheless, afaics it's not used like this right now.
https://github.com/corona-school/backend/blob/feat/achievements/common/achievement/template.ts#L48

> Also wouldn't it be cleaner to just pass in the achievement template in the parameter?

Why do you think so?

common/achievement/bucket.spec.ts
dhenkel92: I don't know 😅 
I just followed the [prisma schema naming](https://github.com/corona-school/backend/blob/ceaf0f2a56ee769643c6948369862049dfd752f2/prisma/schema.prisma#L410).

common/achievement/aggregator.spec.ts
dhenkel92: The next test is ensuring that the evaluation oder is correct. Nevertheless, I'll update this test to do the same.

common/achievement/bucket.ts
dhenkel92: Good catch!

jest/singletons.ts
dhenkel92: Amazing, right 😆 

I tried to enable stricktNullChecking, but this opens another huge box of problems.

common/achievement/evaluate.ts
Jonasdoubleyou: Cause its "evaluate achievement", why wouldn't it need the achievement template as a whole?

Jonasdoubleyou: >  I think the main idea was to have a field, which can be queried

I see, I think this is now obsolete as we cache them in memory anyways. Also I fear that this could be a source of bugs if they deviate and some parts of the code look at one or the other

dhenkel92: > I see, I think this is now obsolete as we cache them in memory anyways. Also I fear that this could be a source of bugs if they deviate and some parts of the code look at one or the other

I move the comment to the main PR

> Cause its "evaluate achievement", why wouldn't it need the achievement template as a whole?

I don't know. I like that you see the important information that you from the achievement right in the function definition. I have at least the feeling that it's easier to understand.
" We want to evaluate the achievement, and therefore we need a condition, the record value and the data aggr."

Jonasdoubleyou: Also reasonable

PR #944, opened 2024-01-09
feat: achievement system documentation

LucasFalkowsky:
what was done:
a documentation for the backend has ben implemented inside the common/achievement fonder.

### DIFF
--- a/common/achievement/Readme.md
+++ b/common/achievement/Readme.md
@@ -1,4 +1,192 @@
 # Achievement
 
-Achievements are a part of the Gamification.
-Documentation follows...
+Achievements are a part of the Gamification. They are stored in the database as user_achievement and should not be mistaken with the achievement_template database table. The distinct differences are explained in Chapter [1.1 Database Models](#database-models)
+
+### Table of Contents
+
+1.  [Features](#features)
+    1.1. [Database Models](#database-models)
+    1.2. [Components](#components)
+    1.3. [Resolvers](#resolvers)
+2.  [Achievement System](#achievement-system)
+    2.1. [Reward Action](#reward-action)
+    2.2. [Get or Create User Achievement](#get-or-create-user-achievement)
+    2.3. [Create Achievement](#create-achievement)
+    2.4. [Check User Achievement](#check-user-achievement)
+    2.5. [Evaluate Achievement](#evaluate-achievement) & [Aggregation Functions](#aggregation-functions)
+
+## 1. Features
+
+-   Create Achievement Templates: Administrators can create new achievement templates inside the database
+-   Create Achievements: New achievements get created when related events are triggered by a users action or when previous achievements of the same group are marked as achieved
+-   Evaluate Achievements: Achievements are evaluated for fulfillment of the conditions to either issue the reward or update a started achievement.
+
+### 1.1. Database Models
+
+| achievement_template                                                                  | user_achievement                                                                                                      |
+| :------------------------------------------------------------------------------------ | :-------------------------------------------------------------------------------------------------------------------- |
+| predefined blueprint in the database that is used to create new achievements from it. | achievement data related to a user that stores information about the achievement, its status and when it was achieved |
+
+#### achievement_template
+
+| field                                                               | description                                                                                                                                                             |
+| :------------------------------------------------------------------ | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `id`                                                                | unique id                                                                                                                                                               |
+| `name`                                                              | displayed name of the template                                                                                                                                          |
+| `metrics`                                                           | used to identify all events related to a achievement template                                                                                                           |
+| `templateFor`                                                       | discriminator to clarify a relation of an achievement template to 1. one match, 2. all matches, 3. one subcourse, 4. all subcourses or no relation                      |
+| `group`                                                             | Affiliation of templates with the same group                                                                                                                            |
+| `groupOrder`                                                        | indexes to keep the templates inside group arrays in order                                                                                                              |
+| `stempName`                                                         | exclusively used for sequential achievements to be displayed as a step of this template group                                                                           |
+| `type`                                                              | achievement types `SEQUENTIAL`, `TIERED` and `STREAK`                                                                                                                   |
+| `subtitle`                                                          | subtitle of an achievement                                                                                                                                              |
+| `description`                                                       | description text of an achievement                                                                                                                                      |
+| `image`                                                             | image of a achievement / an achievement step                                                                                                                            |
+| `achievedImage`                                                     | exclusively used for streak achievements that are currently achieved                                                                                                    |
+| `actionName`                                                        | optional button label                                                                                                                                                   |
+| `actionRedirectLink`                                                | optional redirect reference for buttons                                                                                                                                 |
+| `actionType`                                                        | type to define what icon is displayed for an achievement action                                                                                                         |
+| `actionName`                                                        | label text for an achievement action                                                                                                                                    |
+| `condition`                                                         | condition to be met for a reward to be issued for an achievement                                                                                                        |
+| <div id="conditionDataAggregation">`conditionDataAggregation`</div> | context information on what buckets to create for the achievements metrics and how to aggregate action events for these conditions to check if they have been fulfilled |
+| `user_achievement`                                                  | all user_achievements that share a relation with this achievement_template                                                                                              |
+
+#### user_achievement
+
+| field         | description                                                                    |
+| :------------ | :----------------------------------------------------------------------------- |
+| `id`          | unique id                                                                      |
+| `templateId`  | id of the related achievement_template                                         |
+| `template`    | related achievement_template found by the templateId                           |
+| `userId`      | id of the related user                                                         |
+| `group`       | copy of the related templates group field                                      |
+| `groupOrder`  | copy of the related templates groupOrder field                                 |
+| `isSeen`      | set true when a user has viewed an achieved achievement on their progress page |
+| `achievedAt`  | timestamp set, if a achievements condition was met                             |
+| `recordValue` | record value on how long the longest streak was held up                        |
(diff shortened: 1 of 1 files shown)


common/achievement/Readme.md
```diff
@@ -1,4 +1,192 @@
...
+    2.4. [Check User Achievement](#check-user-achievement)
+    2.5. [Evaluate Achievement](#evaluate-achievement) & [Aggregation Functions](#aggregation-functions)
+
+## 1. Features
+
+-   Create Achievement Templates: Administrators can create new achievement templates inside the database
+-   Create Achievements: New achievements get created when related events are triggered by a users action or when previous achievements of the same group are marked as achieved
+-   Evaluate Achievements: Achievements are evaluated for fulfillment of the conditions to either issue the reward or update a started achievement.
+
+### 1.1. Database Models
```
LomyW, line 24: I miss the database model `achievement_event`.

common/achievement/Readme.md
```diff
@@ -1,4 +1,192 @@
...
+	end
+	subgraph isAchievementConditionMet
+		BA(isAchievementConditionMet) -.-> BB(evaluateAchievement)
+		BB -.-> |return evaluation result| BC[Data Aggregation - Chapter 2.5]
+	end
+```
+
+[Go to Data Aggregation - Chapter 2.5](#data-aggregation)
+
+### 2.1. Reward Action
```
LomyW, line 129: I think we should add some more details of tracking events here. Soo that we can clearify, that each action which is stored in a metric, tracks an event.

common/achievement/Readme.md
```diff
@@ -1,4 +1,192 @@
 # Achievement
 
-Achievements are a part of the Gamification.
-Documentation follows...
+Achievements are a part of the Gamification. They are stored in the database as user_achievement and should not be mistaken with the achievement_template database table. The distinct differences are explained in Chapter [1.1 Database Models](#database-models)
+
+### Table of Contents
```
LomyW, line 5: In my opinion we should also describe the concept of the whole achievement system, because we can't link the FigJam File of the concept and we should use this documentation as single source. Maybe we can add another markdown file for `Metric.md` and `Bucket.md` (maybe for `Condition.md`, because we have to explain how the condition should be defined), that we can explain what the idea behind the concepts is. So that we put all informations from the concept into the documentation.

I think this diagram shows the best (thanks to Daniel) how buckets work.


```
 . . . . . . . .    -> Raw events
|__||__||__||__|    -> Buckets based on type
 \ / \ / \ /\ /     -> Aggregate bucket events to single value
  .   .   .  .      -> Aggregate values
   \        /
    \      /
       .     -> result for condition
```

LomyW (commented):
Nice documentation! Some ideas ✨

common/achievement/Readme.md
LucasFalkowsky: - [ ] A Graphic for Bucket aggregation was introduced to the documentation
- [ ] An explanation for metrics was added to the documentation
- [ ] The concept for conditions and how to write them was added

LucasFalkowsky: I would still keep this all in one file. The advantage of which offering the ability to reference the chapters, where to find information e.g about the buckets, with a hyper reference. This could be very beneficial when searching for information in the documentation.

common/achievement/Readme.md
```diff
@@ -1,4 +1,338 @@
...
+
+All operations and and types can be found in the [swan-js documentation](https://github.com/onlabsorg/swan-js/blob/main/docs/swan.md).
+
+### 4.2. Writing Conditions
+
+When writing conditions for an achievement, keep in mind that the [conditionDataAggregation](#conditionDataAggregation) will produce values that should be used for such conditions.
+
+#### How conditionDataAggregations are built
+
+A conditionDataAggregation has the form of a json object and can contain aggregation information for multiple conditions. In order to understand this process, it should be clear what metrics, [aggregators](#aggregators), [bucket aggregators](#bucket-aggregator), and [buckets](#buckets) are. The Objects are structured as shown below:
```
LucasFalkowsky, line 288: @LomyW please expalin metrics in extra chapter.

common/achievement/Readme.md
```diff
@@ -1,4 +1,352 @@
...
+
+### 1.1. Database Models
+
+| achievement_template                                                                  | user_achievement                                                                                                      |
+| :------------------------------------------------------------------------------------ | :-------------------------------------------------------------------------------------------------------------------- |
+| predefined blueprint in the database that is used to create new achievements from it. | achievement data related to a user that stores information about the achievement, its status and when it was achieved |
+
+#### achievement_template
+
+| field                                                               | description                                                                                                                                                             |
```
dhenkel92, line 43: I like the idea of these tables, but I don't think they are adding a lot of value here.
I would suggest to remove them from this doc, but move the description to the `schema.prisma`.

dhenkel92: Maybe it makes sense to replace the tables with a high level ER diagram. It doesn't need to be complete as it will be outdated over time anyways.

common/achievement/Readme.md
```diff
@@ -1,4 +1,352 @@
...
+## 2. Achievement System
+
+The achievement system is set up as follows
+
+```mermaid
+flowchart TD
+	A(actionTaken) --> B(rewardActionTaken)
+	C[Achievement Template] --> getOrCreateUserAchievement
+	D[User Id] --> getOrCreateUserAchievement
+	E[Action Event Context] --> getOrCreateUserAchievement
```
dhenkel92, line 126:
```suggestion
	C[/Achievement Template/] --> getOrCreateUserAchievement
	D[/User Id/] --> getOrCreateUserAchievement
	E[/Action Event Context/] --> getOrCreateUserAchievement
```

common/achievement/Readme.md
```diff
@@ -1,4 +1,352 @@
...
+3.  `getFurtherAchievements`: function to get the achievement templates, which do not yet have an associated user achievement for the corresponding user but belong to an already initialized achievement group, as inactive achievements.
+4.  `getUserAchievements`: function to get all user achievements found for a user
+
+## 2. Achievement System
+
+The achievement system is set up as follows
+
+```mermaid
+flowchart TD
+	A(actionTaken) --> B(rewardActionTaken)
```
dhenkel92, line 123:
```suggestion
	A(actionTaken) --> B[rewardActionTaken]
```

common/achievement/Readme.md
```diff
@@ -1,4 +1,352 @@
...
+	H --> L{condition Met}
+	L -->|true| I
+	L -->|false| M(updateUserAchievement)
+	I(rewardUser) --> J[Awarded Achievement]
+	J --> K(createUserAchievement)
+
+	subgraph getOrCreateUserAchievement
+		AA(findUserAchievement) -.-> AB{User Achievement ?}
+		AB -.-> |true| AD(return User Achievement)
+		AB -.-> |false| AC(createUserAchievement)
```
dhenkel92, line 141:
```suggestion
		AB -.-> |false| AC[createUserAchievement]
```

common/achievement/Readme.md
```diff
@@ -1,4 +1,352 @@
...
+2.  `getNextStepAchievements`: function to get a users achievements that will be displayed as a next step card on their dashboard
+3.  `getFurtherAchievements`: function to get the achievement templates, which do not yet have an associated user achievement for the corresponding user but belong to an already initialized achievement group, as inactive achievements.
+4.  `getUserAchievements`: function to get all user achievements found for a user
+
+## 2. Achievement System
+
+The achievement system is set up as follows
+
+```mermaid
+flowchart TD
```
dhenkel92, line 122: I like the idea of flowcharts ❤️ 

I would suggest focusing more on the concepts rather than simply mirroring what the code does. For instance, in the code, we have a wrapper function (isAchievementConditionMet) around evaluateAchievement. While it may make sense in that context, it doesn't add value to this diagram. The diagram should aid in understanding the underlying idea; the specifics can be found in the code, and these are likely to change over time.

LucasFalkowsky: I removed the explainations of the functions and just kept the basic process.

prisma/migrations/20231123075042_add_achievements_achieved_image/migration.sql
dhenkel92, line 1: 🤔  I guess this file should not be here?

graphql/course/mutations.ts
dhenkel92, line 1: 🤔  I guess this file should not be here?

common/courses/states.ts
```diff
@@ -81,6 +81,18 @@ export async function allowCourse(course: Course, screeningComment: string | nul
             await publishSubcourse(subcourse);
         }
     }
+
```
dhenkel92, line 84: 🤔  I guess this file should not be here?

common/achievement/Readme.md
```diff
@@ -1,4 +1,367 @@
...
+
+### 2.2. Get or Create User Achievement
+
+The `getOrCreateUserAchievement` function is trying to find a pre-existing user achievement that matches the values passed, in order to return this achievement to our [`rewardActionTaken`](#reward-action) function. If no user achievement was found the [`createAchievement`](#create-achievement) function is triggered to return the newly create user_achievement instead.
+
+### 2.3. Create Achievement
+
+Inside the `createAchievement` function all of the users pre-existing achievements for the template, passed down from the [`rewardActionTaken`](#reward-action) function, are taken to determine the groupOrder of the following achievement that has to be created. The `groupOrder` value then is used to create a new user_achievement in the database for the achievement_template inside the achievement template group array, that has the matching value in its `groupOrder` field. The creation process is not finished, if the groupOrder of the subsequent achievement was outside the given range of group elements, meaning the group was already finished and in no need to be continued.
```
Jonasdoubleyou, line 181: What is the difference between sequential, streak etc.?

LomyW (approved):
Nice documentation! 💯

PR #951, opened 2024-01-16
Feat: offer course template

LucasFalkowsky:
### What was done:

- new metrics for offer course achievement
- new templates for offer course achievement
- connections to mutations

### DIFF
--- a/common/achievement/index.ts
+++ b/common/achievement/index.ts
@@ -129,7 +129,10 @@ async function checkUserAchievement<ID extends ActionID>(userAchievement: UserAc
         const evaluationResultValue =
             typeof evaluationResult.resultObject[dataAggregationKey] === 'number' ? Number(evaluationResult.resultObject[dataAggregationKey]) : null;
         const awardedAchievement = await rewardUser(evaluationResultValue, userAchievement, event);
-        await createAchievement(awardedAchievement.template, userAchievement.userId, event.context);
+        const newAchievement = await createAchievement(awardedAchievement.template, userAchievement.userId, event.context);
+        if (newAchievement) {
+            await checkUserAchievement(newAchievement, event);
+        }
     } else {
         await prisma.user_achievement.update({
             where: { id: userAchievement.id },
--- a/common/achievement/index.ts
+++ b/common/achievement/index.ts
@@ -129,10 +129,7 @@ async function checkUserAchievement<ID extends ActionID>(userAchievement: UserAc
         const evaluationResultValue =
             typeof evaluationResult.resultObject[dataAggregationKey] === 'number' ? Number(evaluationResult.resultObject[dataAggregationKey]) : null;
         const awardedAchievement = await rewardUser(evaluationResultValue, userAchievement, event);
-        const newAchievement = await createAchievement(awardedAchievement.template, userAchievement.userId, event.context);
-        if (newAchievement) {
-            await checkUserAchievement(newAchievement, event);
-        }
+        await createAchievement(awardedAchievement.template, userAchievement.userId, event.context);
     } else {
         await prisma.user_achievement.update({
             where: { id: userAchievement.id },
--- a/common/achievement/types.ts
+++ b/common/achievement/types.ts
@@ -114,6 +114,6 @@ export type ContextSubcourse = {
 
 export type AchievementContextType = {
     user?: User;
-    match: ContextMatch[];
-    subcourse: ContextSubcourse[];
+    match?: ContextMatch[];
+    subcourse?: ContextSubcourse[];
 };
--- a/common/achievement/util.ts
+++ b/common/achievement/util.ts
@@ -98,15 +98,9 @@ export function transformPrismaJson(user: User, json: Prisma.JsonValue): Achieve
     if (!keys) {
         return null;
     }
-    const transformedJson: AchievementContextType = {
-        user: user,
-        match: json['match'] ? json['match'] : undefined,
-        subcourse: json['subcourse'] ? json['subcourse'] : undefined,
-    };
+    const transformedJson: AchievementContextType = { user: user };
     keys.forEach((key) => {
-        if (key !== 'match' && key !== 'subcourse') {
-            transformedJson[key] = json[key];
-        }
+        transformedJson[key] = json[key];
     });
     return transformedJson;
 }
--- a/common/achievement/util.ts
+++ b/common/achievement/util.ts
@@ -94,7 +94,8 @@ export async function getBucketContext(userID: string, relation?: string): Promi
 }
 
 export function transformPrismaJson(user: User, json: Prisma.JsonValue): AchievementContextType | null {
-    if (!json['match'] && !json['subcourse']) {
+    const keys = Object.keys(json);
+    if (!keys) {
         return null;
     }
     const transformedJson: AchievementContextType = {
@@ -102,6 +103,11 @@ export function transformPrismaJson(user: User, json: Prisma.JsonValue): Achieve
         match: json['match'] ? json['match'] : undefined,
         subcourse: json['subcourse'] ? json['subcourse'] : undefined,
     };
+    keys.forEach((key) => {
+        if (key !== 'match' && key !== 'subcourse') {
+            transformedJson[key] = json[key];
+        }
+    });
     return transformedJson;
 }
 
--- a/common/courses/states.ts
+++ b/common/courses/states.ts
@@ -71,16 +71,18 @@ export async function allowCourse(course: Course, screeningComment: string | nul
     // Usually when a new course is created, instructors also create a proper subcourse with it
     // and then forget to publish it after it was approved. Thus we just publish during approval,
     // assuming the subcourses are ready:
-    const subcourses = await prisma.subcourse.findMany({ where: { courseId: course.id } });
+    const subcourses = await prisma.subcourse.findMany({
+        where: { courseId: course.id },
+        include: { subcourse_instructors_student: { select: { student: true } } },
+    });
     for (const subcourse of subcourses) {
         if (await canPublish(subcourse)) {
             await publishSubcourse(subcourse);
         }
     }
 
-    const subcourse = await prisma.subcourse.findFirst({ where: { courseId: course.id } });
-    const instructors = await prisma.subcourse_instructors_student.findMany({ where: { subcourseId: subcourse.id }, select: { student: true } });
-    instructors.forEach(async (instructor) => {
+    const [subcourse] = subcourses;
+    subcourse.subcourse_instructors_student.forEach(async (instructor) => {
         await Notification.actionTaken(userForStudent(instructor.student), 'instructor_course_approved', {
             courseName: course.name,
             relation: `subcourse/${subcourse.id}`,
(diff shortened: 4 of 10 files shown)


common/achievement/util.ts
```diff
@@ -94,14 +94,20 @@ export async function getBucketContext(userID: string, relation?: string): Promi
 }
 
 export function transformPrismaJson(user: User, json: Prisma.JsonValue): AchievementContextType | null {
-    if (!json['match'] && !json['subcourse']) {
+    const keys = Object.keys(json);
+    if (!keys) {
         return null;
     }
```
dhenkel92, line 100: Do we even need this check?

common/achievement/util.ts
```diff
@@ -94,14 +94,20 @@ export async function getBucketContext(userID: string, relation?: string): Promi
...
     const transformedJson: AchievementContextType = {
         user: user,
         match: json['match'] ? json['match'] : undefined,
         subcourse: json['subcourse'] ? json['subcourse'] : undefined,
     };
+    keys.forEach((key) => {
+        if (key !== 'match' && key !== 'subcourse') {
```
dhenkel92, line 107: Can't we just copy all the keys and remove line 103 + 104? They are basically doing the same thing.

common/courses/states.ts
```diff
@@ -76,6 +77,15 @@ export async function allowCourse(course: Course, screeningComment: string | nul
             await publishSubcourse(subcourse);
         }
     }
+
+    const subcourse = await prisma.subcourse.findFirst({ where: { courseId: course.id } });
```
dhenkel92, line 81: Can't we use the subcourses from line 74?

common/courses/states.ts
```diff
@@ -76,6 +77,15 @@ export async function allowCourse(course: Course, screeningComment: string | nul
...
+    const instructors = await prisma.subcourse_instructors_student.findMany({ where: { subcourseId: subcourse.id }, select: { student: true } });
```
dhenkel92, line 82: This can be added to the subcourse query as `include`

common/courses/states.ts
```diff
@@ -152,6 +162,20 @@ export async function cancelSubcourse(user: User, subcourse: Subcourse) {
...
     logger.info(`Subcourse (${subcourse.id}) was cancelled`);
+
+    const courseInstructors = await prisma.subcourse_instructors_student.findMany({ where: { subcourseId: subcourse.id }, select: { student: true } });
+    await prisma.user_achievement.deleteMany({
+        where: {
+            userId: {
+                in: courseInstructors.map((instructor) => {
+                    const user = userForStudent(instructor.student);
+                    return user.userID;
+                }),
```
dhenkel92, line 173:
```suggestion
                in: courseInstructors.map((instructor) => userForStudent(instructor.student).userID),
```

common/courses/states.ts
```diff
@@ -152,6 +162,20 @@ export async function cancelSubcourse(user: User, subcourse: Subcourse) {
...
+            },
+            group: 'student_offer_course',
```
dhenkel92, line 175: Do we need to use the group selector? Maybe the relation is enough?

graphql/course/mutations.ts
```diff
@@ -106,6 +108,27 @@ export class MutateCourseResolver {
         }
         const result = await prisma.course.update({ data, where: { id: courseId } });
         logger.info(`Course (${result.id}) updated by Student (${context.user.studentId})`);
+
+        const subcourse = await prisma.subcourse.findFirst({ where: { courseId: courseId } });
+        const instructors = await prisma.subcourse_instructors_student.findMany({ where: { subcourseId: subcourse.id }, select: { student: true } });
```
dhenkel92, line 113: This could be done in one query

graphql/course/mutations.ts
```diff
@@ -106,6 +108,27 @@ export class MutateCourseResolver {
...
+        const { context: contextData } = await prisma.user_achievement.findFirst({
+            where: { group: 'student_offer_course', context: { path: ['relation'], equals: `subcourse/${subcourse.id}` } },
```
dhenkel92, line 115: Do we need the group here? Maybe it's enough to filter by relation?

graphql/course/mutations.ts
```diff
@@ -106,6 +108,27 @@ export class MutateCourseResolver {
...
+        contextData['courseName'] = result.name;
+        await prisma.user_achievement.updateMany({
+            where: {
+                userId: {
+                    in: instructors.map((it) => {
+                        const user = userForStudent(it.student);
+                        return user.userID;
+                    }),
+                },
+                group: 'student_offer_course',
```
dhenkel92, line 126: Do we need the group?

graphql/course/mutations.ts
```diff
@@ -193,6 +216,15 @@ export class MutateCourseResolver {
         await hasAccess(context, 'Course', course);
         await prisma.course.update({ data: { courseState: 'submitted' }, where: { id: courseId } });
         logger.info(`Course (${courseId}) submitted by Student (${context.user.studentId})`);
+
+        const subcourse = await prisma.subcourse.findFirst({ where: { courseId: courseId } });
+        const instructors = await prisma.subcourse_instructors_student.findMany({ where: { subcourseId: subcourse.id }, select: { student: true } });
```
dhenkel92, line 221: These could be combined into one query.

graphql/course/mutations.ts
```diff
@@ -106,6 +108,27 @@ export class MutateCourseResolver {
...
         const result = await prisma.course.update({ data, where: { id: courseId } });
         logger.info(`Course (${result.id}) updated by Student (${context.user.studentId})`);
+
+        const subcourse = await prisma.subcourse.findFirst({ where: { courseId: courseId } });
+        const instructors = await prisma.subcourse_instructors_student.findMany({ where: { subcourseId: subcourse.id }, select: { student: true } });
+        const { context: contextData } = await prisma.user_achievement.findFirst({
+            where: { group: 'student_offer_course', context: { path: ['relation'], equals: `subcourse/${subcourse.id}` } },
+        });
+        contextData['courseName'] = result.name;
+        await prisma.user_achievement.updateMany({
```
dhenkel92, line 118: Are we sure that all the user achievements will have the same context? Maybe we should iterate of the list and update each individually.

common/achievement/index.ts
```diff
@@ -129,7 +129,10 @@ async function checkUserAchievement<ID extends ActionID>(userAchievement: UserAc
         const evaluationResultValue =
             typeof evaluationResult.resultObject[dataAggregationKey] === 'number' ? Number(evaluationResult.resultObject[dataAggregationKey]) : null;
         const awardedAchievement = await rewardUser(evaluationResultValue, userAchievement, event);
-        await createAchievement(awardedAchievement.template, userAchievement.userId, event.context);
+        const newAchievement = await createAchievement(awardedAchievement.template, userAchievement.userId, event.context);
+        if (newAchievement) {
+            await checkUserAchievement(newAchievement, event);
```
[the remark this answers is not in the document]
LucasFalkowsky: When we create a new achievement we want to directly evaluate if it was already achieved, so we could generate the subsequent achievement.

This could only loop if we had endless acheivements in a group, which is not the case. the createAchievement function only works, if there are subsequent achievements in our currently checked achievement group.

graphql/subcourse/mutations.ts
```diff
@@ -138,6 +141,16 @@ export class MutateSubcourseResolver {
             await removeParticipantFromCourseChat(instructorUser, subcourse.conversationId);
         }
         logger.info(`Student(${studentId}) was deleted from Subcourse(${subcourseId}) by User(${context.user.userID})`);
+
+        const metrics = ['student_create_course', 'student_submit_course', 'student_approve_course'];
+        await prisma.user_achievement.deleteMany({
+            where: {
+                template: { metrics: { hasSome: metrics } },
```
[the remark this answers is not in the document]
LucasFalkowsky: Oh right, the relation subcourse/{{subcourseId}} is enough to find the user achievements in question.

common/achievement/util.ts
LucasFalkowsky: Yes, if we do not check for existing keys, the forEach loop on keys will throw an error, since forEach() is not a function for null / undefined.

common/achievement/util.ts
LucasFalkowsky: this would then require us to change the AchievementContextType. I set match and subcourse to optional values.

common/courses/states.ts
LucasFalkowsky: I destructured the first element from the subcourses array.

common/courses/states.ts
LucasFalkowsky: good point. We could delete all achievements related to that subcourse.

graphql/course/mutations.ts
LucasFalkowsky: This group selector is necessary. If we were not filtering for the specific group, every achievement that had subcourse/{{id}} as a relation would be added a courseName, even if it had no use for it.

LucasFalkowsky: An other issue could be that the wrong contextData was fetched and data will be saved to the other achievement of an other group, missing essential information, previously existent.

graphql/course/mutations.ts
LucasFalkowsky: yes as previously explained.

graphql/course/mutations.ts
LucasFalkowsky: yes we are sure, because we selected the achievements of the same group. Achievements of the same group always share the same context.

graphql/course/mutations.ts
dhenkel92: > This group selector is necessary. If we were not filtering for the specific group, every achievement that had subcourse/{{id}} as a relation would be added a courseName, even if it had no use for it.

The downside of referencing specific groups is that it might break completely over time if someone decides to rename the group. I guess having extra unnecessary data in the context doesn't hurt.

> An other issue could be that the wrong contextData was fetched and data will be saved to the other achievement of an other group, missing essential information, previously existent.

This could be solved by iterating over the achievements instead of updateMany

dhenkel92: Generally, we should try to not hardcode information that will be set dynamically in the database. The groups are used for grouping and not identification right now.

graphql/course/mutations.ts
dhenkel92: > yes we are sure, because we selected the achievements of the same group. Achievements of the same group always share the same context.

🤔 is this actually true? I thought that the achievement will get the context of the event, which might be different. I think I'm missing something. Can you please elaborate more on where/how this is happening?

common/achievement/index.ts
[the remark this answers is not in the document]
dhenkel92: But can't the context change between these achievements? Every metric has it's own set of actions, which has it's own context.
This means that event A coming from action A will have a different context then event B coming from action B.

graphql/course/mutations.ts
LucasFalkowsky: Alright I fixed this part and removed the group from the filters.

graphql/course/mutations.ts
LucasFalkowsky: I did update this part, because I think you have a point here and I then was not satisfied the way I initially built it.
I'd appreciate you having another look at it.

common/achievement/index.ts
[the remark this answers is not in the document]
LucasFalkowsky: You have a point here. We included a check for a sequential achievement being the last in the group to determine, whether to directly set its achievedAt value to the current Date or not.

common/achievement/types.ts
```diff
@@ -114,6 +114,6 @@ export type ContextSubcourse = {
 
 export type AchievementContextType = {
     user?: User;
-    match: ContextMatch[];
-    subcourse: ContextSubcourse[];
+    match?: ContextMatch[];
+    subcourse?: ContextSubcourse[];
```
dhenkel92, line 118: Why are these optional now? 🤔

dhenkel92 (approved):

PR #953, opened 2024-01-17
feat: resolver for no existing achievements

LomyW:


dhenkel92 (commented):
Couldn't we extend the [user achievements resolver](https://github.com/corona-school/backend/pull/866/files#diff-a91bf026163906d34f760f13d971ec1a547e628ea37d417f3c44e48e14945ac4R244-R249) with a filter instead?

PR #956, opened 2024-01-19
feat: achievement integration tests

LucasFalkowsky:
closes https://github.com/corona-school/project-user/issues/1034

## What was done

1. tests for sequential achievements for students and pupils
2. tests for tiered achievements for students and pupils
3. tests for streak achievements for students and pupils
4. test for all achievements resolver
5. test for further achievements resolver
6. test for next step achievements resolver
7. test for achievement by id resolver


common/achievement/template.ts
```diff
@@ -16,7 +16,7 @@ export enum TemplateSelectEnum {
 const achievementTemplates: Map<TemplateSelectEnum, Map<string, Achievement_template[]>> = new Map();
 
 async function getAchievementTemplates(select: TemplateSelectEnum): Promise<Map<string, Achievement_template[]>> {
-    if (!achievementTemplates.has(select)) {
+    if (!achievementTemplates.has(select) || !achievementTemplates[select]) {
```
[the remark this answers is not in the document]
dhenkel92: Maybe it makes sense to create a `purgeCache` function that can be called from the tests.

integration-tests/15_achievements.ts
```diff
@@ -0,0 +1,1043 @@
...
+
+void test('Reward student onboarding achievement sequence', async () => {
+    await adminClient.request(`mutation ResetRateLimits { _resetRateLimits }`);
+    await createStudentOnboardingTemplates();
+
+    // Verify Email
+    const { student, client } = await createNewStudent();
+    const user = await getUser(student.userID);
+
+    const student_onboarding_1 = await prisma.user_achievement.findFirst({
```
dhenkel92, line 22: Why are all these variables underscored and not camel cased?

integration-tests/15_achievements.ts
```diff
@@ -0,0 +1,1043 @@
...
+            studentTutorScreeningCreate(
+                studentId: ${student.student.id}
+                screening: {success: true comment: "" knowsCoronaSchoolFrom: ""}
+            )
+        }
+    `);
+    const student_onboarding_2 = await prisma.user_achievement.findFirst({
+        where: {
+            group: 'student_onboarding',
+            groupOrder: 3,
```
dhenkel92, line 45: What about the achievement with groupOrder 2?

integration-tests/15_achievements.ts
```diff
@@ -0,0 +1,1043 @@
...
+    const uuid = generateUUID();
+    const match = await prisma.match.create({
+        data: {
+            uuid,
+            source: 'matchedinternal',
+            matchPool: 'lern-fair-now',
+            studentId: student.student.id,
+            pupilId: 1,
+        },
+    });
```
dhenkel92, line 138: You could try to use `CreateManualMatch` here

https://github.com/corona-school/backend/blob/7bb6eed4f8fe3dd1107ac2315898c56a9eaa4117/integration-tests/03_matching.ts#L83-L87

integration-tests/15_achievements.ts
```diff
@@ -0,0 +1,1043 @@
...
+    const { student, client } = await studentOne;
+    const user = await getUser(student.userID);
+
+    const match = await prisma.match.findFirst({
+        where: {
+            studentId: student.student.id,
+            dissolved: false,
+        },
+        select: { id: true },
+    });
```
dhenkel92, line 194: We should try to use as much Graphql queries as possible

integration-tests/15_achievements.ts
```diff
@@ -0,0 +1,1043 @@
...
+
+    const student_match_regular_learning = await prisma.user_achievement.findFirst({
+        where: {
+            userId: user.userID,
+            group: 'student_match_regular_learning',
+            achievedAt: null,
+            recordValue: 2,
+        },
+    });
+    assert.ok(student_match_regular_learning);
```
dhenkel92, line 246: What case is tested here?

integration-tests/15_achievements.ts
```diff
@@ -0,0 +1,1043 @@
...
+    const { pupil, client } = await pupilTwo;
+    const user = await getUser(pupil.userID);
+
+    const match = await prisma.match.findFirst({
+        where: {
+            pupilId: pupil.pupil.id,
+            dissolved: false,
+        },
+        select: { id: true },
+    });
```
dhenkel92, line 262: We should try to use as much Graphql queries as possible

integration-tests/15_achievements.ts
```diff
@@ -0,0 +1,1043 @@
...
+    // request to generate the achievement with initial record value 1
```
dhenkel92, line 263: Maybe we should test if this is true?

integration-tests/15_achievements.ts
```diff
@@ -0,0 +1,1043 @@
...
+    const user = await getUser(student.userID);
+
+    const match = await prisma.match.findFirst({
+        where: {
+            studentId: student.student.id,
+            dissolved: false,
+        },
+        select: { id: true },
+    });
+    // request to generate the achievement with initial record value 1
```
dhenkel92, line 195: Maybe we should test if this is true?

integration-tests/15_achievements.ts
```diff
@@ -0,0 +1,1043 @@
...
+
+    const pupil_match_regular_learning = await prisma.user_achievement.findFirst({
+        where: {
+            userId: user.userID,
+            group: 'pupil_match_regular_learning',
+            achievedAt: null,
+            recordValue: 2,
+        },
+    });
+    assert.ok(pupil_match_regular_learning);
```
dhenkel92, line 314: What case is tested here?

common/achievement/util.ts
```diff
@@ -62,12 +63,12 @@ export async function getBucketContext(userID: string, relation?: string): Promi
...
-            subcourseWhere = { ...subcourseWhere, subcourse_instructors_student: { some: { studentId: id } } };
-        } else {
-            subcourseWhere = { ...subcourseWhere, subcourse_participants_pupil: { some: { pupilId: id } } };
-        }
+        delete whereClause[`${userType}Id`];
+        const userClause =
+            userType === 'student'
+                ? { subcourse_instructors_student: { some: { studentId: id } } }
+                : { subcourse_participants_pupil: { some: { pupilId: id } } };
+        const subcourseWhere = { ...whereClause, ...userClause };
```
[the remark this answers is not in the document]
LucasFalkowsky: I think the upper solution was merged in from another branch, since this was not yet implemented when branching of this one. I'd say though, that I like the new solution slightly better, what do you think?

integration-tests/15_achievements.ts
LucasFalkowsky: this group order is set to isActive = false so we do not create it as next achievement, therefore we have to check for group order 3 not 2.

integration-tests/15_achievements.ts
LucasFalkowsky: It's check, if this achievement is set to achievedAt = null and still hast recordValue 2, when the current streak does not have the highest number of events linked to the achievement. 

- First it is to be seen if all events included in the evaluation generate the record Value 2.
- The first event is deleted and the match meeting join ist executed to trigger the achievement evaluation again.
- It's been checked if the current achievement does not have the highest return value in the evaluation.

integration-tests/15_achievements.ts
LucasFalkowsky: This is the same as previously explained for students

integration-tests/15_achievements.ts
LucasFalkowsky: nope, and I changed it to restore consistency.

common/achievement/util.ts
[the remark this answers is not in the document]
dhenkel92: Makes sense 👍

integration-tests/15_achievements.ts
dhenkel92: Could you add this explanation as a comment?

dhenkel92 (approved):

PR #959, opened 2024-01-21
Fix and amend statistics

realmayus:
- count registrations of helpers/pupils (instead of just tutors/tutees)
- consider both screenings and instructor_screenings in `helperScreenings` endpoint
- `nowFirstMatchesStudent` now considers all students that had their first match in a certain month; not just students that only had a single match which happens to be in that month
- `nowDissolvedMatchesAfter/BeforeThreeMonths` now considers `dissolvedAt` instead of `updatedAt`
- add category controls to `offeredLectures` and `numSubcourses`
- consider start date of first lecture for `numSubcourses`, only consider courses that are allowed
- consider start date of lectures instead of creation date in `offeredLectures`, only consider group type appointments
- split `rateSuccessfulCoCs` into tutor and instructor endpoints
- rate of successful CoCs is now calculated differently: _out of all tutors that must have turned in their CoC (max 8 weeks after screening) within the selected timeframe, how many actually did it?_
- compute trends for dissolved matches (cannot do that in retool as selected timeframe may not align with a specific month)
- `matchesByDuration` now counts the matches that were _created after_ `statistics.from` and _dissolved before_ `statistics.to`
- add `helpersWithoutScreening` metric for determining number of helpers that were invited for a screening
- add `medianTimeToMatch` to provide a better estimate; do this for both students and pupils (used in retool and perhaps in match request form in user app later down the road)

### DIFF
--- a/graphql/statistics/fields.ts
+++ b/graphql/statistics/fields.ts
@@ -48,6 +48,15 @@ class Statistics {
     to: string; // ISO Date String
 }
 
+@ObjectType()
+class MedianTimeToMatch {
+    @Field((type) => Float)
+    median_days_pupil: number;
+
+    @Field((type) => Float)
+    median_days_student: number;
+}
+
 @Resolver((of) => Statistics)
 export class StatisticsResolver {
     @Query((returns) => Statistics)
@@ -282,7 +291,7 @@ export class StatisticsResolver {
                                       WHERE "start" > ${statistics.from}::timestamp
                                         AND "start" < ${statistics.to}::timestamp
                                         AND "appointmentType" = 'group'
-                                        AND course.category = ${category}
+                                        AND course.category = ${category}::course_category_enum
                                       GROUP BY "year", "month"
                                       ORDER BY "year" ASC, "month" ASC;`;
     }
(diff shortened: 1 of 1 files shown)


graphql/statistics/fields.ts
```diff
@@ -721,4 +817,40 @@ export class StatisticsResolver {
...
     }
+
+    @FieldResolver(() => Int)
+    @Authorized(Role.ADMIN)
+    async helpersWithoutScreening(@Root() statistics: Statistics) {
+        return (
+            await prisma.$queryRaw`
+            SELECT COUNT(*)::int AS value
+            FROM student
+            LEFT JOIN instructor_screening ON student.id = instructor_screening."studentId"
```
Jonasdoubleyou, line 828: this also needs to take screenings into account, not just instructor_screenings?

realmayus: Oh this endpoint should actually be called tutorsWithoutMatch, as it's for the tutoring statistics page. Good catch!

Jonasdoubleyou: Before someone was screened, we do not know whether they will be tutors or instructors?

realmayus: The query should query for screenings, not instructor_screenings.

Hmm, is there no way to know how many wannabe-tutors have been invited for screening? 
What about `isStudent=true and (not exists screening where screening.studentId = student.id)`?

Jonasdoubleyou: that will be 0 for now

realmayus: Oh I see. I thought a student called the `meBecomeTutor` endpoint by themselves, but that seem to not be the case in the user app.

Jonasdoubleyou (approved):

PR #960, opened 2024-01-22
fix: update texts and images

LucasFalkowsky:
NOT FINISHED!

closes https://github.com/corona-school/project-user/issues/1082


seed-db.ts
```diff
@@ -1246,13 +1255,14 @@ void (async function setupDevDB() {
             stepName: '',
             type: achievement_type_enum.STREAK,
             subtitle: 'Nachhilfe mit {{matchpartner}}',
-            description: 'Dieser Text muss noch geliefert werden.',
-            image: 'gamification/achievements/tmp/finished_course_sucessfully/finished_course_sucessfully.jpg',
-            achievedImage: 'Hat_gold',
-            actionName: null,
+            description:
+                'Du hast {{num}} Woche(n) in Folge mit {{name}} gelernt! Um diese Serie aufrechtzuerhalten, setze deine gemeinsamen Lernsessions mit {{name}} weiter fort. Regelmäßiges Lernen bringt eine Fülle an Vorteilen mit sich, von verbessertem Wissen und Verständnis bis hin zu gesteigerter Effizienz und Selbstvertrauen. Ihr seid definitiv auf dem richtigen Weg, um eure Ziele zu erreichen!',
```
LomyW, line 1259: `{{ num }}` doesnt exist as variable, right?

seed-db.ts
```diff
@@ -1246,13 +1255,14 @@ void (async function setupDevDB() {
...
+            image: 'gamification/achievements/tmp/streaks/regular_learning_set.png',
+            achievedImage: 'gamification/achievements/tmp/streaks/regular_learning_achieved.png',
+            actionName: 'Noch {{num}} Woche(n) bis zum neuen Rekord!',
```
LomyW, line 1262: also here `{{ num }}` has to be the value, how many weeks are missing to achieve a new record.

LomyW (commented):
Nice! ✨

seed-db.ts
```diff
@@ -757,8 +757,9 @@ void (async function setupDevDB() {
             stepName: 'Verifizieren',
             type: achievement_type_enum.SEQUENTIAL,
             subtitle: 'Jetzt durchstarten',
-            description: 'Dieser Text muss noch geliefert werden.',
-            image: 'gamification/achievements/tmp/finish_onboarding/four_pieces/empty_state.png',
+            description:
+                'Hurra! Am {{date}} haben wir eine E-Mail an deine Adresse {{email}} gesendet. Um deine E-Mail zu bestätigen, klicke einfach auf den Button in der Nachricht. Solltest du unsere E-Mail nicht finden, kannst du hier eine erneute Zustellung anfordern und voller Vorfreude auf unser Weiterkommen warten.',
```
LomyW, line 761: We should also add the date on which the verification message was sent - and the email - to the context. We get this information from `concreteNotification.createdAt`

seed-db.ts
```diff
@@ -865,8 +870,9 @@ void (async function setupDevDB() {
...
```
LomyW, line 874: also here we should the date and the email to the action context.

LomyW (changes requested):
overall looks good to me! 👍🏼

seed-db.ts
LucasFalkowsky: changed it to progress

seed-db.ts
LucasFalkowsky: same as above was done

seed-db.ts
LucasFalkowsky: I used the varifiedAt value inside the function, called by the mutation.

common/certificate-of-conduct/certificateOfConduct.ts
```diff
@@ -18,7 +18,9 @@ export async function create(dateOfInspection: Date, dateOfIssue: Date, criminal
         },
     });
     logger.info(`Certificate of Conduct (${result.id}) created\n`);
-    await Notification.actionTaken(userForStudent(student), 'student_coc_updated', {});
+    await Notification.actionTaken(userForStudent(student), 'student_coc_updated', {
+        date: dateOfIssue.toString(),
```
dhenkel92, line 22:
```suggestion
        dateOfIssue: dateOfIssue.toString(),
```

common/secret/token.ts
```diff
@@ -158,6 +158,13 @@ export async function loginToken(token: string): Promise<User | never> {
 }
 
 export async function verifyEmail(user: User) {
+    const { id: notificationID } = await prisma.notification.findFirst({
+        where: { onActions: { has: 'user-verify-email' } },
```
dhenkel92, line 162: Could use move the string to an constant of type ActionID? It will ensure that there will never be a typo in the string or that we have to update this part when `user-verify-email` is removed.

common/secret/token.ts
```diff
@@ -158,6 +158,13 @@ export async function loginToken(token: string): Promise<User | never> {
...
+        select: { id: true },
+    });
+    const concreteNotification = await prisma.concrete_notification.findFirst({
```
dhenkel92, line 165: What if this is null?

common/secret/token.ts
```diff
@@ -168,7 +175,18 @@ export async function verifyEmail(user: User) {
...
-            await Notification.actionTaken(user, 'student_registration_verified_email', {});
+            if (concreteNotification.sentAt) {
+                await Notification.actionTaken(user, 'student_registration_verified_email', {
+                    date: concreteNotification.sentAt
+                        .toLocaleDateString('de-DE', {
+                            day: 'numeric',
+                            month: 'long',
+                            year: 'numeric',
+                        })
+                        .replace('.', ''),
```
dhenkel92, line 186: Maybe it's better to use moment here? It should be configured properly already, and you have more control over the format.

common/secret/token.ts
LomyW: @dhenkel92 Does it even make sense to look for the concrete notification here? `verifyEmail` creates the concrete notification for `user-verify-email` for the first time.

dhenkel92 (approved):

PR #961, opened 2024-01-22
feat: Exclude existing pupils from requiring a screening

Jonasdoubleyou:
We use matches / course participation to exlude existing active pupils from needing a screening


realmayus (approved):
Out of interest: Why don't we just create successful screenings for these pupils?

Jonasdoubleyou:
@realmayus that was my initial plan, but it also comes with the downside that then we have thousands of "fake screenings", i.e. it would mess up our screening statistics for example, and screeners would be confused by "previous successful screenings". Additionally there might be edge cases where this might actually be desired, i.e. Plus Pupils that are manually matched are then also allowed to access the user app without having been screened

PR #967, opened 2024-01-26
allow cancel/edit a subcourse always when user is an admin

paul-jaworski:


common/courses/states.ts
```diff
@@ -131,15 +133,19 @@ export async function canCancel(subcourse: Subcourse): Promise<Decision> {
...
     if (await subcourseOver(subcourse)) {
         return { allowed: false, reason: 'subcourse-ended' };
     }
 
     return { allowed: true };
 }
 
-export async function cancelSubcourse(user: User, subcourse: Subcourse) {
-    const can = await canCancel(subcourse);
+export async function cancelSubcourse(user: User, subcourse: Subcourse, context: GraphQLContext) {
```
realmayus, line 147: Please only pass an isAdmin boolean to this function, as the functions in common/ are supposed to be as general as possible. The graphql context should be contained in the graphql mutations/resolvers.

realmayus (changes requested):

PR #970, opened 2024-01-28
feature: add dedicated relation field to user achievement table

dhenkel92:
We identify user achievements by the templateID, userID, and relation. Previously, the relation was embedded within the JSON context, making it difficult to use or query. This PR moves the relation to a dedicated field in the user achievement table, allowing for easier querying and the creation of a unique index. This ensures we avoid accidentally creating multiple achievements for the same template and relation.

It should prevent issues like this from our UAT tests, where we've created multiple achievements of the same template for the same user and relation.

image

### DIFF
--- a/common/achievement/get.ts
+++ b/common/achievement/get.ts
@@ -35,8 +35,7 @@ const getNextStepAchievements = async (user: User): Promise<Achievement[]> => {
     });
     const userAchievementGroups: { [groupRelation: string]: achievements_with_template } = {};
     userAchievements.forEach((ua) => {
-        const relation = (ua.context as Prisma.JsonObject)['relation'] || null;
-        const key = relation ? `${ua.template.group}/${relation}` : ua.template.group;
+        const key = ua.relation ? `${ua.template.group}/${ua.relation}` : ua.template.group;
         if (!userAchievementGroups[key]) {
             userAchievementGroups[key] = [];
         }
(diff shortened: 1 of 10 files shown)


common/achievement/get.ts
```diff
@@ -164,14 +167,12 @@ const assembleAchievementData = async (userAchievements: achievements_with_templ
...
         const dataAggregationKeys = Object.keys(currentAchievementTemplate.conditionDataAggregations as Prisma.JsonObject);
-        const currentAchievementContext = userAchievements[currentAchievementIndex].context as Prisma.JsonObject;
-        const relation = currentAchievementContext['relation'] as string;
         const evaluationResult = await evaluateAchievement(
             user.userID,
             currentAchievementTemplate.condition,
             currentAchievementTemplate.conditionDataAggregations as ConditionDataAggregations,
             userAchievements[currentAchievementIndex].recordValue || undefined,
-            relation
+            userAchievements[currentAchievementIndex].relation || undefined
```
Jonasdoubleyou, line 175: Why || undefined?

prisma/migrations/20240128111430_add_relation_as_db_field/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- AlterTable
+ALTER TABLE "user_achievement" ADD COLUMN     "relation" TEXT;
+
+-- CreateIndex
+CREATE UNIQUE INDEX "user_achievement_relation_userId_templateId_key" ON "user_achievement"("relation", "userId", "templateId");
```
Jonasdoubleyou, line 5: What about achievements that can be achieved multiple times?

Jonasdoubleyou (approved):

prisma/migrations/20240128111430_add_relation_as_db_field/migration.sql
Jonasdoubleyou: Or can every achievement only be achieved once per relation?

dhenkel92: Yes. We can have multiple achievements of the same template, but they have to have different relations. Otherwise, we cannot distinguish them afterwards.

common/achievement/get.ts
dhenkel92: It's the difference between TS and DB notion. In typescript we are often using `relation?: string` whereas the DB is using `string | null`...

Jonasdoubleyou: I see, and an empty string should also become undefined?

Jonasdoubleyou: Otherwise I'd prefer ?? undefined

Jonasdoubleyou: Also then some care must be taken with Prisma's update, as null means "set to null" whereas undefined means "keep as is" ...

dhenkel92: 👍 It should never be an empty string. Therefore, I would keep `||` in this case.
Besides that I'll make sure that we cannot save an empty string for the creation.

dhenkel92:
I'll merge the PR to unblock myself. Feel free to continue the discussion 🙂

PR #974, opened 2024-02-02
feat: filter out deactivated chats from the cronjob

LomyW:
# Description
- Closes https://github.com/corona-school/project-user/issues/1093

Already deactivated chats should not be requested, so we added a filter to the talkjs request.

## ✅ What was done
- [x] add filter for deactivated chats 
- [x] reactivated chats have also a finished state...
 
## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [x] Tested on different browsers.

### DIFF
--- a/common/chat/conversation.ts
+++ b/common/chat/conversation.ts
@@ -95,7 +95,7 @@ async function* getAllConversations(onlyActive?: boolean): AsyncIterable<TJConve
     assert(TALKJS_SECRET_KEY, `No TalkJS secret key found to get all conversations.`);
     assureChatFeatureActive();
 
-    const filter = { custom: { finished: '!exists' } };
+    const filter = { custom: { finished: ['!oneOf', ['course_over', 'match_dissolved']] } };
 
     const encodedFilter = encodeURIComponent(JSON.stringify(filter));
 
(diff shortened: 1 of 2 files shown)


Jonasdoubleyou (approved):

common/chat/conversation.ts
```diff
@@ -91,15 +91,21 @@ const getMatcheeConversation = async (matchees: { studentId: number; pupilId: nu
     return { conversation, conversationId };
 };
 
-async function* getAllConversations(): AsyncIterable<TJConversation> {
+async function* getAllConversations(onlyActive?: boolean): AsyncIterable<TJConversation> {
     assert(TALKJS_SECRET_KEY, `No TalkJS secret key found to get all conversations.`);
     assureChatFeatureActive();
 
+    const filter = { custom: { finished: '!exists' } };
```
Jonasdoubleyou, line 98: What if finished: false?

LomyW: We set 
`custom { finished: "course_over" | "match_dissolved"}` if we deactivate chats. So finished can't be false. Or have I misunderstood the question?
https://github.com/corona-school/backend/blob/2f41d00ff9b11814b0bd6e6434122ce4d2caf732/common/chat/deactivation.ts#L58C1-L63C8

But I have forgotten one case: if we reactivate chats. I will add this. @Jonasdoubleyou

Jonasdoubleyou (approved):
nice!

PR #981, opened 2024-02-04
feat: Update Session roles after successful Screening

Jonasdoubleyou:
After a user was screened they can access more things in the User-App, but they do not see it as their roles are cached both in the UserApp and in the backend session. By updating them in the Backend Session, the next time the user refreshes the UserApp they will get the new roles.


common/user/session.ts
```diff
@@ -40,3 +44,16 @@ export const toPublicToken = (token: string) => token.slice(0, -5);
 export async function getUserForSession(sessionToken: string) {
     return await userSessions.get(sessionToken);
 }
+
+// As roles are only evaluated once per session, sometimes it makes sense to update sessions roles in-flight
+// so that clients directly see the new roles once the client updates (i.e. the user app caches it till a refresh happens)
+export async function updateSessionRolesOfUser(userID: string) {
+    // This for sure is O(n) with the number of authenticated users - but as this is rather rare,
+    // I guess there is no need yet to maintain a userID -> session bimap
```
dhenkel92, line 52: 👍 I don't think this is going to be a problem.

dhenkel92 (approved):

PR #990, opened 2024-02-10
refactor: Track Appointment Meeting Join

Jonasdoubleyou:
We tried to abstract away courses and matches with "appointments", so in my eyes it does not make sense to now introduce "match meetings" and "course meetings" if we can also just have "appointment meetings". Also in the future we could do plausibility checks. 

This also removes the need to pass around subcourseId and matchId everywhere in https://github.com/corona-school/user-app/pull/420

### DIFF
--- a/common/appointment/tracking.ts
+++ b/common/appointment/tracking.ts
@@ -1,0 +1,44 @@
+import { lecture as Appointment } from '@prisma/client';
+import { User } from '../user';
+import { prisma } from '../prisma';
+import * as Notification from '../notification';
+import { getLogger } from '../logger/logger';
+
+const logger = getLogger('Appointment Tracking');
+
+// To track achievements we trigger actions whenever a user
+// joins an appointment meeting
+export async function trackUserJoinAppointmentMeeting(user: User, appointment: Appointment) {
+    if (appointment.subcourseId) {
+        const subcourse = await prisma.subcourse.findUniqueOrThrow({ where: { id: appointment.subcourseId }, include: { lecture: true } });
+        if (user.studentId) {
+            const lecturesCount = subcourse.lecture.reduce((acc, lecture) => acc + (lecture.isCanceled ? 0 : 1), 0);
+            await Notification.actionTaken(user, 'student_joined_subcourse_meeting', {
+                relation: `subcourse/${subcourse.id}`,
+                subcourseLecturesCount: lecturesCount.toString(),
+            });
+        } else if (user.pupilId) {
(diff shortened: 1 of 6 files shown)


integration-tests/15_achievements.ts
```diff
@@ -221,12 +216,15 @@ void test('Reward pupil conducted match appointment', async () => {
         }
     `);
     const [match] = matches;
+
+    const dates = createDates();
+    const appointments = await generateLectures(dates, match, student.userID, pupil.userID);
     await client.request(`
```
Jonasdoubleyou, line 222: (funny that previously one could track meeting joins for non existing appointments ...)

common/appointment/tracking.ts
```diff
@@ -0,0 +1,44 @@
...
+            });
+        } else if (user.pupilId) {
+            await Notification.actionTaken(user, 'pupil_joined_match_meeting', {
+                relation: `match/${appointment.matchId}`,
+            });
+        }
+
+        logger.info(`Tracked User(${user.userID}) joining Appointment(${appointment.id}) for a Match(${appointment.matchId})`);
+    } else {
+        logger.info(`Did not track User(${user.userID}) joining Appointment(${appointment.id}) as it is neither associated to a subcourse or match`);
```
dhenkel92, line 42: Is there any other type of appointment in our system? If not, I would change it to error.

common/appointment/tracking.ts
```diff
@@ -0,0 +1,44 @@
...
+// joins an appointment meeting
+export async function trackUserJoinAppointmentMeeting(user: User, appointment: Appointment) {
+    if (appointment.subcourseId) {
+        const subcourse = await prisma.subcourse.findUniqueOrThrow({ where: { id: appointment.subcourseId }, include: { lecture: true } });
+        if (user.studentId) {
+            const lecturesCount = subcourse.lecture.reduce((acc, lecture) => acc + (lecture.isCanceled ? 0 : 1), 0);
+            await Notification.actionTaken(user, 'student_joined_subcourse_meeting', {
+                relation: `subcourse/${subcourse.id}`,
+                subcourseLecturesCount: lecturesCount.toString(),
+            });
```
dhenkel92, line 19: What about adding return statements after the action was tracked?

dhenkel92 (approved):

common/appointment/tracking.ts
Jonasdoubleyou: Not yet, but maybe in the future (i.e. we might also want to do screenings with appointments) - In my eyes "not tracking it" shouldn't fail someone from joining a meeting

PR #992, opened 2024-02-12
feat: assign lecture start to event creation time

LucasFalkowsky:
- closes https://github.com/corona-school/project-user/issues/1147

## What was done:

- refactoring of bucket creation für lecture related buckets
- additional context for meeting join actions
- lecture start as date for joined meeting events

### DIFF
--- a/common/achievement/bucket.ts
+++ b/common/achievement/bucket.ts
@@ -3,39 +3,19 @@ import { BucketFormula, DefaultBucket, GenericBucketConfig, TimeBucket, ContextM
 
 type BucketCreatorDefs = Record<string, BucketFormula>;
 
-enum LectureBucketMeasuringType {
-    start = 'start',
-    participation = 'participation',
-}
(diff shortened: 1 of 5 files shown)


common/achievement/bucket.ts
dhenkel92, line 1: 🤔 This change looks suspicious

dhenkel92 (commented):
We had the same problem for the notification system, why not taking the same solution?

common/achievement/bucket.ts
LucasFalkowsky: what do you mean, suspicious? Filtering is no longer necessary, when setting the event timestamp to lecture.start.

But If you want to keep this, we can.

dhenkel92: I wasn't expecting any changes here 😅 They seemed unrelated and more like merge errors.
Fine for me to remove that part 👍

PR #997, opened 2024-02-14
feature: use template-for to build global relation

dhenkel92:


common/achievement/evaluate.ts
```diff
@@ -25,7 +25,7 @@ async function _evaluateAchievement(
         where: {
             userId,
             metric: { in: metrics },
-            AND: relation ? { relation: { equals: relation } } : {},
+            relation: relation ? { startsWith: relation } : null,
```
dhenkel92, line 28: Not sure if null is right here. Maybe we should take all events, in case we don't have a relation?

PR #998, opened 2024-02-15
fix: adjustments for relations

LucasFalkowsky:
## What was done

- Adjustments to integrate the new relation-system by @dhenkel92 into the image-resolver branch

### DIFF
--- a/common/achievement/bucket.ts
+++ b/common/achievement/bucket.ts
@@ -20,10 +20,7 @@ function createLectureBuckets<T extends ContextMatch | ContextSubcourse>(data: T
             relation: data.relation,
             startTime:
                 measuringType === LectureBucketMeasuringType.start ? moment(lecture.start).subtract(10, 'minutes').toDate() : moment(lecture.start).toDate(),
-            endTime:
-                measuringType === LectureBucketMeasuringType.start
-                    ? moment(lecture.start).add(5, 'minutes').toDate()
-                    : moment(lecture.start).add(lecture.duration, 'minutes').toDate(),
+            endTime: measuringType === LectureBucketMeasuringType.start ? moment(lecture.start).add(5, 'minutes').toDate() : moment(lecture.start).toDate(),
         }));
     return buckets;
 }
--- a/common/achievement/evaluate.ts
+++ b/common/achievement/evaluate.ts
@@ -17,6 +17,7 @@ async function _evaluateAchievement(
     userId: string,
     condition: string,
     dataAggregation: ConditionDataAggregations,
+    templateFor: achievement_template_for_enum,
     recordValue?: number,
     relation?: string
 ): Promise<EvaluationResult | undefined> {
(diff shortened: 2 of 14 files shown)


common/achievement/evaluate.ts
```diff
@@ -25,7 +26,7 @@ async function _evaluateAchievement(
         where: {
             userId,
             metric: { in: metrics },
-            relation: relation ? { startsWith: relation } : null,
+            relation: relation ? { startsWith: relation } : '',
```
LucasFalkowsky, line 29: Relation is not nullable so we check for empty string.

common/achievement/get.ts
```diff
@@ -73,7 +73,7 @@ const getFurtherAchievements = async (user: User): Promise<Achievement[]> => {
         const maxValue = Object.keys(dataAggr)
             .map((key) => {
                 const val = dataAggr[key]['valueToAchieve'] as number;
-                return Number(val);
+                return Number(val || 0);
```
LucasFalkowsky, line 76: from image resolver branch

common/achievement/util.ts
```diff
@@ -67,19 +66,23 @@ export async function getBucketContext(userID: string, relation?: string): Promi
...
     let matches: any[] = [];
-    if (!relationType || relationType === 'match') {
+    if (relation?.includes('match') || !relation) {
         matches = await prisma.match.findMany({
             where: { ...whereClause, [`${userType}Id`]: id },
             select: {
                 id: true,
-                lecture: { where: { NOT: { declinedBy: { hasSome: [`${userType}/${id}`] } } }, select: { start: true, duration: true } },
+                lecture: {
+                    where: { start: { lte: nowPlusFiveMinutes }, NOT: { declinedBy: { hasSome: [`${userType}/${id}`] } } },
```
LucasFalkowsky, line 77: filter to only get events, that already took place, guaranteeing our streaks to be evaluated with the correct dataset.

common/achievement/bucket.ts
```diff
@@ -20,10 +20,7 @@ function createLectureBuckets<T extends ContextMatch | ContextSubcourse>(data: T
             relation: data.relation,
             startTime:
                 measuringType === LectureBucketMeasuringType.start ? moment(lecture.start).subtract(10, 'minutes').toDate() : moment(lecture.start).toDate(),
-            endTime:
-                measuringType === LectureBucketMeasuringType.start
-                    ? moment(lecture.start).add(5, 'minutes').toDate()
-                    : moment(lecture.start).add(lecture.duration, 'minutes').toDate(),
+            endTime: measuringType === LectureBucketMeasuringType.start ? moment(lecture.start).add(5, 'minutes').toDate() : moment(lecture.start).toDate(),
```
dhenkel92, line 23: 🤔

LucasFalkowsky: We removed the duration, in order to prevent further overlapping containers resulting is weird interactions.

This is possible since we agreed on setting the events dates to the lecture start time, when they monitor participation not punctuality.

dhenkel92: Why only for LectureBucketMeasuringType participation?
Because this is the only event where we add this hack?

PR #999, opened 2024-02-16
feature: replace lectureStart in ctx

dhenkel92:


common/achievement/index.ts
```diff
@@ -94,30 +93,15 @@ async function _rewardActionTaken<ID extends ActionID>(user: User, actionId: ID,
...
     if (metricsForEvent.length === 0) {
         logger.warn(`Can't track event, because no metrics found for action`, { actionId: event.actionId });
         return false;
     }
 
-    const lectureStart = event.context['lectureStart'] ? new Date(event.context['lectureStart']) : undefined;
-    const { relation, ...contextWithoutRelation } = event.context;
-
     for (const metric of metricsForEvent) {
-        let templateForMetric: achievement_template | undefined = undefined;
```
LomyW, line 109: just for understanding: why don't we need it anymore? Because we don't save the lecture start in the event context?

dhenkel92: No, this was more like a leftover.

PR #1000, opened 2024-02-16
feat: provided meeting link

LomyW:
# Description
It was discussed that the user has the following options for "creating" a meeting in the front end:
- Zoom Create meeting (default)
- Add your own meeting link
- If the last appointment has an override meeting link, it can be adopted - but can also be overwritten (case 2)
Which link is to be stored in the appointment is therefore handled in the frontend.

- **works with:** https://github.com/corona-school/user-app/pull/474

## ✅ What was done?
- [x] The transferred meeting link is now saved to the lecture


LucasFalkowsky (dismissed):

Jonasdoubleyou (dismissed):
nice, PR 1000 :)

Jonasdoubleyou (approved):

PR #1001, opened 2024-02-16
fix: adjust bucket filter

LucasFalkowsky:
## What was done:

- time frame filters were removed from getBucketContext function
- filter function was introduced to pull out buckets for future lectures not containing elements


common/achievement/util.ts
```diff
@@ -116,6 +115,18 @@ export async function getBucketContext(userID: string, relation?: string): Promi
     return achievementContext;
 }
 
+export function filterBucketEvents(bucketEvents: BucketEvents[]) {
```
dhenkel92, line 118: Can you please add a comment to the function to explain why we are doing this?

LucasFalkowsky: DONE!

dhenkel92 (approved):

PR #1002, opened 2024-02-16
feat: Get achievement templates from prod

Jonasdoubleyou:
resolves https://github.com/corona-school/project-user/issues/1131


Jonasdoubleyou:
(only makes sense to merge once we have some on Prod)

PR #1006, opened 2024-02-23
feature: remove sequential achievement hack

dhenkel92:
This PR is adding a new achievedDescription field to replace the sequential achievement hack.

Solves
https://github.com/corona-school/project-user/issues/1109

### DIFF
--- a/common/achievement/create.ts
+++ b/common/achievement/create.ts
@@ -100,8 +100,6 @@ async function createNextUserAchievement<ID extends ActionID>(
     }
 
     const nextStepTemplate = templatesForGroup[nextStepIndex];
-    const achievedAt =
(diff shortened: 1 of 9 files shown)


Jonasdoubleyou (approved):

Jonasdoubleyou (commented):
I guess we also want to manage it via GraphQL, and ensure that it is only set for the last template if a group?

Jonasdoubleyou:
I guess we also want to set it via GraphQL, ensuring it is only set for the last template in a group?

dhenkel92:
> I guess we also want to set it via GraphQL, ensuring it is only set for the last template in a group (if it is a sequential achievement)?

Very good catch!

Jonasdoubleyou (approved):

PR #1015, opened 2024-02-29
fix: Prohibit mails to unverified accounts, exclude unverified pupils from course promotions

Jonasdoubleyou:
For sure we need to allow certain mails be sent to unverified accounts, especially the mail to verify the account :) 
Also we already start sending some mails directly after the registration, I guess we might want to send them even if the account was not yet verified. Thus this introduces a grace period of 30 days where an unverified account can still get notifications, afterwards we do not send notifications anymore. Also exclude verification & password reset, just in case someone wants to regain access after these 30 days.

### DIFF
--- a/common/courses/notifications.ts
+++ b/common/courses/notifications.ts
@@ -112,7 +112,13 @@ export async function sendPupilCoursePromotion(subcourse: Prisma.subcourse, coun
     }
 
     const pupils = await prisma.pupil.findMany({
-        where: { active: true, isParticipant: true, grade: { in: grades }, subcourse_participants_pupil: { none: { subcourseId: subcourse.id } } },
+        where: {
+            active: true,
+            verification: null,
+            isParticipant: true,
+            grade: { in: grades },
+            subcourse_participants_pupil: { none: { subcourseId: subcourse.id } },
+        },
     });
 
     const courseSubject = course.subject;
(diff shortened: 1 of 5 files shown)


common/courses/notifications.ts
```diff
@@ -112,7 +112,13 @@ export async function sendPupilCoursePromotion(subcourse: Prisma.subcourse, coun
     }
 
     const pupils = await prisma.pupil.findMany({
-        where: { active: true, isParticipant: true, grade: { in: grades }, subcourse_participants_pupil: { none: { subcourseId: subcourse.id } } },
+        where: {
+            active: true,
+            verification: null,
```
dhenkel92, line 117: What does this line do?

integration-tests/13_deactivation.ts
```diff
@@ -226,3 +228,35 @@ void test('Student not should be deactivated as it is not inactive for long enou
 
     assert.strictEqual(dbStudentNew?.active, true);
 });
+
+void test('User should not receive notifications after deactivation', async () => {
```
dhenkel92, line 232: What about creating another test for an old unverified account?

common/courses/notifications.ts
Jonasdoubleyou: Historically, when users registered, we would write a random token into "verification", and then when they verified their email with that token we would NULL it. Later "verifiedAt" was introduced which is the other way round, it is NULL upon registration and then later set. Funnily users could also ignore it, so they could just log in (using a token that was sent to their email) but not verify their email. I somewhen cleaned this up, and now whenever a user logs in (with a token sent to their email) and is not verified yet, we will NULL verified and set verifiedAt. So `verification: null` and `verifiedAt: { not: null }` should be more or less equivalent, though `verification: null` might still be more accurate for historic accounts - we also use it in many other places.

Jonasdoubleyou: So as a quick fix I would bring it in this way, let me create an issue for cleaning up this legacy (we should drop verification, set verifiedAt for all the remaining users, and then consistently use `verifiedAt: { not: null }`

dhenkel92: Thank you for the explanation 👍

PR #1016, opened 2024-03-04
fix: don't create achievement for screening old users

dhenkel92:
We should avoid creating a new achievement if the next step precedes the current step.
This scenario could occur, for instance, if an old student account, created before the implementation of the achievement system, applies to become an instructor.
Consequently, the screening process would trigger the creation of the Onboarding achievement, implying several steps that are inappropriate for the student.
Thus, this PR ensures that only the current or subsequent achievement steps are created, while others are automatically bypassed.


Jonasdoubleyou (approved):

Jonasdoubleyou:
How does this interact with non-sequential achievements?

dhenkel92:
> How does this interact with non-sequential achievements?

It should not change anything for them. Tiered achievements are behaving similar to sequential ones and streaks are only created ones per relation anyways.

PR #1024, opened 2024-03-09
feature: create regular learning achievements

dhenkel92:


common/notification/actions.ts
```diff
@@ -503,6 +503,14 @@ const _notificationActions = {
...
         },
     },
+    student_login: {
+        description: 'Student logged in',
+        sampleContext: {},
+    },
+    pupil_login: {
+        description: 'Pupil logged in',
+        sampleContext: {},
+    },
```
dhenkel92, line 513: I'm not sure if we should create dedicated events for login with PW and login with Token

Jonasdoubleyou: Don't think we need it

Jonasdoubleyou (approved):

PR #1035, opened 2024-03-17
fix: Chat deactivation of matches

Jonasdoubleyou:
Chat metadata points to the wrong match if a user creates a match chat and _their match partner_ has more than one match. Thus if that match is dissolved, all chats of all matches that were wrongfully assigned are deactivated. 

The prisma query got the match partners that had an active match with the current user, but then retrieved *all matches* of that user (the where query does not affect the include), and then takes the id of the first match (whatever that is).

----

Also did some cleanup:

- Drop myConversations - not implemented and not used.
- Move chatSignature to fields/user where it belongs.
- Offer contactOptions as a field resolver on the user entity instead of that global "myContactOptions" query. Thus also Admins can query it for users (useful for debugging)


common/chat/contacts.ts
```diff
@@ -61,32 +61,34 @@ export type MatchContactStudent = {
...
             },
-            include: { match: true },
+            include: { student: true },
         });
 
-        const studentWithMatchId: MatchContactStudent[] = studentsWithMatchId.map((studentWithMatchId) => {
+        const studentWithMatchId: MatchContactStudent[] = matchesWithStudent.map((matchWithStudent) => {
             return {
-                student: studentWithMatchId,
-                matchId: studentWithMatchId.match[0].id,
```
Jonasdoubleyou, line 74: this was the bug

common/chat/deactivation.ts
```diff
@@ -12,8 +12,13 @@ const logger = getLogger('Chat Deactivation');
...
-    const today = moment().endOf('day');
     const match = await prisma.match.findUniqueOrThrow({ where: { id } });
+
+    if (!match.dissolved) {
+        return true;
+    }
+
+    const today = moment().endOf('day');
     const dissolvedAtPlus30Days = moment(match.dissolvedAt).add(30, 'days');
     return !dissolvedAtPlus30Days.isBefore(today);
```
Jonasdoubleyou, line 23: (technically this works as `moment(match.dissolvedAt)` is an Invalid Date for active matches, and then `isBefore` always returns false, so we always return true - But I thought it might be better to actively check for match.dissolved ...)

dhenkel92 (approved):

PR #1037, opened 2024-03-17
fix: update example achievement templating

dhenkel92:


common/courses/states.ts
```diff
@@ -126,7 +126,8 @@ export async function publishSubcourse(subcourse: Prisma.subcourseGetPayload<{ i
     await Promise.all(
         subcourse.subcourse_instructors_student.map((instructor) =>
             Notification.actionTaken(userForStudent(instructor.student), 'instructor_course_approved', {
-                courseName: course.name,
+                course: { name: course.name },
```
Jonasdoubleyou, line 129: This is dangerous - templates might still use `courseName`. Also it is hard to change if we have reminders, as then we have still old contexts around in the DB.

Jonasdoubleyou: Please double check existing Notifications if we have reminders, and prepare the change of the affected templates - But in general I like cleaning this up!

Jonasdoubleyou (approved):

common/courses/states.ts
dhenkel92: Done ✅ 
No notifications are using any of the updated actions.

PR #1038, opened 2024-03-17
fix: filter never achieved streaks

dhenkel92:
This PR prevents something like this

image

Streaks have to be generated in advance for evaluation. Unfotunately, this doesn't automatically mean that we'll reach a first record.
Therefore, we hide streaks that never had a record yet.

### DIFF
--- a/common/achievement/get.ts
+++ b/common/achievement/get.ts
@@ -9,7 +9,14 @@ import { isAchievementConditionMet } from '.';
 
 export async function getUserAchievementsWithTemplates(user: User) {
     const userAchievementsWithTemplates = await prisma.user_achievement.findMany({
-        where: { userId: user.userID, AND: { template: { isActive: true } } },
+        where: {
+            userId: user.userID,
+            template: { isActive: true },
+            // This will ensure that we only get achievements that are either not streaks or have a recordValue of at least 1
+            // Otherwise, we would get all streaks that have not been started yet.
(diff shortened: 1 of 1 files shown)


scripts/test/index.js
```diff
@@ -0,0 +1,7 @@
+const moment = require('moment');
+
+const date = moment().add('5 minutes').toDate();
+console.log(date);
+
+const date2 = moment().add('5', 'm').toDate();
+console.log(date2);
```
Jonasdoubleyou, line 7: ?

scripts/test/package.json
```diff
@@ -0,0 +1,15 @@
+{
```
Jonasdoubleyou, line 1: ?

Jonasdoubleyou (approved):

scripts/test/index.js
dhenkel92: This is why you have reivews....
I've comitted a script that I was using for testing.

Jonasdoubleyou (approved):

scripts/test/index.js
Jonasdoubleyou: nah, thats why you self-review before requesting ;)

dhenkel92: Not if you are in a hurry and don't do it properly 😄 

Therefore we have a second layer of security.

PR #1039, opened 2024-03-18
refactor: achievement evaluation

dhenkel92:
Right now, streaks will be reset to not achieved even tho the evaluation result equals to the record value. (Which is expected in many cases).

So this PR, ensures that we only reset the streak if the evaluation result < record value.

### DIFF
--- a/common/achievement/evaluate.ts
+++ b/common/achievement/evaluate.ts
@@ -1,4 +1,13 @@
-import { BucketConfig, BucketEvents, ConditionDataAggregations, EvaluationResult, GenericBucketConfig, TimeBucket } from './types';
+import {
+    BucketConfig,
+    BucketEvents,
+    ConditionDataAggregations,
+    EvaluationResult,
+    GenericBucketConfig,
+    TimeBucket,
+    AchievementToCheck,
+    AchievementType,
+} from './types';
 import { prisma } from '../prisma';
 import { aggregators } from './aggregator';
 import swan from '@onlabsorg/swan-js';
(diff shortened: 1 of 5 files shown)


integration-tests/15_achievements.ts
dhenkel92, line 1: The tests only worked because of the bug. 😅

common/achievement/evaluate.ts
```diff
@@ -10,17 +19,62 @@ import { achievement_event } from '@prisma/client';
...
+        userId,
+        condition,
+        conditionDataAggregations as ConditionDataAggregations,
+        recordValue,
+        relation,
+        skipAchievementBucketsBefore,
+        skipAchievementBucketsAfter
+    );
+
+    const aggrResult = Object.keys(conditionDataAggregations).reduce((acc, key) => acc + result.resultObject[key], 0);
```
Jonasdoubleyou, line 56: Are we certain that every key of conditionDataAggregations is in resultObject? Maybe add an assertion that checks for NaN?

common/achievement/get.ts
```diff
@@ -156,37 +156,32 @@ const assembleAchievementData = async (userAchievements: achievements_with_templ
...
+        maxValue = userAchievements[currentAchievementIndex].recordValue;
+
+        // Streaks are usually invalidated if a user does not show up for a certain meeting / course / etc.
+        // This means we don't have an event to reevaluate the streak naturally.
+        // There are two ways to solve this:
+        // 1. We could reevaluate the streak every time the user opens the app.
+        // 2. We could create a cronjob that reevaluates all streaks every now and then.
+        // As the cron would be more complex and less efficient, we chose the first option, which is implemented here:
+        if (evaluationResult.shouldInvalidateStreak) {
+            await prisma.user_achievement.update({
```
Jonasdoubleyou, line 171: Maybe restrict this to streak achievements?

Jonasdoubleyou: And maybe add some logging? This looks like it could be the source for bugs in the future?

common/achievement/get.ts
```diff
@@ -156,37 +156,32 @@ const assembleAchievementData = async (userAchievements: achievements_with_templ
...
+                where: { id: userAchievements[currentAchievementIndex].id },
+                data: { achievedAt: null, isSeen: false },
+            });
         }
     }
+    if (userAchievements[currentAchievementIndex].template.type === AchievementType.TIERED) {
+        const evaluationResult = await isAchievementConditionMet(userAchievements[currentAchievementIndex]);
```
Jonasdoubleyou, line 178: Why do we evaluate a second time here?

common/achievement/get.ts
```diff
@@ -156,37 +156,32 @@ const assembleAchievementData = async (userAchievements: achievements_with_templ
...
+        const {
+            template: { conditionDataAggregations },
+        } = userAchievements[currentAchievementIndex];
+        currentValue = evaluationResult.aggregationResult;
+        maxValue = Object.keys(conditionDataAggregations).reduce((acc, key) => acc + conditionDataAggregations[key].valueToAchieve, 0);
```
Jonasdoubleyou, line 183: Previously maxValue was set differently for STREAK / non-STREAK, no it is set for TIERED / non-TIERED? Can you add a comment here why we need to treat TIERED differently?

common/achievement/evaluate.ts
```diff
@@ -10,17 +19,62 @@ import { achievement_event } from '@prisma/client';
...
+    let shouldInvalidateStreak: boolean | null = null;
```
Jonasdoubleyou, line 57: Why boolean | null, if we only consider true/false?

common/achievement/get.ts
```diff
@@ -156,37 +156,32 @@ const assembleAchievementData = async (userAchievements: achievements_with_templ
...
-                currentValue < maxValue &&
-                userAchievements[currentAchievementIndex].achievedAt
-            ) {
-                await prisma.user_achievement.update({
-                    where: { id: userAchievements[currentAchievementIndex].id },
-                    data: { achievedAt: null, isSeen: false },
-                });
-            }
+        currentValue = evaluationResult.aggregationResult;
+        maxValue = userAchievements[currentAchievementIndex].recordValue;
```
Jonasdoubleyou, line 162: What is recordValue for non-TIERED? It's a bit confusing that it is reassigned below.

Jonasdoubleyou (approved):

common/achievement/evaluate.ts
dhenkel92: The idea was to be able to differentiate between "not streak" and "streak that shouldn't be invalidated", as these are the three states.
Not sure if we'll have a use-case at some point tho.

common/achievement/evaluate.ts
dhenkel92: Good Idea 👍

common/achievement/get.ts
dhenkel92: It is restricted to streaks already, as it's wrapped by
```
    if (userAchievements[currentAchievementIndex].template.type === AchievementType.STREAK) {
```

I'll add a log message 👍

common/achievement/get.ts
dhenkel92: Theoretically, parts of the function are a big switch, like:
```
switch(type)
  case Streak:
     evaluate();
     setCurrent;
     setMax;
  case Tiered:
     evaluate();
     setCurrent;
     setMax;
  default:
     setCurrent;
     setMax;
```

But to make it more readable i've implemented it like this:
```
setDefaultCurrent;
setDefaultMax;
if (isStreak)
     evaluate();
     setCurrent;
     setMax;
}
if (isTiered)
     evaluate();
     setCurrent;
     setMax;
}
```

So, it's actually not called a second time, but only for the types we need the re-evaluation.

common/achievement/get.ts
dhenkel92: That's not true, it's actually the same, but I shuffled the code a little bit.

Before:
```
if (isStreak or isTiered) {
  currentValue = evaluationResult;
  maxValue = isStreak ? recordValue : aggregate valueToAchieve;
}
```

Now:
```
if (isStreak) {
  currentValue = evaluationResult;
  maxValue =recordValue
}
if (isTiered) {
  currentValue = evaluationResult;
  maxValue = aggregate valueToAchieve;
}
```

common/achievement/get.ts
dhenkel92: 🤔  The record value is only used for streaks, but the maxValue will be used differently based on the type.

Tiered maxValue => How many events do you need to achieve? Like: 1 / 3 Wochen gelernt.
Streak maxValue => The record value
Sequential maxValue => How many steps do we have in the sequence.

Jonasdoubleyou: Is this documented somewhere?

dhenkel92: I'll start writing the texts when everything is live

PR #1043, opened 2024-03-20
fix: Allow tutors to access zoomMeetingUrl field

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1164

## What was done?

- Allow Tutors to access the `zoomMeetingUrl` field, Replaced Instructor with `Role.OWNER`

I tested it locally before and after, and now users that are tutors _but not instructors_, can access the `zoomMeetingUrl` field, meaning that the appointment detail page works again for them.

### DIFF
--- a/graphql/appointment/fields.ts
+++ b/graphql/appointment/fields.ts
@@ -173,7 +173,7 @@ export class ExtendedFieldsLectureResolver {
         return await getZoomMeeting(appointment);
     }
     @FieldResolver((returns) => String, { nullable: true })
-    @Authorized(Role.ADMIN, Role.APPOINTMENT_PARTICIPANT, Role.INSTRUCTOR, Role.TUTOR)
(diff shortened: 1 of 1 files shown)


graphql/appointment/fields.ts
```diff
@@ -173,7 +173,7 @@ export class ExtendedFieldsLectureResolver {
         return await getZoomMeeting(appointment);
     }
     @FieldResolver((returns) => String, { nullable: true })
-    @Authorized(Role.ADMIN, Role.APPOINTMENT_PARTICIPANT, Role.INSTRUCTOR)
+    @Authorized(Role.ADMIN, Role.APPOINTMENT_PARTICIPANT, Role.INSTRUCTOR, Role.TUTOR)
```
Jonasdoubleyou, line 176: No, this should be Role.OWNER, not Instructor or Tutor

JeangelLF: Nice, thanks for the correction!

I've tested it again and still works fine

Jonasdoubleyou (approved):

PR #1044, opened 2024-03-22
Feat: Add actions for sus screening after registration

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1168

## What was done

- Added the new actions
- Updated the logic to trigger these new actions _(only on their first screening)_

Tested it locally and the new actions are being executed as expected only on the first screening

### DIFF
--- a/common/pupil/screening.ts
+++ b/common/pupil/screening.ts
@@ -59,13 +59,23 @@ export async function updatePupilScreening(screener: Screener, pupilScreeningId:
         return;
     }
 
+    const validScreeningCount = await prisma.pupil_screening.count({ where: { pupilId: screening.pupilId, invalidated: false } });
+    const isFirstScreening = validScreeningCount === 1;
     const asUser = userForPupil(screening.pupil);
     switch (screeningUpdate.status) {
         case PupilScreeningStatus.rejection:
-            await Notification.actionTaken(asUser, 'pupil_screening_rejected', {});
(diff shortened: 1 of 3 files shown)


common/pupil/screening.ts
```diff
@@ -59,13 +59,23 @@ export async function updatePupilScreening(screener: Screener, pupilScreeningId:
         return;
     }
 
+    const validScreeningCount = await prisma.pupil_screening.count({ where: { pupilId: screening.pupilId, invalidated: false } });
```
Jonasdoubleyou, line 62: Also count invalidated, a screening is invalidated if the pupil is matched

common/pupil/screening.ts
```diff
@@ -74,6 +84,11 @@ export async function updatePupilScreening(screener: Screener, pupilScreeningId:
             break;
 
         case PupilScreeningStatus.pending:
+            if (isFirstScreening) {
```
Jonasdoubleyou, line 87: No this is wrong, a screening will also be pending if it is just edited by a screener without taking an action.

Jonasdoubleyou (approved):

common/pupil/screening.ts
Jonasdoubleyou: Cause "missing the screening" is not really a "state", a user can miss the screening multiple times. Also it does not really matter if the user does not show up at all or misses the screening.

PR #1046, opened 2024-03-23
feature: prepare production achievements

dhenkel92:
The idea is that we can copy & execute the `graphql-seed` & `graphql-enable` mutations in our production apollo playground.

The alternative would be writing a script to execute the queries, but I don't see an advantage.


common/achievement/template.ts
```diff
@@ -168,11 +168,11 @@ export async function checkTemplateConsistencyBeforeActivating(template: Achieve
     logger.info(`Checking AchievementTemplate(${template.id}) for consistency`, { template, group });
 
     // ---- Template Metadata -----
-    if (!template.title || !template.description || !template.subtitle) {
+    if (!template.title || !template.description) {
```
dhenkel92, line 171: Subtitle is not required

common/achievement/template.ts
```diff
@@ -168,11 +168,11 @@ export async function checkTemplateConsistencyBeforeActivating(template: Achieve
...
         throw new PrerequisiteError(`AchievementTemplates need a name, description and subtitle`);
     }
 
-    const actionFields = [!!template.actionName, !!template.actionRedirectLink, !!template.actionType];
+    const actionFields = [!!template.actionName, !!template.actionType];
```
dhenkel92, line 175: There might be actions that don't have a link, like waiting for course approval.

common/achievement/template.ts
```diff
@@ -192,7 +192,7 @@ export async function checkTemplateConsistencyBeforeActivating(template: Achieve
             throw new PrerequisiteError(`Aggregation ${name} uses unknown metric ${aggregation.metric}`);
         }
 
-        if ((template.type === AchievementType.TIERED) !== !!aggregation.valueToAchieve) {
+        if ((template.type === AchievementType.TIERED) !== (typeof aggregation.valueToAchieve === 'number')) {
```
dhenkel92, line 195: This was failing if valueToAchieve is 0

graphql/achievement_template/mutations.ts
```diff
@@ -39,8 +38,8 @@ class AchievementTemplateCreateInput implements AchievementTemplateCreate {
 
     @Field()
     condition: string;
-    @Field((type) => JSONResolver)
-    conditionDataAggregations: ConditionDataAggregations;
+    @Field()
+    conditionDataAggregations: string;
```
dhenkel92, line 42: I wasn't able to generate the graphql json scalar via node. Therefore, I switched the input type to string and move the parsing to the resolver function.

Jonasdoubleyou (approved):

PR #1047, opened 2024-03-25
fix: achievement release changes

dhenkel92:
Some small adaption for things that we noticed during our last testing session


common/achievement/metric.ts
```diff
@@ -40,9 +40,13 @@ const batchOfMetrics = [
     createMetric('pupil_onboarding_appointment_booked', ['pupil_calendly_appointment_booked'], () => {
         return 1;
     }),
-    createMetric('pupil_onboarding_screened', ['pupil_screening_appointment_done', 'pupil_screening_add', 'pupil_screening_succeeded'], () => {
-        return 1;
-    }),
+    createMetric(
+        'pupil_onboarding_screened',
+        ['pupil_screening_appointment_done', 'pupil_screening_add', 'pupil_screening_after_registration_succeeded'],
```
dhenkel92, line 45: The action was changed recently.

common/achievement/index.ts
```diff
@@ -168,6 +169,8 @@ async function rewardUser<ID extends ActionID>(evaluationResult: number | null,
 
     const { type, group, groupOrder } = updatedAchievement.template;
 
+    // We want to delay the notification by one minute
+    const triggerNotificationAt = moment().add(1, 'minute').toDate();
```
dhenkel92, line 173: @Jonasdoubleyou Does this work with user-app notifications?

dhenkel92: It seems like it will be delivered during the next `checkReminders` schedule, which is every ten minutes...

Jonasdoubleyou: exactly

PR #1049, opened 2024-03-27
Exclude unscreened pupils from course promotions and campaigns

realmayus:
Exclude unscreened pupils from course promotions and campaigns

Closes https://github.com/corona-school/project-user/issues/1166.

The issue also mentions helpers, but as far as I can tell we already exclude unscreened helpers in campaigns: https://github.com/corona-school/backend/blob/19860a2694b3ad0ade309aac0cd286b87a96ffc5/graphql/user/fields.ts#L203-L207

### DIFF
--- a/common/courses/notifications.ts
+++ b/common/courses/notifications.ts
@@ -119,7 +119,6 @@ export async function sendPupilCoursePromotion(subcourse: Prisma.subcourse, coun
             isParticipant: true,
             pupil_screening: {
                 some: {
-                    invalidated: false,
                     status: 'success',
                 },
             },
--- a/common/courses/notifications.ts
+++ b/common/courses/notifications.ts
@@ -117,6 +117,12 @@ export async function sendPupilCoursePromotion(subcourse: Prisma.subcourse, coun
             active: true,
             verification: null,
             isParticipant: true,
+            pupil_screening: {
+                some: {
+                    invalidated: false,
+                    status: 'success',
+                },
+            },
             grade: { in: grades },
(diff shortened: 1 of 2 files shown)


common/courses/notifications.ts
```diff
@@ -117,6 +117,12 @@ export async function sendPupilCoursePromotion(subcourse: Prisma.subcourse, coun
             active: true,
             verification: null,
             isParticipant: true,
+            pupil_screening: {
+                some: {
+                    invalidated: false,
```
Jonasdoubleyou, line 122: why?

graphql/user/fields.ts
```diff
@@ -175,7 +175,17 @@ export class UserFieldsResolver {
...
             const pupils = await prisma.pupil.findMany({
                 select: { firstname: true, lastname: true, email: true, id: true },
-                where: { ...pupilQuery, active: true, verification: null },
+                where: {
+                    ...pupilQuery,
+                    active: true,
+                    verification: null,
+                    pupil_screening: {
+                        some: {
+                            invalidated: false,
```
Jonasdoubleyou, line 184: why?

common/courses/notifications.ts
realmayus: I assumed that we don't want to send emails to these people either, as they cannot join courses anyways, as their screening is invalidated?

Jonasdoubleyou: they can. Invalidation is only relevant for matching

Jonasdoubleyou (approved):

PR #1056, opened 2024-04-05
Refactor: Add userID to transaction log

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1160

## What was done?

- Added a new `userID` column to the `log` table
- Replace usages of `log.wix_id` with `log.userID`
- `logTransaction` function now requires a `User` as parameter
- Added a [migration](https://github.com/corona-school/dev-db/blob/main/maintenance/FillUserIDColumnOnLogs.pgsql) to fill the new `userID` column

## Note

In this PR my attempt is not to delete the wix_id column yet, as we need it for the migration. After the migration of filling the new `userID` column, I'll follow up with a PR that deletes the `wix_id`, including the TODOs I left to preserve the existing behavior


graphql/pupil/fields.ts
```diff
@@ -100,7 +100,7 @@ export class ExtendFieldsPupilResolver {
     @LimitEstimated(100)
     async logs(@Root() pupil: Required<Pupil>) {
         return await prisma.log.findMany({
-            where: { user: pupil.wix_id },
+            where: { userID: userForPupil(pupil).userID },
```
Jonasdoubleyou, line 103: Don't we have the same field resolver for students? Also while at it, we could maybe move this field resolver to the user entity

Jonasdoubleyou (dismissed):

graphql/pupil/fields.ts
JeangelLF: We don't have it for students, but I'll move it to the user entity 👍

Jonasdoubleyou (approved):

PR #1068, opened 2024-04-23
Fix: delete and recreate zoom user when changing emails

realmayus:
We need to delete and recreate a Zoom user when a Lern-Fair user changes their email address, as we cannot use the changeEmail endpoint because we're not using managed domains on Zoom.


JeangelLF (approved):
Very interesting that we're not storing the zoomUserId anywhere else in our database apart from the student table, nice!

Also interesting to see how disposable are zoom users. Like "worst case scenario" in the `changeEmail` function is that the zoom user gets deleted and not created, which shouldn't cause any issues because we always do `getOrCreateZoomUser` right?

Thanks for the fix!

realmayus:
I think we would run into trouble if the new user wasn't created, as we only do `getOrCreate` when creating meetings.

PR #1074, opened 2024-04-29
fix: Add fallback when fetching zoom user by email does not work

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1192

## What was done?

Added a fallback to fetch users by `zoomUserId` when the email doesn't work. I decided to do this as a fallback because I saw on the zoom forum some reports of the same issue with `zoomUserId`...

### DIFF
--- a/common/zoom/user.ts
+++ b/common/zoom/user.ts
@@ -27,6 +27,12 @@ type ZAKResponse = {
     token: string;
 };
 
+type GetZoomUserArgs = {
+    id: student['id'];
+    email: student['email'];
+    zoomUserId?: student['zoomUserId'];
+};
+
 enum ZoomLicenseType {
     BASIC = 1,
     LICENSED = 2,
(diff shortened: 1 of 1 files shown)


common/zoom/user.ts
```diff
@@ -100,37 +106,55 @@ const createZoomUser = async (student: Pick<student, 'id' | 'firstname' | 'lastn
...
+
+    const byEmailResponse = await zoomFetchUser(email);
+    if (byEmailResponse.status === 404) {
+        logger.info(`Zoom - No Zoom user found for Student(${id}) with email ${email}`);
+        if (!zoomUserId) {
+            return null;
+        }
+
+        /** Let's try with the user id because sometimes the zoom API works with one or the other... */
+        const byIdResponse = await zoomFetchUser(zoomUserId);
```
Jonasdoubleyou, line 136: Can't we just always fetch by userId?

Jonasdoubleyou (approved):

common/zoom/user.ts
JeangelLF: We can, yes. But I'm afraid this will end up in the same issue for other users.
I saw some reports from others on the Zoom support forum that they get 404 for some users when they use the `userId` but not when they use the `email`. So the recommended zoom solution is:
- Can't you find the user by email? Use the id
- Can't you find the user by id? Use the email
🙈

Jonasdoubleyou: Sweet.

PR #1079, opened 2024-05-02
feat/filter for courseStates in SubcourseSearch query

Fabian5150:
Little backend part for https://github.com/corona-school/project-user/issues/975

Allows to only fetch subcourses with specific courseStates with the subcourseSearch-Query.
So that the frontend doesn't have to fetch courses that it doesn't need. 

To be merged before the frontend.

### DIFF
--- a/graphql/subcourse/fields.ts
+++ b/graphql/subcourse/fields.ts
@@ -8,7 +8,7 @@ import { Role } from '../authorizations';
 import { PublicCache } from '../cache';
 import { LimitedQuery, LimitEstimated } from '../complexity';
 import { GraphQLContext } from '../context';
-import { Course, Lecture, Pupil, pupil_schooltype_enum, Subcourse } from '../generated';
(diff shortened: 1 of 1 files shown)


graphql/subcourse/fields.ts
```diff
@@ -137,11 +137,15 @@ export class ExtendedFieldsSubcourseResolver {
     @LimitedQuery()
     async subcourseSearch(
         @Arg('search') search: string,
+        @Arg('courseStates', () => [String], { nullable: true }) courseStates: course_coursestate_enum[],
```
Jonasdoubleyou, line 140: What happens when you omit the courseStates?

Jonasdoubleyou (approved):

graphql/subcourse/fields.ts
Fabian5150: It just fetches all the subcourses as before, ofc I tried that out. :)

PR #1082, opened 2024-05-09
feat: Web Push / Manage Subscriptions

Jonasdoubleyou:
Users can add, get and remove push subscriptions. 

https://github.com/corona-school/project-user/issues/606

### DIFF
--- a/common/notification/channels/push.ts
+++ b/common/notification/channels/push.ts
@@ -1,0 +1,13 @@
+import { getLogger } from '../../logger/logger';
+import { prisma } from '../../prisma';
+import { User } from '../../user';
+
+const logger = getLogger('Push');
+
+export async function removeAllSubcriptions(user: User) {
+    await prisma.push_subscription.deleteMany({
+        where: { userID: user.userID },
+    });
+
+    logger.info(`Removed all Subcriptions of User(${user.userID})`);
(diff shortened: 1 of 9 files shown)


common/notification/channels/push.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export { PushSubscription };
+export type CreatePushSubscription = Omit<PushSubscription, 'id' | 'userID'>;
+
+export async function getPushSubscriptions(user: User) {
+    return await prisma.push_subscription.findMany({
+        where: {
+            userID: user.userID,
+            // Exclude expired subscriptions:
+            OR: [{ expirationTime: { gte: new Date() } }, { expirationTime: null }],
```
JeangelLF, line 20: We should later add a job to remove these, right?

JeangelLF (approved):
I see this as a first step towards Option 1/2 described [here](https://github.com/corona-school/project-user/issues/552) 👀 , which is great!!

common/notification/channels/push.ts
Jonasdoubleyou: right, but I don't think it is that important. Also not sure if browsers actually set this - It's defined by the standard so I implemented it ...

PR #1083, opened 2024-05-11
feature: migrate to webflow api v2

dhenkel92:
https://github.com/corona-school/project-user/issues/1203

This PR updates the webflow sync to use the new V2 API. Previously, with V1, we used a hash to determine if a record had changed. Unfortunately, with the new API structure, this would mean that all records would need to be recreated. Therefore, I've replaced the hash mechanism with a comparator to ensure that we only recreate records if it's really necessary.

Additionally, I've introduced a new dry-run environment variable to allow testing the changes before executing real actions.


Jonasdoubleyou (dismissed):

dhenkel92:
I've added a dry-run flag to see what updates would be applied with the change.

Jonasdoubleyou (approved):

PR #1085, opened 2024-05-16
fix courseState & subcourseSearch filter combination

Fabian5150:
Fixing filter combination of courseState and subcourseSearch to be AND instead of OR


Fabian5150:
Closes https://github.com/corona-school/project-user/issues/1214

realmayus (approved):

PR #1086, opened 2024-05-17
Feat: Allow admin to promote subcourses / Course promotion adjustments

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1207

## What was done?

- Added a new table `subcourse_promotion` to replace the existing `alreadyPromoted` field
- Added missing/new validations to the `canPromoteSubcourse` _(previously `isPromotionValid`)_ function
- Removed the existing MAX_PROMOTIONS emails limit and now we also filter pupils by their email preferences `suggestion.email`
- Added new resolvers for subcourse `canPromoteSubcourse`, `wasPromotedByInstructor`, and `wasPromotedByAdmin`

## Migration

We need to run the following migration to seed the new `subcourse_promotion` table based on our existing fields `published` and `alreadyPromoted`
https://github.com/corona-school/dev-db/commit/a8651977492802dad0542ac8cb6599dffd0d1c9d

After this, I can follow up with a PR to remove the `alreadyPromoted` field from the subcourse table

## Frontend PR

https://github.com/corona-school/user-app/pull/541


common/courses/notifications.ts
```diff
@@ -82,28 +82,117 @@ const getDaysDifference = (date: Date): number => {
     return diffInDays;
 };
 
-const isPromotionValid = (publishedAt: Date, capacity: number, alreadyPromoted: boolean): boolean => {
-    const daysDiff: number | null = publishedAt ? getDaysDifference(publishedAt) : null;
-    return capacity < 0.75 && alreadyPromoted === false && (daysDiff === null || daysDiff > 3);
-};
+interface CanPromoteSubcourseResponse {
```
Jonasdoubleyou, line 85: (we also have the Decision interface which is basically the same thing)

common/courses/notifications.ts
```diff
@@ -82,28 +82,117 @@ const getDaysDifference = (date: Date): number => {
...
+    if (!meetCapacityRequirements) {
+        return { canPromote: false, reason: 'course-capacity-too-high' };
+    }
+
+    // -----------  Subcourse must be published and have been so for at least 3 days (3 days requirement doesn't apply for system promotions) -----------
+    const isSystemPromotion = attemptedPromotionType === Prisma.subcourse_promotion_type_enum.system;
+    const MIN_DAYS_PUBLISHED_FOR_PROMOTIONS = 3;
+    const daysSincePublished = getDaysDifference(publishedAt);
+    const isMatureForPromotion = daysSincePublished > MIN_DAYS_PUBLISHED_FOR_PROMOTIONS || isSystemPromotion;
+    const meetPublishedRequirements = published && isMatureForPromotion;
```
Jonasdoubleyou, line 131: (could move the `published` requirement into a separate reason)

common/courses/notifications.ts
```diff
@@ -148,26 +241,13 @@ export async function sendPupilCoursePromotion(subcourse: Prisma.subcourse, coun
         return !courseSubject || isPupilsSubject;
     });
 
-    const shuffledFilteredPupils = shuffleArray(filteredPupils);
-
-    // get random subset of filtered pupils
-    const seatsLeft = await getCourseFreePlaces(subcourse);
-    const randomPupilSampleSize = Math.min((seatsLeft / PREDICTED_PUPIL_RESPONSE_RATE) * 100, MAX_PROMOTIONS);
-    const randomFilteredPupilSample = shuffledFilteredPupils.slice(0, randomPupilSampleSize);
```
Jonasdoubleyou, line 156: Why is this removed? Taking all pupils potentially notifies thousands of users, we had problems with this in the past

Jonasdoubleyou (approved):

common/courses/notifications.ts
```diff
@@ -136,6 +225,10 @@ export async function sendPupilCoursePromotion(subcourse: Prisma.subcourse, coun
                     },
                 },
             ],
+            notificationPreferences: {
```
Jonasdoubleyou, line 228: What is the default for users that do not have this preference?

JeangelLF: This might also be related to the other comment.

> Why is this removed? Taking all pupils potentially notifies thousands of users, we had problems with this in the past

[Here](https://www.notion.so/lern-fair/Retool-Allow-community-manager-to-promote-courses-manually-c897b7b08cee48318271cab9eb802484?pvs=4#b7e963c8e0d24c05a023d68f85b0a123) is more information about it in the product discussions or @FForstbach could also provide you with more details about this.

But in short, we now want to remove the max limit and at the same time, make some other improvements that should bring a balance, like:
- We moved some of the notification categories that were marked as `system` to `newsletter`, so they should be disabled by default for future users unless they check the `newsletter` checkbox during registration or manually enable the suggestions on their preferences. I tried to tackle that [in this PR](https://github.com/corona-school/backend/pull/1084/files)
- We also recently added a 1-click-opt-out link `/notifications/newsletter?unsubscribe` for newsletters. So it's easier for users to quickly opt out of these emails

> What is the default for users that do not have this preference?

If they don't have this preference set, we shouldn't send them anything, only to those who explicitly agreed to receive these emails. 
I just realized we are already doing this [on the deliverNotification function](https://github.com/corona-school/backend/blob/1096095b09253168864d8cb1cecc9a13610eb35e/common/notification/index.ts#L119). We just have to update the [DEFAULT_PREFERENCES](https://github.com/corona-school/backend/blob/7264c1eb7c000d7b925175fc1545005e8c2506f2/common/notification/defaultPreferences.ts#L45) object which I _(quite embarrassingly)_ missed in my previous PR

However, these measures should heavily impact how many users we reach but also ensure that they're actually interested in these emails.

-------

> What is the default for users that do not have this preference?

Side question, why do we have users who do not have this preference set and depend on the `DEFAULT_PREFERENCES` object? I'm thinking if we should migrate them depending on what they have in the `newsletter` property

Jonasdoubleyou: I would prefer to keep the limit (but increase it if necessary). Just as an additional safeguard in case our assumptions are wrong (or we introduce bugs in the future)

Jonasdoubleyou: With the default preferences we can retrospectively change the defaults without having to update the preferences of all users somehow (potentially overwriting their manually set preference). Since we set them during registration depending on the newsletter field, it lost its value a bit.

JeangelLF: > I would prefer to keep the limit (but increase it if necessary). Just as an additional safeguard in case our assumptions are wrong (or we introduce bugs in the future)

Ok, I brought up the topic again and it was decided that we'll leave the limit behind a feature flag so we can try different things and see what fits us best.

> With the default preferences we can retrospectively change the defaults without having to update the preferences of all users somehow (potentially overwriting their manually set preference). Since we set them during registration depending on the newsletter field, it lost its value a bit.

I see. Then yeah, it would be nice to migrate those users that don't have the `notificationPreferences` to have the default ones _(which I already updated to be opted-out)_

Jonasdoubleyou (approved):

PR #1090, opened 2024-05-23
Feat/derived achievements in important information

dhenkel92:


graphql/user/fields.ts
```diff
@@ -55,6 +57,54 @@ export class Contact {
...
+    @Field((_type) => Int, { nullable: true })
+    finishedSteps: number | null;
+    // @Field((_type) => important_information_recipients_enum)
+    @Field((_type) => String)
+    recipients: 'students' | 'pupils';
+    // @Field((_type) => important_information_language_enum)
+    @Field((_type) => String)
+    language: 'en' | 'de';
+    @Field((_type) => ImportantInformationModal)
+    modal: typeof ImportantInformationModal;
```
dhenkel92, line 81: To be able to show the achievement modal, we have to have all the required information. Therefore, I would introduced a modal union.

common/achievement/derive.ts
```diff
@@ -0,0 +1,156 @@
...
+        templateFor: AchievementTemplateFor.Match,
+        group: 'pupil_new_match',
+        groupOrder: 1,
+        type: AchievementType.SEQUENTIAL,
+        image: 'gamification/achievements/release/finish_onboarding/two_pieces/step_1.png',
+        tagline: 'Starte eine Lernpatenschaft',
+        title: 'Neue Lernunterstützung',
+        subtitle: null,
+        description:
+            'Es war großartig, dich am {{date}} besser kennenzulernen und freuen uns, dass du gemeinsam mit uns die Bildungschancen von Schüler:innen verbessern möchtest. Um dir eine:n passende:n Lernpartner:in zuzuweisen, bitten wir dich zunächst, eine Anfrage auf unserer Plattform zu stellen. Hier kannst du die Fächer und Jahrgangsstufe angeben, die für dich passend sind. Wir freuen uns auf den Start!',
```
Jonasdoubleyou, line 27: Yet another case were it would be good to have translation on the backend :/

graphql/user/fields.ts
Jonasdoubleyou: Can't we return the achievements through the regular achievement resolver? And only the important information through a different one? I would prefer to keep the GraphQL interface simple (and hide all the complexity of the achievement system behind it)

Jonasdoubleyou: (downside is that we potentially evaluate twice, but we could potentially cache in the User Context)

dhenkel92: So you mean that we should remove the Achievement from the ImportantInformationModal and only provide an achievement ID instead? Wouldn't this just move the complexity from the graphql API to the Frontend?

common/achievement/derive.ts
dhenkel92: That's true 😞 
It's actually the case for all the achievements.

Why don't we have any translations on the backend side?

Jonasdoubleyou: Not needed so far ... Maybe time to introduce it

graphql/user/fields.ts
Jonasdoubleyou: I meant

```
me {
  achievements {
    // achievement data
  }

  importantInformations {
    // important information
  }
}
```

Jonasdoubleyou: To avoid the union and keep the concepts separate at the API layer

dhenkel92: But how do we show the achievements within the important information then? The same way typedigital had in mind back then?

https://github.com/corona-school/user-app/pull/546

Jonasdoubleyou: Yes, the UI is different for the two anyways? So dont see that much difference between looping a union and looping two different arrays. 

There might be other places where we want to show just the achievements but not the important information (vice-versa)

common/notification/actions.ts
```diff
@@ -551,12 +551,20 @@ const _notificationActions = {
...
     student_add_appointment_match: {
-        description: 'Tutee / Tutor added Match Appointment',
+        description: 'Tutee / Tutor added Match Appointment (Pupil event)',
         sampleContext: {
             student: sampleUser,
             matchId: '1',
         },
     },
+    student_add_appointment_match_with_pupil: {
+        description: 'Tutee / Tutor added Match Appointment (Tutee / Tutor event)',
```
dhenkel92, line 561: It's the same as the action before. The difference is that `student_add_appointment_match` is called for the pupil and `student_add_appointment_match_with_pupil` for the student.

dhenkel92:
You can test it here

https://user-app-revert-achieve-m8qo7z.herokuapp.com/start

This is the frontend PR

https://github.com/corona-school/user-app/pull/546

common/achievement/derive.ts
```diff
@@ -0,0 +1,302 @@
...
+            isSeen: true,
+            template: groups.get(PupilNewMatchGroup)[0],
+            context: ctx,
+            recordValue: null,
+            achievedAt: null,
+            relation: randomRelation,
+        });
+    }
+
+    console.log('iso', hasRequest, hasSuccessfulScreening, achievement);
```
Jonasdoubleyou, line 154: rm

common/achievement/derive.ts
```diff
@@ -0,0 +1,302 @@
...
+    result.push({
+        id: -1,
+        templateId: -1,
+        userId: user.userID,
+        isSeen: true,
+        template: GhostAchievements.pupil_new_match_1,
+        context: ctx,
+        recordValue: null,
+        achievedAt: new Date(),
+        // achievedAt: hasRequest || achievement ? new Date() : null,
```
Jonasdoubleyou, line 165: the comment looks correct?

common/achievement/derive.ts
```diff
@@ -0,0 +1,302 @@
...
+    return result;
+}
+
+interface PupilNewMatchGhostContext extends Prisma.JsonObject {
+    lastScreeningDate: string | null;
+}
+
+async function derivePupilMatching(user: User, pupil: Pupil, result: achievement_with_template[], userAchievements: achievement_with_template[]) {
+    const hasRequest = pupil.openMatchRequestCount > 0;
+    const successfulScreenings = await prisma.pupil_screening.findMany({
```
Jonasdoubleyou, line 190: .count(...) might be faster.

common/achievement/derive.ts
```diff
@@ -0,0 +1,302 @@
...
+        (row) => row.template.group === PupilNewMatchGroup && row.template.groupOrder === PupilNewMatchGroupOrder
+    );
+
+    const ctx: PupilNewMatchGhostContext = {
+        lastScreeningDate: null,
+    };
+    if (successfulScreenings.length > 0) {
+        ctx.lastScreeningDate = successfulScreenings[0].updatedAt.toISOString();
+    }
+    // This case happens when the student just registered and had a successful screening
```
Jonasdoubleyou, line 207: typo: pupil

common/achievement/derive.ts
```diff
@@ -0,0 +1,302 @@
...
+    }
+}
+
+interface StudentNewMatchGhostContext extends Prisma.JsonObject {
+    lastScreeningDate: string | null;
+}
+
+async function deriveStudentMatching(user: User, student: Student, result: achievement_with_template[], userAchievements: achievement_with_template[]) {
+    const hasRequest = student.openMatchRequestCount > 0;
+    const successfulScreenings = await prisma.screening.findMany({
```
Jonasdoubleyou, line 228: same here

common/achievement/derive.ts
```diff
@@ -0,0 +1,302 @@
...
+}
+
+async function generateStudentMatching(
+    achievement: achievement_with_template | null,
+    user: User,
+    hasRequest: boolean,
+    ctx: StudentNewMatchGhostContext
+): Promise<achievement_with_template[]> {
+    const result: achievement_with_template[] = [];
+    // Generating a ramdom relation to be able to show multiple sequences of this kind in parallel
```
Jonasdoubleyou, line 268: typo: ramdom

common/achievement/get.ts
```diff
@@ -156,6 +162,9 @@ const assembleAchievementData = async (userAchievements: achievements_with_templ
         where: { group: userAchievements[currentAchievementIndex].template.group, isActive: true },
         orderBy: { groupOrder: 'asc' },
     });
+    const derivedTemplates = deriveAchievementTemplates(userAchievements[currentAchievementIndex].template.group);
+    achievementTemplates.push(...derivedTemplates);
+    achievementTemplates.sort((left, right) => left.groupOrder - right.groupOrder);
```
Jonasdoubleyou, line 167: this looks strange - shouldn't it sort by group AND groupOrder? Or why by groupOrder?

common/notification/actions.ts
Jonasdoubleyou: Urgh, naming conventions ... initial plan was to prefix the action id with the user for which the action is triggered, not the user that triggers the action for "student_add_appointment_match" that obviously went wrong ...

common/notification/actions.ts
```diff
@@ -602,6 +610,20 @@ const _notificationActions = {
             appointment: sampleAppointment,
         },
     },
+    pupil_create_new_match_chat: {
+        description: 'User has clicked on new chat with a match partner',
```
Jonasdoubleyou, line 614: Tutee / Created a new chat with a match partner

common/notification/actions.ts
```diff
@@ -602,6 +610,20 @@ const _notificationActions = {
...
+        sampleContext: {
+            user: sampleUser,
+            match: { id: '1' },
+        },
+    },
+    student_create_new_match_chat: {
+        description: 'User has clicked on new chat with a match partner',
```
Jonasdoubleyou, line 621: Tutor / Created a new chat with a match partner

The descriptions show up in the Retool UI, maybe good to distinguish them

Jonasdoubleyou (commented):
Looks good overall!

common/achievement/derive.ts
dhenkel92: Good catch!

common/achievement/get.ts
dhenkel92: Because in this case all achievements in the list are of the same group

common/notification/actions.ts
dhenkel92: Yeah 😞 
Unfortunately, it's not easy to enforce these kind of conventions and it's even harder to change the naming afterwards.

PR #1098, opened 2024-06-09
optional arg for ordering subcourseSearch results descending

Fabian5150:
Added optional argument to the subcourseSearch query to provide a option for 'orderBy'.
This was necessary for the Screening Tool, as it used to fetch only the oldest few courses, which made it a bit useless


graphql/subcourse/fields.ts
```diff
@@ -148,14 +148,16 @@ export class ExtendedFieldsSubcourseResolver {
         @Arg('search') search: string,
         @Arg('courseStates', () => [String], { nullable: true }) courseStates: course_coursestate_enum[],
         @Arg('take', () => GraphQLInt) take: number,
-        @Arg('skip', () => GraphQLInt, { nullable: true }) skip: number = 0
+        @Arg('skip', () => GraphQLInt, { nullable: true }) skip: number = 0,
+        @Arg('orderDesc', { nullable: true }) orderDesc: boolean
```
Jonasdoubleyou, line 152: We could potentially just expose the Prisma OrderBy type here (there should be a generated GraphQL Type for it) - Then we are flexible towards future requirements

Jonasdoubleyou (approved):

graphql/subcourse/fields.ts
Jonasdoubleyou: Also maybe rather orderByLastUpdate or something like that? orderDesc is a bit generic

Jonasdoubleyou: Or maybe orderBy: "last-update", then we can add further sorting variants in the future?

Fabian5150: Went for the third option now, sounds like a good idea.

Jonasdoubleyou (approved):

PR #1099, opened 2024-06-10
feat: Add mutation to send user app feedback email

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1205

## What was done?

- Added a mutation to send emails regarding user-app feedback

[Frontend PR here](https://github.com/corona-school/user-app/pull/564)


common/notification/channels/mailjet.ts
```diff
@@ -163,6 +164,10 @@ export async function sendMail(
         TextPart: text,
     };
 
+    if (attachments?.length) {
+        message.Attachments = attachments;
```
Jonasdoubleyou, line 168: Mails are usually limited to 5MB, this will fail for larger attachments

graphql/user/mutations.ts
```diff
@@ -28,6 +28,29 @@ class SupportMessage {
     subject: string;
 }
 
+@InputType()
+class AppFeedbackAttachment implements Attachment {
+    @Field()
+    Base64Content: string;
```
Jonasdoubleyou, line 34: limit to some reasonable size?

graphql/user/mutations.ts
```diff
@@ -88,6 +111,33 @@ export class MutateUserResolver {
         return true;
     }
 
+    @Mutation((returns) => Boolean)
+    @Authorized(Role.USER)
+    async userSendAppFeedback(@Ctx() context: GraphQLContext, @Arg('message') message: AppFeedback) {
+        const body =
+            `From: ${context.user.email}\n\n` +
```
Jonasdoubleyou, line 118: Von:

Jonasdoubleyou (approved):

graphql/user/mutations.ts
JeangelLF: Good catch.
I've added `2 * 1024 * 1024`... It's not exactly accurate because of the encoding, but I guess for a screenshot image it should be way more than enough.

Jonasdoubleyou (approved):

PR #1105, opened 2024-06-23
log: create diff log for webflow sync

dhenkel92:


jobs/periodic/sync-to-webflow/diff.ts
```diff
@@ -31,6 +32,7 @@ export function diff<T extends WebflowMetadata>(left: T[], right: T[]): { new: T
             // We have to save the old item id, so that it can be used for the update operation
             rightMap[dbId].id = leftMap[dbId].id;
             changedEntries.push(rightMap[dbId]);
+            logger.info('found diff in items', { left: JSON.stringify(left, null, 2), right: JSON.stringify(right, null, 2) });
```
dhenkel92, line 35: This should help to understand what update will be pushed

Jonasdoubleyou (commented):
Looks good

PR #1107, opened 2024-06-28
fix: Catch WebPush errors and remove invalid subscriptions

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1258

## What was done?

- Catch errors from`webPush.sendNotification`
- Remove invalid/expired subscriptions

Tested it locally and expired subscriptions are correctly removed 🚀

@Jonasdoubleyou I'll merge to test it with Fredrik on STG :) Let me know if there is something we could improve here.

### DIFF
--- a/common/notification/channels/push.ts
+++ b/common/notification/channels/push.ts
@@ -123,19 +123,27 @@ export const webpushChannel: Channel = {
 
         for (const subscription of subscriptions) {
             logger.info(`Sending ConcreteNotification(${concreteID}) to Subscription(${subscription.id}) of User(${to.userID})`);
-            const result = await WebPush.sendNotification(
-                { endpoint: subscription.endpoint, keys: subscription.keys as any },
-                JSON.stringify({
-                    // As we use the keys of the client to encrypt the message, we can include
-                    // more details here, as only the user's device will be able to decrypt it
-                    concreteNotificationId: concreteID,
-                    message,
-                })
-            );
+            try {
+                const result = await WebPush.sendNotification(
+                    { endpoint: subscription.endpoint, keys: subscription.keys as any },
+                    JSON.stringify({
+                        // As we use the keys of the client to encrypt the message, we can include
+                        // more details here, as only the user's device will be able to decrypt it
+                        concreteNotificationId: concreteID,
+                        message,
+                    })
+                );
 
-            logger.info(`Sent ConcreteNotification(${concreteID}) to Subscription(${subscription.id}) of User(${to.userID})`, {
-                result,
-            });
+                logger.info(`Sent ConcreteNotification(${concreteID}) to Subscription(${subscription.id}) of User(${to.userID})`, {
+                    result,
+                });
+            } catch (error) {
+                const webPushError = error as WebPush.WebPushError;
+                // Subscription has expired or is no longer valid
+                if ([404, 410].includes(webPushError.statusCode)) {
+                    removePushSubscription(to, subscription.id).catch((e) => {});
+                }
+            }
         }
     },
 };


common/notification/channels/push.ts
```diff
@@ -123,19 +123,27 @@ export const webpushChannel: Channel = {
...
+
+                logger.info(`Sent ConcreteNotification(${concreteID}) to Subscription(${subscription.id}) of User(${to.userID})`, {
+                    result,
+                });
+            } catch (error) {
+                const webPushError = error as WebPush.WebPushError;
+                // Subscription has expired or is no longer valid
+                if ([404, 410].includes(webPushError.statusCode)) {
+                    removePushSubscription(to, subscription.id).catch((e) => {});
+                }
```
Jonasdoubleyou, line 145: else rethrow!

common/notification/channels/push.ts
```diff
@@ -123,19 +123,27 @@ export const webpushChannel: Channel = {
...
+                );
+
+                logger.info(`Sent ConcreteNotification(${concreteID}) to Subscription(${subscription.id}) of User(${to.userID})`, {
+                    result,
+                });
+            } catch (error) {
+                const webPushError = error as WebPush.WebPushError;
+                // Subscription has expired or is no longer valid
+                if ([404, 410].includes(webPushError.statusCode)) {
+                    removePushSubscription(to, subscription.id).catch((e) => {});
```
Jonasdoubleyou, line 144: await!

Jonasdoubleyou: DO NOT IGNORE ERRORS!

Jonasdoubleyou: Logging?

JeangelLF: ### RE: Caps... I'm always all in to talk about it if I made a mistake or wrong assumptions. It's not necessary to "shout"
----------
>await! 

It was actually on purpose. We don't really need to wait for that. I can add it though.

> DO NOT IGNORE ERRORS!

It was partially on purpose. If we throw an error for a single invalid subscription the other subscriptions won't receive the notification. That is _exactly_ the problem I'm trying to fix.

> else rethrow!

Ditto here.

> Logging?

Yep, I missed it. Actually, I think this should be the right thing to do here instead of throwing errors everywhere


I'll follow up with a PR adding logging. Please let me know if my thoughts are wrong and we should do this differently.

Jonasdoubleyou: Yes, error logging is also fine. We should just not ignore errors, these will be bugs in the future and the more we silently ignore, the harder it is to debug. 

Regarding await - If you do not await, it could hang forever (e.g. in a deadlock) and "wont just do anything". By awaiting here the request will also hang, and we will notice (as the client times out + produces an error). Omitting an await should be done in exceptional cases.

Sorry for "shouting", just read through this in a hurry and saw it was already merged so wanted to raise attention to this before it goes to production. Wasn't meant as shouting.

JeangelLF: > Yes, error logging is also fine. We should just not ignore errors, these will be bugs in the future and the more we silently ignore, the harder it is to debug.
Regarding await - If you do not await, it could hang forever (e.g. in a deadlock) and "wont just do anything". By awaiting here the request will also hang, and we will notice (as the client times out + produces an error). Omitting an await should be done in exceptional cases.

Got it, thanks for the clarification! 🙌 I also added the await in the other PR, so I think this one is good to go.


>Sorry for "shouting", just read through this in a hurry and saw it was already merged so wanted to raise attention to this before it goes to production. Wasn't meant as shouting.

Alrighty, sorry for the misunderstanding :)

PR #1109, opened 2024-06-29
refactor: Drop verification column

paul-jaworski:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1153

## What was done?

- Replaced usage of `verification` with `verifiedAt`
- Drop `verification` columns from pupil/student/screener tables
- In a separate PR we already made sure that all of our users have the `verifiedAt` set, this was done in https://github.com/corona-school/dev-db/pull/1


common/courses/notifications.ts
```diff
@@ -11,6 +11,7 @@ import { NotificationContext } from '../notification/types';
 import moment from 'moment';
 import { Decision } from '../util/decision';
 import { shuffleArray } from '../util/basic';
+import { notNull } from 'jest-mock-extended';
```
JeangelLF, line 14: This import is not needed, we can remove it.

common/notification/index.ts
```diff
@@ -600,22 +600,6 @@ export async function _actionTakenAt<ID extends ActionID>(
             }
         }
 
-        // --------------- Send out Notifications to send directly --------------------------------------
```
JeangelLF, line 603: Why did you remove this whole block? 🤔  Let's bring it back and fix the parts that use the `verification` field.

prisma/migrations/0_init/migration.sql
```diff
@@ -347,7 +347,6 @@ CREATE TABLE "mentor" (
     "lastname" VARCHAR,
     "active" BOOLEAN NOT NULL DEFAULT true,
     "email" VARCHAR NOT NULL,
-    "verification" VARCHAR,
```
JeangelLF, line 350: Undo the changes to this file. It doesn't make sense to do this on the migration files.

prisma/schema.prisma
```diff
@@ -425,7 +425,6 @@ model mentor {
   lastname                     String?                 @db.VarChar
```
JeangelLF, line 425: Also, undo the changes to this file. The idea is that we first stop using the verification column in favor of `verifiedAt`.
The part of dropping the verification column from the DB should be done in a separate pull request :)

common/pupil/registration.ts
```diff
@@ -90,6 +90,7 @@ export async function registerPupil(data: RegisterPupilData, noEmail = false, pr
             isPupil: true,
 
             // the authToken is used to verify the e-mail instead
+            // PR Note: How do we drop this?
             verification,
```
JeangelLF, line 94: No need to drop this for now, we can do it in a separate PR (the same that drops the column from the `schema.prisma`)

common/pupil/registration.ts
```diff
@@ -103,6 +104,7 @@ export async function registerPupil(data: RegisterPupilData, noEmail = false, pr
         await Notification.actionTaken(userForPupil(pupil), 'pupil_registration_started', { redirectTo: data.redirectTo ?? '' });
     }
 
+    //PR Note: How do we drop this?
     await logTransaction('verificationRequets', userForPupil(pupil), {});
```
JeangelLF, line 108: No need to drop this.

common/pupil/token.ts
```diff
@@ -7,6 +7,7 @@ import { userForPupil } from '../user';
 const logger = getLogger('Pupil Token Auth');
 
 export async function refreshToken(pupil: pupil): Promise<string | never> {
+    //PR Note: Do we have to drop this?
     if (pupil.verification !== null) {
```
JeangelLF, line 11: We have to change it to check if `verifiedAt` is not set

common/student/registration.ts
```diff
@@ -79,6 +79,7 @@ export async function registerStudent(data: RegisterStudentData, noEmail = false
             wix_creation_date: new Date(),
 
             // the authToken is used to verify the e-mail instead
+            //PR Note: Do we have to drop this?
             verification: uuidv4(),
```
JeangelLF, line 83: Not yet.

JeangelLF (changes requested):
Thanks for the PR.

I left some comments about some changes that need to be made.

common/pupil/registration.ts
JeangelLF: Could you remove the comment or change it to `TODO: Remove after migration`?
You can leave PR Notes as a comments here in Github.

common/pupil/registration.ts
JeangelLF: Could you remove the comment or change it to `TODO: Remove after migration`?

common/pupil/token.ts
JeangelLF: We still need to change this validation to check if verified is not set @paul-jaworski
Also, remove the comment

common/student/registration.ts
JeangelLF: Could you remove the comment or change it to `TODO: Remove after migration`?

prisma/migrations/0_init/migration.sql
JeangelLF: There are still some order/formatting changes. Could you also fix those?
Ideally, we leave this file untouched

prisma/schema.prisma
JeangelLF: Here there are also changes with the ordering / formatting. Could you also fix those?

JeangelLF (changes requested):
Thanks, most of it was already fixed. I left again some comments about some adjustments we have to make.

Here on Github, on the `Files changed` tab, you can check by yourself the final changes of the pull request. Make sure that all of them are intentional :)

Jonasdoubleyou (approved):

PR #1112, opened 2024-07-02
Feat: Add disabled channels for notifications

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1263

## What was done?

- Added a new field `disabledNotifications` on the `notification` model
- Added validation to check for disabledChannels

### DIFF
--- a/common/notification/index.ts
+++ b/common/notification/index.ts
@@ -121,7 +121,9 @@ async function deliverNotification(
 
         activeChannels = [];
         for (const enabled of enabledChannels) {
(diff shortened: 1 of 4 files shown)


Jonasdoubleyou (dismissed):
nice! Maybe worth adding an integration test

JeangelLF:
Thanks for the recommendation! I've added an integration test similar to the one that checks the user preferences to send the notification via a channel _(email in this case)_.

PR #1113, opened 2024-07-04
Fix: Filter out non relevant notifications

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1267

## What was done?

- Added notification - concrete_notification relationship 
- Added filters to skip non-relevant notifications for the user app

### DIFF
--- a/graphql/user/fields.ts
+++ b/graphql/user/fields.ts
@@ -9,6 +9,7 @@ import {
     PupilWhereInput,
     Log,
     Push_subscription as PushSubscription,
+    notification_channel_enum as NotificationChannelEnum,
 } from '../generated';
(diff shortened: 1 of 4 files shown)


graphql/user/fields.ts
```diff
@@ -136,13 +137,27 @@ export class UserFieldsResolver {
     @Authorized(Role.OWNER, Role.ADMIN)
     @LimitedQuery()
     async concreteNotifications(
```
Jonasdoubleyou, line 139: Maybe there are other scenarios in which one wants alls sent notifications (e.g. for a GDPR data export or to show notifications sent via email etc.). So maybe it makes sense to filter through a parameter. But YAGNI ...

Jonasdoubleyou (approved):

PR #1114, opened 2024-07-10
Feat: Increase pupils grade mutation

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/574

## What was done?

- Add mutation `adminIncreasePupilGrades`
- Execute action when increasing pupil grade via this mutation

### DIFF
--- a/graphql/pupil/mutations.ts
+++ b/graphql/pupil/mutations.ts
@@ -375,22 +375,30 @@ export class MutatePupilResolver {
         });
         logger.info(`Attempting to increase the grade of ${pupils.length} pupils`);
         let validCount = 0;
-        for (const pupil of pupils) {
-            const grade = gradeAsInt(pupil.grade);
-            try {
-                await prisma.pupil.update({
-                    where: { id: pupil.id },
-                    data: {
-                        grade: `${grade + 1}. Klasse`,
-                        gradeUpdatedAt: new Date(),
-                    },
-                });
-                await Notification.actionTaken(userForPupil(pupil), 'pupil_grade_increased', {});
-                validCount += 1;
-            } catch (error) {
-                logger.error(`Error increasing Pupil(${pupil.id}) grade`, error);
-            }
+        const batchSize = 1000;
+        for (let i = 0; i < pupils.length; i += batchSize) {
+            const batch = pupils.slice(i, i + batchSize);
+
+            await Promise.all(
+                batch.map(async (pupil) => {
+                    try {
+                        const grade = gradeAsInt(pupil.grade);
+                        await prisma.pupil.update({
+                            where: { id: pupil.id },
+                            data: {
+                                grade: `${grade + 1}. Klasse`,
+                                gradeUpdatedAt: new Date(),
+                            },
+                        });
+                        await Notification.actionTaken(userForPupil(pupil), 'pupil_grade_increased', {});
+                        validCount += 1;
+                    } catch (error) {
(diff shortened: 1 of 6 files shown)


graphql/pupil/mutations.ts
```diff
@@ -361,4 +363,43 @@ export class MutatePupilResolver {
...
+    async adminIncreasePupilGrades() {
+        const pupils = await prisma.pupil.findMany({
+            where: {
+                grade: { not: null, notIn: ['13. Klasse', '14. Klasse'] },
+                OR: [{ gradeUpdatedAt: { lt: moment().subtract(10, 'months').toDate() } }, { gradeUpdatedAt: { equals: null } }],
+            },
+        });
+        logger.info(`Attempting to increase the grade of ${pupils.length} pupils`);
+        let validCount = 0;
+        const batchSize = 1000;
```
JeangelLF, line 378: I tested this with 20k records and it took ~20 seconds, so I guess that's fine for a one-time execution per year?

graphql/pupil/mutations.ts
```diff
@@ -361,4 +363,43 @@ export class MutatePupilResolver {
...
+
+            await Promise.all(
+                batch.map(async (pupil) => {
+                    try {
+                        const grade = gradeAsInt(pupil.grade);
+                        await prisma.pupil.update({
+                            where: { id: pupil.id },
+                            data: {
+                                grade: `${grade + 1}. Klasse`,
+                                gradeUpdatedAt: new Date(),
```
JeangelLF, line 390: I decided to set only this value here since it seems we want to update everyone's grade by 1 no matter if the user recently updated it... which is a bit weird.. But in general, all this grade stuff is a bit weird.

>Increment grade on August 1 by 1 - no matter when the grade was last updated

https://www.notion.so/lern-fair/Automatische-Anpassung-der-Jahrgangsstufe-ee4c95a8b8564cb08841aa8ec6d319d9?pvs=4#b1c9f041e09846bf81528742238075fa

Jonasdoubleyou: I also find it weird and also would prefer to respect it if a user updated it recently - e.g. if they register during the summer break and already take their "new class". Maybe double check with Fredrik?

graphql/pupil/mutations.ts
```diff
@@ -361,4 +363,43 @@ export class MutatePupilResolver {
...
+        let validCount = 0;
+        const batchSize = 1000;
+        for (let i = 0; i < pupils.length; i += batchSize) {
+            const batch = pupils.slice(i, i + batchSize);
+
+            await Promise.all(
+                batch.map(async (pupil) => {
+                    try {
+                        const grade = gradeAsInt(pupil.grade);
+                        await prisma.pupil.update({
```
Jonasdoubleyou, line 386: Concurrent updates on one table are dangerous and unnecessarily causes lock collisions on the db (in the end it will be sequential) - We had problems in the past were Prisma was giving up on many concurrent transactions and started deadlocking. Can't we use updateMany by grouping by grade? That should be by magnitudes better, as it only requires 13 transactions (one for each grade).

So

for grade in grades:
   get all pupils with grade
   update all pupils with grade
   for pupil in pupils
      take action (Or use the bulk variant?)

JeangelLF: Alright, I've updated the approach. Thanks for the suggestions!

Jonasdoubleyou (approved):
nice

PR #1117, opened 2024-07-13
fix: If pupils leave a course, add other pupils from the waiting list & Avoid annoying course instructors with notifications

Jonasdoubleyou:
- If a pupil leaves a course, the course is joinable and there are pupils on the waiting list, add pupils from the waiting list
- If a course is filled by the system, do not send a notification to the instructor (either the instructor themself requested it, support did it on behalf of the instructor, or we fill the subcourse after a pupil left the subcourse, so it was full and now is full again)

resolves https://github.com/corona-school/project-user/issues/1271

### DIFF
--- a/common/courses/participants.ts
+++ b/common/courses/participants.ts
@@ -146,23 +146,33 @@ export async function canJoinSubcourse(subcourse: Subcourse, pupil: Pupil): Prom
     return { allowed: true };
 }
 
+async function subcourseJoinable(subcourse: Subcourse): Promise<Decision<CourseDecision>> {
+    const firstLecture = await prisma.lecture.findMany({
+        where: { subcourseId: subcourse.id },
+        orderBy: { start: 'asc' },
+        take: 1,
+    });
+
+    if (firstLecture.length !== 1) {
+        return { allowed: false, reason: 'no-lectures' };
+    }
+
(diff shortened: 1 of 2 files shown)


common/courses/participants.ts
```diff
@@ -146,23 +148,41 @@ export async function canJoinSubcourse(subcourse: Subcourse, pupil: Pupil): Prom
     return { allowed: true };
 }
 
-// Whether the pupil could join the course if the course was not full (i.e. pupils can still join the waitinglist)
-export async function couldJoinSubcourse(subcourse: Subcourse, pupil: Pupil): Promise<Decision<CourseDecision>> {
-    if (!canJoinSubcourses(pupil).allowed) {
-        return canJoinSubcourses(pupil);
-    }
-
+// Whether a subcourse is in principle joinable at this point in time, not considering the number of participants
```
realmayus, line 151: What's the difference to couldJoinSubcourse? According to comments, both seem to check if a pupil can join the subcourse regardless of whether there are still free course places.

realmayus (approved):

common/courses/participants.ts
Jonasdoubleyou: Yeah we always have a permutation of some of the checks and some are skipped. Could potentially be refactored at some point.

PR #1119, opened 2024-07-14
fix: use live endpoint to remove webflow collection item

dhenkel92:
This PR should fix the issue of old courses on the website. It will ensure that deleted collection items will be unpublished automatically.


dhenkel92:
https://lern-fair.slack.com/archives/C03E45MES4X/p1720454783390729

Jonasdoubleyou (approved):

PR #1122, opened 2024-07-19
Feat: School during registration

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1261

## What was done?

- Repurposed our existing `School` model to save the information of SuS during registration optionally
- Added optional `school` field in the `meRegisterPupil` mutation input
- Added machinery to search external schools, using jedeschule API


graphql/external_school/fields.ts
```diff
@@ -0,0 +1,43 @@
...
+class ExternalSchoolSearchOptions {
+    @Field((type) => Int, { defaultValue: 20 })
+    limit: number;
+}
+
+@Resolver((of) => ExternalSchoolSearch)
+export class ExternalSchoolResolver {
+    @Query((returns) => [ExternalSchoolSearch])
+    @Authorized(Role.UNAUTHENTICATED)
+    externalSchoolSearch(@Arg('filters') filters: ExternalSchoolSearchFilters, @Arg('options') options: ExternalSchoolSearchOptions) {
```
JeangelLF, line 40: I think it would make sense to rate-limit this one. Not sure what limit though... I'll do some tests / research and see what fits best

graphql/external_school/fields.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+@InputType()
+class ExternalSchoolSearchOptions {
+    @Field((type) => Int, { defaultValue: 20 })
+    limit: number;
+}
+
+@Resolver((of) => ExternalSchoolSearch)
+export class ExternalSchoolResolver {
```
JeangelLF, line 37: Not really sure about the name... Probably this should live under `schools` as a query 🤔

common/school/externalSearch.ts
```diff
@@ -0,0 +1,61 @@
...
+export interface SchoolResult extends Omit<ExternalSchool, 'school_type'> {
+    schooltype?: SchoolType;
+    state: State;
+}
+
+const logger = getLogger('Jedeschule');
+
+const getStateFromExternalSchool = (school?: ExternalSchool) => {
+    const state = school?.id.substring(0, 2).toLowerCase();
+    const isValid = Object.values(State).some((e) => e === state);
```
Jonasdoubleyou, line 23: (.includes(...))

common/school/externalSearch.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (!isValid) {
+        /**
+         * The state is always the first part of the school id, if for some reason the state doesn't match
+         * we need the heads up about that.
+         */
+        const msg = 'Could not get state from school';
+        logger.error(msg, new Error(msg), school);
```
Jonasdoubleyou, line 30: (not so much a fan of non-error handling ... why not just throw an error here?)

common/school/externalSearch.ts
```diff
@@ -0,0 +1,61 @@
...
+interface SearchSchoolsArgs {
+    filters: { name: string };
+    options: { limit: number };
+}
+
+export const searchExternalSchools = async (params: SearchSchoolsArgs): Promise<SchoolResult[]> => {
+    const { filters, options } = params;
+    try {
+        const response = await fetch(`https://jedeschule.codefor.de/schools/?limit=${Math.min(options.limit, 50)}&include_raw=false&name=${filters.name}`);
+        if (response.ok) {
```
Jonasdoubleyou, line 53: (same here - do we expect this to fail, and if yes, is "not found" the right behaviour?)

prisma/migrations/20240719095833_new_school_structure/migration.sql
```diff
@@ -0,0 +1,28 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `activeCooperation` on the `school` table. All the data in the column will be lost.
+  - You are about to drop the column `emailDomain` on the `school` table. All the data in the column will be lost.
+  - You are about to drop the column `website` on the `school` table. All the data in the column will be lost.
+  - A unique constraint covering the columns `[name,state,city]` on the table `school` will be added. If there are existing duplicate values, this will fail.
+  - Made the column `state` on table `school` required. This step will fail if there are existing NULL values in that column.
```
Jonasdoubleyou, line 8: Do we have entries in the schools table?

integration-tests/01_user.ts
```diff
@@ -37,8 +37,10 @@ export async function createNewPupil() {
                 lastname: "lastname:${userRandom}"
                 email: "test+${userRandom}@lern-fair.de"
                 newsletter: true
-                state: bw
                 registrationSource: normal
+                school: {
+                  state: bw
```
Jonasdoubleyou, line 42: What does just passing the state here do? It will not find a school and then do nothing? Or will it create a school for each pupil?

Jonasdoubleyou (dismissed):

common/school/externalSearch.ts
JeangelLF: I'm handling the potential "error" with the validation. I don't want the whole endpoint to fail due to a single corrupted result. The Frontend can still work with the other results (and even with this one by not skipping the state step).

common/school/externalSearch.ts
JeangelLF: Here it makes sense to throw an error, I've just added it

prisma/migrations/20240719095833_new_school_structure/migration.sql
JeangelLF: We have around ~20 incomplete entries from 2020/2021.
I'll confirm if we still need any of the info from that table

integration-tests/01_user.ts
JeangelLF: This is the equivalent of having this new feature disabled OR users refusing to provide school information during registration, which is fine because it's optional.

So, In this case, we just fill the state field on the [pupil](https://github.com/corona-school/backend/blob/fd4c2578c219bd9721646a3e4f7af69926cf0f38/prisma/schema.prisma#L579) model since for schools we need the name as well.
I think the next step would be to move the school information we have on the pupil model to the school table...

Jonasdoubleyou (approved):

common/pupil/registration.ts
```diff
@@ -48,36 +48,33 @@ export interface BecomeStatePupilData {
...
             email: data.email.toLowerCase(),
             firstname: data.firstname,
             lastname: data.lastname,
             newsletter: data.newsletter,
             createdAt: new Date(),
-            schooltype: data.schooltype,
-            schoolId: data.schoolId,
-            state: data.state,
+            schooltype: data.school?.schooltype,
+            state: data.school?.state,
```
Jonasdoubleyou, line 75: Aaah, here you are taking data.school not school - overlooked that ...

Jonasdoubleyou (dismissed):

Jonasdoubleyou (approved):

PR #1126, opened 2024-08-07
added useful error messages why a subcourse can not be promoted

paul-jaworski:


common/courses/notifications.ts
```diff
@@ -185,8 +185,47 @@ export const canPromoteSubcourse = async (subcourse: Prisma.subcourse, attempted
 export async function sendPupilCoursePromotion(subcourse: Prisma.subcourse, promotionType: Prisma.subcourse_promotion_type_enum) {
     const { allowed, reason } = await canPromoteSubcourse(subcourse, promotionType);
     if (!allowed) {
-        logger.info(`Can't promote Subcourse(${subcourse.id}). Reason: ${reason}`);
```
JeangelLF, line 188: I think we could leave this here and remove all the duplicates in each `if` statement :)

common/courses/notifications.ts
```diff
@@ -185,8 +185,47 @@ export const canPromoteSubcourse = async (subcourse: Prisma.subcourse, attempted
...
+            throw new Error(`Instructors can only promote a Subcourse once!`);
+        }
+        if (reason == 'no-admin-promotions-left') {
+            logger.info(`Can't promote Subcourse(${subcourse.id}). Reason: ${reason}`);
+            throw new Error(`Admins can only promote a Subcourse once!`);
+        }
+        if (reason == 'recently-promoted') {
+            logger.info(`Can't promote Subcourse(${subcourse.id}). Reason: ${reason}`);
+            throw new Error(`Subcourse(${subcourse.id}) has already been promoted less than 3 days ago!`);
+        }
```
JeangelLF, line 228: Just to be completely sure that this fails, even if `reason` doesn't match any of the `if`s we should throw an error in the end
```
        throw new Error(`Promotion for Subcourse(${subcourse.id}) is not valid!`);
```

common/courses/notifications.ts
```diff
@@ -185,8 +185,47 @@ export const canPromoteSubcourse = async (subcourse: Prisma.subcourse, attempted
 export async function sendPupilCoursePromotion(subcourse: Prisma.subcourse, promotionType: Prisma.subcourse_promotion_type_enum) {
     const { allowed, reason } = await canPromoteSubcourse(subcourse, promotionType);
     if (!allowed) {
-        logger.info(`Can't promote Subcourse(${subcourse.id}). Reason: ${reason}`);
-        throw new Error(`Promotion for Subcourse(${subcourse.id}) is not valid!`);
+        if (reason == 'invalid-promotion-type') {
+            logger.info(`Can't promote Subcourse(${subcourse.id}). Reason: ${reason}`);
+            throw new Error(`Promotion type for Subcourse(${subcourse.id}) is not valid!`);
```
Jonasdoubleyou, line 190: What is the purpose of this change? Throwing an error will be reported as an internal error to the client.

paul-jaworski:
This PR belongs to the Issue: https://github.com/corona-school/project-user/issues/1197

How should I throw the Error instead?

common/courses/notifications.ts
paul-jaworski: This PR belongs to the Issue: corona-school/project-user#1197

I searched the code and found this way of throwing Errors. How should I throw the Error instead?

Jonasdoubleyou: Extend https://github.com/corona-school/backend/blob/master/common/util/error.ts with a new error class, maybe "NotAllowedError"

Jonasdoubleyou: That error class should have the reason as property

Jonasdoubleyou: Then adapt formatMessage in https://github.com/corona-school/backend/blob/master/graphql/error.ts to pass that reason to the client. It can then be catched and handled in the frontend.

Jonasdoubleyou: Maybe also adapt https://github.com/corona-school/backend/blob/master/common/util/decision.ts assertAllowed to throw that new error class and reuse that here

paul-jaworski: I created a new commit to implement your suggestion. However this seems like duplicate code to me. Maybe it would make more sense to export a new client error variable called "NotAllowedError" and then give the reason as the publicMessage?

Jonasdoubleyou: Ja well "That error class should have the reason as property" then it would not be redundant. 

But would also be OK to just create a new error type and then set the message to the reason

Jonasdoubleyou: throw new NotAllowedError(reason); ?

paul-jaworski: Yes I looked at how the other types of client errors were thrown and imitated that.

common/courses/notifications.ts
```diff
@@ -186,7 +187,29 @@ export async function sendPupilCoursePromotion(subcourse: Prisma.subcourse, prom
...
+        } else if (reason == 'already-auto-promoted') {
+            throw new NotAllowedError(`Subcourse(${subcourse.id}) is already promoted automatically!`);
+        } else if (reason == 'no-instructor-promotions-left') {
+            throw new NotAllowedError(`Instructors can only promote a Subcourse once!`);
+        } else if (reason == 'no-admin-promotions-left') {
+            throw new NotAllowedError(`Admins can only promote a Subcourse once!`);
+        } else if (reason == 'recently-promoted') {
+            throw new NotAllowedError(`Subcourse(${subcourse.id}) has already been promoted less than 3 days ago!`);
+        } else {
+            throw new NotAllowedError('Subcourse(${subcourse.id}) can not be promoted for a different reason: ${reason}!');
```
JeangelLF, line 211: This should use ` instead of '

JeangelLF (changes requested):

common/courses/notifications.ts
Jonasdoubleyou: Again: throw new NotAllowedError(reason); - there is no point in showing english errors to our users, translation to a text should be done in the frontend.

common/courses/notifications.ts
```diff
@@ -185,8 +186,7 @@ export const canPromoteSubcourse = async (subcourse: Prisma.subcourse, attempted
 export async function sendPupilCoursePromotion(subcourse: Prisma.subcourse, promotionType: Prisma.subcourse_promotion_type_enum) {
     const { allowed, reason } = await canPromoteSubcourse(subcourse, promotionType);
     if (!allowed) {
-        logger.info(`Can't promote Subcourse(${subcourse.id}). Reason: ${reason}`);
```
JeangelLF, line 188: Could we keep the log? It's very helpful for debugging purposes :)

JeangelLF (approved):
Thank you!

PR #1134, opened 2024-09-05
Feat/screener course edit

Fabian5150:
Closes part of https://github.com/corona-school/project-user/issues/1201
(Appointments can NOT be edited, the rest of the course can now
@FForstbach decided that the rest of the ticket is not needed for now as the need for this feature is not that big)


common/appointment/cancel.ts
```diff
@@ -1,12 +1,13 @@
...
 import { prisma } from '../prisma';
-import { getStudent, User, userForPupil, userForStudent } from '../user';
+import { getScreener, getStudent, User, userForPupil, userForStudent } from '../user';
 import * as Notification from '../notification';
 import { getLogger } from '../logger/logger';
 import { getAppointmentForNotification } from './util';
 import { deleteZoomMeeting } from '../zoom/scheduled-meeting';
 import { PrerequisiteError, RedundantError } from '../util/error';
 import { getNotificationContextForSubcourse } from '../courses/notifications';
+import { Student } from '../../graphql/generated';
```
Jonasdoubleyou, line 10: Please avoid importing ./graphql (the API layer) in ./common (c.f. "onion architecture"). You can use the student (lowercase :/) import from `@prisma/client`

common/appointment/cancel.ts
```diff
@@ -50,7 +51,7 @@ export async function cancelAppointment(user: User, appointment: Appointment, si
                 if (!silent) {
                     await Notification.actionTaken(userForPupil(participant.pupil), 'student_cancel_appointment_group', {
                         appointment: getAppointmentForNotification(appointment),
-                        student,
+                        student: _user,
```
Jonasdoubleyou, line 54: The build (rightfully) fails here - Do we want to send the same notification in this case? Or a different one? And do we want to expose the name of the screener to the course participants?

common/appointment/cancel.ts
Fabian5150: That seems to have been a false auto-inserted import done by vscode, it wasn't even used.
I removed it. :)

Fabian5150:
@Jonasdoubleyou I completely forogot, that I have already made some little changes here to allow Screeners to edit/delete appointments.
As mentioned this feature isn't needed for now, so all the backend changes here aren't needed anyway.
I will just remove this PR and change the frontend so that screeners don't get any buttons for appointments...

graphql/appointment/fields.ts
```diff
@@ -80,7 +80,7 @@ export class ExtendedFieldsLectureResolver {
         return isParticipant;
     }
     @FieldResolver((returns) => [AppointmentParticipant], { nullable: true })
-    @Authorized(Role.OWNER, Role.APPOINTMENT_PARTICIPANT, Role.ADMIN)
```
Jonasdoubleyou, line 83: @Fabian5150 I think you still need this one - I currently get

```
Requiring one of the following roles: OWNER,APPOINTMENT_PARTICIPANT,ADMIN
```

Fabian5150: Sorry I missed your comment here. John added it by now and it seems to work. :)

PR #1135, opened 2024-09-08
feat: Matching in Typescript

Jonasdoubleyou:
Moves the Matching Algorithm to Typescript. Replaces the much more complicated minimum cost maximum flow problem with a much simpler assignment problem - solutions should be equivalent (given the same cost function).

Only introduces the new algorithm, does not yet switch to it. 

https://github.com/corona-school/project-user/issues/1316


Jonasdoubleyou:
First mini step, next I would:
- Improve the cost function (don't think we need the complexity of the current cost function though)
- Add some larger test sets and run the old against the new algorithm
- Switch to the new implementation with a feature toggle
- Then someday retire the C++ variant

JeangelLF (approved):
Nice! Thanks for the overview of the next steps. It sounds good 💯

PR #1142, opened 2024-09-14
feat: Matching algo performance + Exclude duplicate matchings

Jonasdoubleyou:
Splits the tests into small unit tests that are fast to run (`npm run unit`) and perf tests that might take longer (`npm run perf`) - `npm run test` runs all of them in the barrier. 

Adds a performance test for the matching algorithm, which simulates a real world matching - pupils and students continously request matches, every N days the matching algorithm is run using the current pools, matched users are removed from the pool. At the end, reports how many matches are created, how many subjects and states match, and how long the pupils and students had to wait for the match.

Extends the Matching Algorithm to exclude duplicate matches (as the old implementation does). Currently implemented as a postfilter after the matching, dont think it happens often and is worth considering in the assignment itself. 

https://github.com/corona-school/project-user/issues/1316

### DIFF
--- a/common/match/matching.perf.ts
+++ b/common/match/matching.perf.ts
@@ -74,8 +74,8 @@ async function computeOldMatchings(
     });
 
     const matching: Matching = result.matches.map((it) => ({
-        request: requests[+it.helpee.uuid],
-        offer: offers[+it.helper.uuid],
+        request: requests[+it.helpee.id],
+        offer: offers[+it.helper.id],
     }));
 
     return matching;
--- a/common/match/matching.perf.ts
+++ b/common/match/matching.perf.ts
@@ -73,9 +73,12 @@ async function computeOldMatchings(
         },
     });
 
+    const requestByUUID = new Map(requests.map((it) => [`${it.pupilId}`, it]));
+    const offerByUUID = new Map(offers.map((it) => [`${it.studentId}`, it]));
+
     const matching: Matching = result.matches.map((it) => ({
-        request: requests[+it.helpee.id],
-        offer: offers[+it.helper.id],
+        request: requestByUUID.get(it.helpee.uuid),
+        offer: offerByUUID.get(it.helper.uuid),
     }));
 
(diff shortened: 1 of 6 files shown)


realmayus (approved):
Nice work!

common/match/matching.perf.ts
```diff
@@ -0,0 +1,339 @@
...
+                pupilWaitingTimeAvg: 20.26238745497861,
+                studentWaitingTimeAvg: 61.03486072712683,
+                matchRuns: 67,
+            },
+        ],
+        [
+            'new',
+            1,
+            {
+                matchCountSum: 1044,
```
Jonasdoubleyou, line 140: As the new algorithm is able to create more matches with the same "quality" (about the same number of matching subjects and states) I think it is worth to switch.

common/match/matching.perf.ts
```diff
@@ -0,0 +1,339 @@
...
+            },
+        ],
+        // Old Algorithm
+        // Interestingly the old algorithm is not deterministic
+        // Also running it in the barrier does not make sense as it wont change anyways
+        /* [
+            'old',
+            1000,
+            {
+                matchCountSum: 883,
```
Jonasdoubleyou, line 157: It's interesting that the old algorithm prioritizes the number of matching states over the number of matching subjects and the number of matches created ...

common/match/matching.perf.ts
```diff
@@ -0,0 +1,339 @@
...
+    return matching;
+}
+
+describe('Real World Matching Performance', () => {
+    test.each([
+        // New Algorithm
+        [
+            'new',
+            1000,
+            {
```
Jonasdoubleyou, line 111: Using such a fixed time series as a benchmark is not so optimal - it can easily lead to overfitting onto this time series. Some sort of shuffling of the input would be good, but on the other hand that might loose some time effects that are actually there (e.g. pupils / students signing up in bulk after marketing campaigns)

PR #1144, opened 2024-09-22
feat: Switch to the new matching algorithm

Jonasdoubleyou:
resolves https://github.com/corona-school/project-user/issues/1316


JeangelLF (commented):
Should we do the change using a feature flag just in case? 🤔

Other than that this looks great! 🙌

Jonasdoubleyou:
Would just revert the change in case something goes wrong. There arent many other backend changes in anyways.

JeangelLF (approved):
🚀

PR #1151, opened 2024-10-10
Feat: Statistics - support pupil_screenings in knowsCoronaSchoolFrom

realmayus:
Closes https://github.com/corona-school/project-user/issues/1204.


graphql/statistics/fields.ts
```diff
@@ -680,7 +680,19 @@ export class StatisticsResolver {
...
     @FieldResolver((returns) => [ByMonth])
     @Authorized(Role.ADMIN)
-    async knowsCoronaSchoolFrom(@Root() statistics: Statistics) {
+    async knowsCoronaSchoolFrom(@Root() statistics: Statistics, @Arg('type') type: 'pupil' | 'student') {
+        if (type === 'pupil') {
+            return await prisma.$queryRaw`SELECT COUNT(*)::INT                         AS value,
+                                             date_part('year', "createdAt"::date)  AS year,
+                                             date_part('month', "createdAt"::date) AS month,
+                                             "knowsCoronaSchoolFrom"               AS group
+                                      FROM "pupil_screening"
```
Jonasdoubleyou, line 689: As there are N screenings for each pupil, doesnt this skew the statistics towards more active users?

Jonasdoubleyou (approved):

graphql/statistics/fields.ts
realmayus: Good catch! How should we handle multiple screenings @FForstbach ? Should we only consider the most recent screening that has a `knowsCoronaSchoolFrom` entry?

FForstbach: so i would say:
1. only count once
2. save the first real answer (dont know, or empty does not count)
3. dont save subsequent entries (since we want to know how they first heard about coronaschool)

does that make sense @realmayus ?

Jonasdoubleyou (approved):

PR #1156, opened 2024-10-24
Fix: return full participants from prospectParticipants field

realmayus:
We should return prospectParticipants as `Participant`s such that we can reuse parts of the waitinglist UI on the front end.


JeangelLF (dismissed):
**Not a blocker**: It would be nice to create a separate function to fetch a participant _(With the `where` as a parameter)_, I think we already have a similar query in this file 👀 
`pupilsOnWaitinglist` and `participants`.

JeangelLF (approved):
Nice!

PR #1159, opened 2024-11-14
Feat: Add new actions for DaZ Matches and Instructors Only Users

JeangelLF:
## Ticket

- Resolves https://github.com/corona-school/project-user/issues/1275
- Resolves https://github.com/corona-school/project-user/issues/1281

## What was done?

- Add/Trigger new action when a DaZ match is created
- Add a hook/action to invite Instructors only users to reflection meetings after completing their very first group appointment

### DIFF
--- a/common/notification/hooks.ts
+++ b/common/notification/hooks.ts
@@ -4,6 +4,10 @@ import { registerPupilHook, registerStudentHook } from './hook';
 //  this ensures that the hooks are always registered when the Notification is loaded (i.e. in the jobs Deno, which only loads parts of the backend)
 import { deactivateStudent } from '../student/activation';
 import { cancelRemissionRequest } from '../remission-request';
+import { prisma } from '../prisma';
+import { userForStudent } from '../user';
+import * as Notification from '../../common/notification';
+import { SpecificNotificationContext } from './actions';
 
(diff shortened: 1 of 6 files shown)


common/notification/hooks.ts
```diff
@@ -17,6 +21,27 @@ registerStudentHook('cancel-remission-request', 'Cancels the remission request(s
     await cancelRemissionRequest(student);
 });
 
+registerStudentHook(
+    'attempt-invite-instructor-to-reflection-meeting',
+    'Instructors are invited to participate in a reflection meeting after their very first group appointment',
+    async (student, context) => {
+        if (student.isStudent) {
```
realmayus, line 28: is it right that we don't send the invitation if the instructor is a tutor in addition to being an instructor?

realmayus (changes requested):

common/notification/hooks.ts
JeangelLF: Yeah, I think that's not fully correct. The requested condition was.
> HuH, die nur Kurse geben und kein Match haben

I adjusted the condition to just verify that the user doesn't have an active match.

Thanks!

realmayus (approved):

PR #1175, opened 2025-01-02
feat: Cleanup match to match_pool_run relation

Jonasdoubleyou:
Add foreign key relationship between match and match run to track all matches that were created at once,
 additionally drop Match.source as it is unused since 4 years + misleading.


realmayus (approved):
Nice!

PR #1179, opened 2025-01-06
fix: Invalid Zoom Meeting Id

JeangelLF:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1395

## What was done?

Added a patch for those rare cases where Zoom doesn't recognize a `meetingId` they gave us. As explained in the ticket, we are trying to recreate a Zoom meeting only if

- The user is the owner of the appointment (?) or admin
- The appointment is not canceled
- The appointment has not yet ended
- There is a ZoomMeetingId
- Zoom doesn't recognize the ZoomMeetingId we have

### DIFF
--- a/graphql/appointment/fields.ts
+++ b/graphql/appointment/fields.ts
@@ -209,9 +209,6 @@ export class ExtendedFieldsLectureResolver {
                 logger.error(`Zoom Meeting Id (${appointment.zoomMeetingId}) expired or deleted`);
                 await deleteZoomMeeting(appointment);
                 await createZoomMeetingForAppointment(await getLecture(appointment.id));
-                if (isAppointmentOrganizer) {
-                    return await getZoomUrl(user, await getLecture(appointment.id));
-                }
             }
         }
 
@@ -220,7 +217,7 @@ export class ExtendedFieldsLectureResolver {
             logger.info(`Admin requested zoom meeting url`);
             return zoomMeeting.join_url;
         }
-        return await getZoomUrl(user, appointment);
+        return await getZoomUrl(user, await getLecture(appointment.id));
     }
 
     @FieldResolver((returns) => Match, { nullable: true })
--- a/graphql/appointment/fields.ts
+++ b/graphql/appointment/fields.ts
@@ -8,14 +8,10 @@ import { LimitEstimated } from '../complexity';
 import { prisma } from '../../common/prisma';
 import { getUserTypeAndIdForUserId, getUsers, getUser } from '../../common/user';
 import { GraphQLJSON } from 'graphql-scalars';
-import { deleteZoomMeeting, getZoomMeeting } from '../../common/zoom/scheduled-meeting';
+import { getZoomMeeting } from '../../common/zoom/scheduled-meeting';
 import { UserType } from '../types/user';
 import { getZoomUrl } from '../../common/zoom/user';
 import { getLogger } from '../../common/logger/logger';
-import moment from 'moment';
-import { getAppointmentEnd } from '../../common/appointment/util';
-import { ZoomError } from '../../common/util/error';
-import { createZoomMeetingForAppointment } from '../../common/appointment/create';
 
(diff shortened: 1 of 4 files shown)


graphql/appointment/fields.ts
```diff
@@ -183,6 +187,34 @@ export class ExtendedFieldsLectureResolver {
...
+        const isOrganizerOrAdmin = isAppointmentOrganizer || isAdmin;
+
+        // If any of these we just use the previous logic
+        if (!isOrganizerOrAdmin || !isAppointmentInTheFuture || isAppointmentCanceled) {
+            return await getZoomUrl(user, appointment);
+        }
+
+        try {
+            // Asking Zoom for the meeting details is the only way to verify if Zoom recognizes the meetingId we have
+            await getZoomMeeting(appointment);
```
Jonasdoubleyou, line 203: Hm, field resolvers should usually be side effect free - so maybe we should turn this into a mutation ...

graphql/appointment/fields.ts
```diff
@@ -183,6 +187,34 @@ export class ExtendedFieldsLectureResolver {
...
+        } catch (error) {
+            const zoomError = error as ZoomError;
+            // If for some reason this meeting is now expired/deleted according to zoom (which shouldn't be the case)
+            // We just try to recreate it.
+            if (zoomError?.status === 404 && zoomError?.code === 3001) {
+                logger.error(`Zoom Meeting Id (${appointment.zoomMeetingId}) expired or deleted`);
+                await deleteZoomMeeting(appointment);
+                await createZoomMeetingForAppointment(await getLecture(appointment.id));
+                if (isAppointmentOrganizer) {
```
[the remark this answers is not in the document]
JeangelLF: Yeah, I think this is not needed because it does the same as we do at the end of this function 👍 I'll update it

graphql/appointment/fields.ts
JeangelLF: Yep, makes sense, I'll do it.

JeangelLF: Here is the frontend PR https://github.com/corona-school/user-app/pull/748

graphql/appointment/mutations.ts
```diff
@@ -178,4 +181,50 @@ export class MutateAppointmentResolver {
...
+            await getZoomMeeting(appointment);
+        } catch (error) {
+            const zoomError = error as ZoomError;
+            // If for some reason this meeting is now expired/deleted according to zoom (which shouldn't be the case)
+            // We just try to recreate it.
+            if (zoomError?.status === 404 && zoomError?.code === 3001) {
+                logger.error(`Zoom Meeting Id (${appointment.zoomMeetingId}) expired or deleted`);
+                await deleteZoomMeeting(appointment);
+                await createZoomMeetingForAppointment(await getLecture(appointment.id));
+            }
```
Jonasdoubleyou, line 220: else rethrow?

Jonasdoubleyou (approved):

graphql/appointment/fields.ts
```diff
@@ -183,6 +187,34 @@ export class ExtendedFieldsLectureResolver {
...
+        try {
+            // Asking Zoom for the meeting details is the only way to verify if Zoom recognizes the meetingId we have
+            await getZoomMeeting(appointment);
+        } catch (error) {
+            const zoomError = error as ZoomError;
+            // If for some reason this meeting is now expired/deleted according to zoom (which shouldn't be the case)
+            // We just try to recreate it.
+            if (zoomError?.status === 404 && zoomError?.code === 3001) {
+                logger.error(`Zoom Meeting Id (${appointment.zoomMeetingId}) expired or deleted`);
+                await deleteZoomMeeting(appointment);
```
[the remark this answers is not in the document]
Jonasdoubleyou: Hm, doing a few requests to the Zoom API takes some time, so I don't think it is that unlikely that it happens ... But up to you

Jonasdoubleyou (approved):

PR #1180, opened 2025-01-08
Feat: Add pupil email owner field

JeangelLF:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1386

## What was done?

- Added a new `emailOwner` field on the pupil
- Updated pupil registration input data
- Adjusted tests & seed data

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -1559,6 +1559,6 @@ enum gender_enum {
 enum pupil_email_owner_enum {
   pupil
   parent
-  supportPerson
+  other
   unknown
 }
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -493,7 +493,7 @@ model match {
 
   // For "automatic matches", stores the Match Run that created this match (since 01/2025)
   matchPoolRunId Int?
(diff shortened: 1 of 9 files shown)


prisma/schema.prisma
```diff
@@ -1552,3 +1555,10 @@ enum gender_enum {
   female
   other
 }
+
+enum pupil_email_owner_enum {
+  pupil
+  parent
+  supportPerson
```
Jonasdoubleyou, line 1562: "support" sounds like our support ....

Jonasdoubleyou (approved):
So we only want to store a single email address? Not multiple?

JeangelLF:
Yeah, for now, we just want to collect this information so we can later decide whether to fully implement a similar feature to what you've described or determine if it won't be necessary

prisma/schema.prisma
JeangelLF: Oh... Now I can't unsee it 🙈.
This was my attempt at translating "Hilfsperson", probably "AssitingAdult" sounds better. Or simply "other"?

Jonasdoubleyou: Yeah maybe other

PR #1182, opened 2025-01-16
Feat: Add homework help course category

realmayus:
Add homework help course category and make it impossible for non-authorized users to create such courses


JeangelLF (commented):
Thanks for the PR! Could we also add a validation for the `courseEdit` scenario? :)

JeangelLF (approved):
🚀

PR #1183, opened 2025-01-16
Fix: token revoke

realmayus:
remove certain `prisma.secret.find` calls, simplify `tokenRevoke` logic and provide better error handling for invalid tokens


Jonasdoubleyou (approved):

graphql/secret/mutation.ts
```diff
@@ -63,24 +63,24 @@ export class MutateSecretResolver {
         @Arg('id', { nullable: true }) id?: number,
         @Arg('token', { nullable: true }) token?: string
     ) {
-        let tokenId = id;
-        let deviceId = undefined;
+        let secret = undefined;
+
         if (id) {
-            deviceId = (await prisma.secret.findUnique({ where: { id: tokenId } })).lastUsedDeviceId;
+            secret = await prisma.secret.findUnique({ where: { id } });
```
Jonasdoubleyou, line 69: It's a bit tricky that here we also allow retrieving secrets of other users, and only later validate in revokeSecret. Maybe we could additionally filter `userId: getSessionUser(context).userID`

Jonasdoubleyou (approved):

common/secret/token.ts
```diff
@@ -16,35 +16,22 @@ import { PrerequisiteError } from '../util/error';
 
 const logger = getLogger('Token');
 
-export async function revokeToken(user: User | null, id: number) {
-    const result = await prisma.secret.deleteMany({ where: { id, userId: user?.userID } });
+export async function revokeSecret(userId: string | undefined, id: number) {
```
Jonasdoubleyou, line 19: I find User a bit more typesafe than string, but both are okay

common/user/session.ts
```diff
@@ -74,13 +74,13 @@ export async function updateSessionRolesOfUser(userID: string) {
...
 // Currently used in session manager to log out all sessions created by a specific device token
-export async function deleteSessionsByDevice(deviceId: string | null, user: User) {
+export async function deleteSessionsByDevice(deviceId: string | null, user: User | undefined) {
     if (!deviceId) {
         return; // do nothing if deviceId is undefined
     }
     const sessionsToDelete = [];
     for await (const [sessionToken, sessionUser] of userSessions.iterator() as AsyncIterable<[string, GraphQLUser]>) {
-        if (sessionUser.deviceId === deviceId && sessionUser.userID === user.userID) {
+        if (sessionUser.deviceId === deviceId && sessionUser.userID === user?.userID) {
```
realmayus, line 83: Do we actually need the userID check here? I can't think of a scenario where we need this, and removing the sessionUser.userID === user?.userID check would also allow admins to log out other users.

Jonasdoubleyou: As DeviceIDs are generated on a user controlled device, malicious users could on purpose generate device ids that collide with other users.

common/user/session.ts
```diff
@@ -74,13 +74,13 @@ export async function updateSessionRolesOfUser(userID: string) {
...
-export async function deleteSessionsByDevice(deviceId: string | null, userID: string) {
+export async function deleteSessionsByDevice(deviceId: string | null, user: User | undefined) {
     if (!deviceId) {
         return; // do nothing if deviceId is undefined
     }
     const sessionsToDelete = [];
-    for await (const [sessionToken, user] of userSessions.iterator() as AsyncIterable<[string, GraphQLUser]>) {
-        if (user.deviceId === deviceId && user.userID === userID) {
+    for await (const [sessionToken, sessionUser] of userSessions.iterator() as AsyncIterable<[string, GraphQLUser]>) {
+        if (sessionUser.deviceId === deviceId && sessionUser.userID === user?.userID) {
```
Jonasdoubleyou, line 83: If user is not passed in, the second condition will always be false ... I guess it should be (!user || ...)

Jonasdoubleyou (approved):

Jonasdoubleyou (approved):

PR #1185, opened 2025-01-17
Fix: Instant Certs: only count lectures that are in the past, add date variable, friendly date formatting

realmayus:
- only count lectures that are in the past
- add date variable so our template renders correctly
- friendly date formatting


common/certificate/index.ts
```diff
@@ -454,14 +457,16 @@ async function createInstantPDFBinary(certificate: InstantCertificate & { studen
     if (process.env.ENV == 'dev') {
         name = `[TEST] ${name}`;
     }
+    console.log(`VERIFICATION ${link}`);
```
JeangelLF, line 460:
```suggestion
```
Let's remove this one. Other than that, it looks good!

realmayus: Oops!

JeangelLF (approved):

PR #1191, opened 2025-01-31
feat: Google SSO

JeangelLF:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1217

## What was done?

- Added mutation to login with SSO (for now only with Google)
- Added new role for users registering with SSO
- Added mutations to create students/pupils using SSO
- Added IDP linking possibility when users authenticate with a password


graphql/authentication.ts
```diff
@@ -246,4 +257,35 @@ export class AuthenticationResolver {
...
+                email,
+                firstname,
+                lastname,
+                idpClientId: clientId,
+            });
+            return SSOAuthStatus.register;
+        }
+        const hasIDPLogin = await userHasIDPLogin(user.userID, clientId);
+        if (!hasIDPLogin) {
+            await createIDPLogin(user.userID, clientId);
```
JeangelLF, line 286: Should we take some other actions in case the user exists but doesn't have a connection with the given IDP? 🤔
For example, a user with a password trying to log in with their Google account

prisma/schema.prisma
```diff
@@ -1136,6 +1136,13 @@ model job_run {
   @@id([job_name, startedAt])
 }
 
+model user_idp_login {
```
JeangelLF, line 1139: I'm now wondering if it would make sense to save this in the `secrets` table instead.

graphql/authentication.ts
```diff
@@ -246,4 +257,35 @@ export class AuthenticationResolver {
...
+    @Authorized(Role.UNAUTHENTICATED)
+    @Mutation((returns) => SSOAuthStatus)
+    async loginWithSSO(@Ctx() context: GraphQLContext, @Arg('code') code: string) {
+        const { email, firstname, lastname, clientId } = await authenticate(code);
+        if (!email || !firstname) {
+            throw new Error('Invalid token payload: Missing required fields (email/name)');
+        }
+        let user: User;
+        try {
+            user = await getUserByEmail(email, true);
```
Jonasdoubleyou, line 270: https://openid.net/specs/openid-connect-core-1_0.html#ClaimStability

>All other Claims carry no such guarantees across different issuers in terms of stability over time or uniqueness across users, and Issuers are permitted to apply local restrictions and policies. For instance, an Issuer MAY re-use an email Claim Value across different End-Users at different points in time, and the claimed email address for a given End-User MAY change over time. Therefore, other Claims such as email, phone_number, preferred_username, and name MUST NOT be used as unique identifiers for the End-User, whether obtained from the ID Token or the UserInfo Endpoint.

To be spec compliant, we should store the `sub` unique identifier, and then use that to find the user, other than relying on the email address - the IDP is not even required to return a verified email here (just "the users preference for contact").

In practice though I guess most IDPs should return a unique + verified email address, so I guess the current variant works find in the 99% case ... But who knows, maybe one day we want to support an IDP that does not provide an email of the user ...

Jonasdoubleyou: Ah, email change might be relevant - If I change the email of my account at IDP XY, I might still want to log in to Lern-Fair using the same account ...

graphql/authentication.ts
Jonasdoubleyou: Good point, once we add the option to "Log in with ..." existing users might take it. I guess the proper way would be to allow "account linking", e.g. redirect to the password prompt, let the user log in with password, then during password login pick up the IDP information and store it.

We should not rely on the fact that the user owns the email just because the IDP says so, that makes this thing tricky ...

Also I think as we "link" the accounts, it would be good to get the users explicit consent to comply with GDPR.

graphql/types/userInputs.ts
```diff
@@ -52,7 +52,22 @@ export class RegistrationSchool {
 }
 
 @InputType()
-export class RegisterPupilInput implements RegisterPupilData {
+class RegisterPupilInputShared {
```
Jonasdoubleyou, line 55: Hm, I do have some accounts at IDPs with bogus names and might want to change the name when I register at Lern-Fair. Maybe we should return name + email, then allow the user to change them? 

If we allow that, it might be easier to use the existing registerPupil/Student, and just pick up IDP info stored in the session

prisma/schema.prisma
Jonasdoubleyou: Hm, good question, my instinct would also be to store it in secrets, but separate table is also fine. Having it in secrets we can reuse the existing logic for users to view + manage their login options, e.g. to revoke secrets.

Jonasdoubleyou: Would also nicely fit, the "secret" would be the IDP "sub", if we have a IDP table we could add a foreign key, otherwise store add some sort of client column to the secrets table.

Jonasdoubleyou (commented):
Sweet, looking forward to "Log in with Lern-Fair" ;)

graphql/types/userInputs.ts
Jonasdoubleyou: Also e.g. if parents sign on, they might want to change the name to their child?

graphql/me/mutation.ts
```diff
@@ -99,21 +121,32 @@ export class MutateMeResolver {
...
-        const pupil = await registerPupil(data, noEmail);
+        if (isSSO && (!sessionUser.idpClientId || !sessionUser.idpSub)) {
+            throw new Error(`Cannot complete request without an IDP ClientID / IDP Sub`);
+        }
+
+        const pupil = await registerPupil({ ...data, email: isSSO ? sessionUser.email : data.email }, noEmail);
         logger.info(`Pupil(${pupil.id}, firstname = ${pupil.firstname}, lastname = ${pupil.lastname}) registered`);
+        if (isSSO) {
+            const user = userForPupil(pupil);
+            await verifyEmail(user);
```
Jonasdoubleyou, line 147: Maybe comment that for now we trust the IDP here to validate the email address?

graphql/me/mutation.ts
```diff
@@ -99,21 +121,32 @@ export class MutateMeResolver {
...
     @Mutation((returns) => Pupil)
-    @Authorized(Role.UNAUTHENTICATED, Role.ADMIN)
+    @Authorized(Role.UNAUTHENTICATED, Role.ADMIN, Role.SSO_REGISTERING_USER)
     @RateLimit('RegisterPupil', 10 /* requests per */, 5 * 60 * 60 * 1000 /* 5 hours */)
     async meRegisterPupil(
         @Ctx() context: GraphQLContext,
         @Arg('data') data: RegisterPupilInput,
         @Arg('noEmail', () => GraphQLBoolean, { nullable: true }) noEmail = false
     ) {
+        const sessionUser = await getUserForSession(context.sessionToken);
```
Jonasdoubleyou, line 131: Any reason to not use context.user?

graphql/me/mutation.ts
```diff
@@ -99,21 +121,32 @@ export class MutateMeResolver {
...
         const byAdmin = context.user.roles.includes(Role.ADMIN);
+        const isSSO = context.user.roles.includes(Role.SSO_REGISTERING_USER);
```
Jonasdoubleyou, line 133: `// User registered after trying to log in via IDP, pick up the IDP user info`

graphql/authentication.ts
```diff
@@ -246,4 +257,38 @@ export class AuthenticationResolver {
...
+                email,
+                firstname,
+                lastname,
+                idpClientId: clientId,
+                idpSub: sub,
+            });
+            return SSOAuthStatus.register;
+        }
+        const hasIDPLogin = await userHasIDPLogin(user.userID, sub, clientId);
+        if (!hasIDPLogin) {
```
Jonasdoubleyou, line 288: Unreachable?

graphql/authentication.ts
```diff
@@ -246,4 +257,38 @@ export class AuthenticationResolver {
...
+            await userSessions.set(context.sessionToken, {
+                ...context.user,
+                roles: Array.from(newRoles),
+                email,
+                firstname,
+                lastname,
+                idpClientId: clientId,
+                idpSub: sub,
+            });
+            return SSOAuthStatus.register;
```
Jonasdoubleyou, line 285: Check for email collision, then return a "link" status instead?

common/idp/index.ts
```diff
@@ -0,0 +1,27 @@
...
+    sub: string;
+}
+
+interface AuthenticateArgs {
+    referrer: string;
+    code: string;
+}
+
+const identityProviders = {
+    'https://accounts.google.com/': authenticateWithGoogle,
```
Jonasdoubleyou, line 17: How generic is the Google OAuthClient? If it is generic, we could just reuse it for other IDPs:

```
'google.com': new OAuthClient(GOOGLE_SECRET),
'other.com': new OAuthClient(OTHER_SECRET)
```

graphql/authentication.ts
JeangelLF: So ideally for this `SSOAuthStatus.link` flow we'd need another role similar to how we built the `SSOAuthStatus.register` flow, right?

graphql/authentication.ts
JeangelLF: Right, this was only relevant when we used email as an identifier

common/idp/index.ts
JeangelLF: Unfortunately it doesn't seem that flexible. I'd say if we end up adding more IDPs we could try something more generic

graphql/authentication.ts
Jonasdoubleyou: Right, I think we can just use the existing role here?

graphql/authentication.ts
```diff
@@ -246,4 +272,52 @@ export class AuthenticationResolver {
...
+                // If so, we'll try to link that email with the given sub/clientId
+                await setSSORegisteringUser();
+                return SSOAuthStatus.link;
+            }
+        }
+
+        let user: User;
+        try {
+            user = await getUser(userId, true);
+        } catch (error) {
```
Jonasdoubleyou, line 313: I don't follow, how can `getUser(...)` ever fail? Shouldn't this be the `else` branch of isEmailAvailable?

graphql/authentication.ts
```diff
@@ -246,4 +272,52 @@ export class AuthenticationResolver {
...
+            // User is not registered. They'll have to complete the registration flow
+            await setSSORegisteringUser();
+            return SSOAuthStatus.register;
+        }
+
+        // A user with the given IDP sub/clientId was found
```
Jonasdoubleyou, line 319: Would find it easier to follow if moved into the try block

common/user/roles.ts
```diff
@@ -49,4 +49,6 @@ export enum Role {
     SUBCOURSE_PARTICIPANT = 'SUBCOURSE_PARTICIPANT',
     /* User is a pupil and participant of a specific appointment */
     APPOINTMENT_PARTICIPANT = 'APPOINTMENT_PARTICIPANT',
+    /** User tried to authenticate with SSO but doesn't have an account yet. They'll need to complete a part of the registration process to continue */
```
Jonasdoubleyou, line 52: Maybe adapt description to account for the loginPassword / link case?

Jonasdoubleyou (dismissed):
LGTM, PR message could be updated :)

JeangelLF:
Thanks for all your comments! 🙌 I'll be tackling the remaining improvements in a separate PR :)

PR #1192, opened 2025-01-31
Fix: Add pupils to future lectures

JeangelLF:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1421

## What was done?

- Updated `addGroupAppointmentsParticipant` to only add new participants/organizers to future lectures
- Adjusted seed data to include participants/organizers for past lectures


Jonasdoubleyou (commented):
What about ongoing lectures - consider the case that I join a course a little late after it was started?

JeangelLF:
Thanks for the heads up! I updated it to also consider `duration`

Jonasdoubleyou (approved):

PR #1194, opened 2025-02-11
feature(achievements): delete unreachable achievements

dhenkel92:
For every course a pupil attends, they can get a "successful finished" achievement.
These achievements cannot be reached if a course is done. Therefore, to not demotivate our pupils with unreachable achievements, we should delete them.

Fixes: https://github.com/corona-school/project-user/issues/1411


dhenkel92:
This change is part of the following stack:

- #1194 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

integration-tests/utils.ts
dhenkel92, line 1: For the new integration tests I had to play with the lecture start times. To not break tests that are depending on our global `subcourseOne` and `subcourseTwo`, I decided to create this helper function to spin up a completely new course.

All the logic was copied from `07_course.ts`.

jobs/periodic/delete-unreachable-achievements/courses.ts
```diff
@@ -0,0 +1,78 @@
...
+const logger = getLogger('CourseAchievementCleanup');
+
+const ParticipationGroupTemplate = 'pupil_course_participation';
+
+export const deleteUnreachableCourseAchievements = tracer.wrap('delete', _deleteUnreachableCourseAchievements);
+async function _deleteUnreachableCourseAchievements(dryRun: boolean = true) {
+    const till = new Date();
+    till.setDate(till.getDate() - 1);
+
+    const unachievedAchievementes = await prisma.user_achievement.findMany({
```
Jonasdoubleyou, line 15: Technically the relations would be enough, right? 

```
await prisma.user_achievement.findMany({
  distinct: ['relation'],
  select: {
    relation: true,
  },
})
```

That would reduce complexity by factor N :)

jobs/periodic/delete-unreachable-achievements/courses.ts
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        logger.info('deleting pupil course participation achievements', {
+            relation,
+            user_ids: userIds,
+        });
+
+        if (!dryRun) {
+            await prisma.user_achievement.deleteMany({
+                where: {
```
Jonasdoubleyou, line 68: `achievedAt: null,` just to be on the safe side (and it could still be achieved while the job is running?)

jobs/list.ts
```diff
@@ -62,4 +64,6 @@ export const regularJobs: ScheduledJob[] = [
     { cronTime: '00 00 10 * * *', name: 'flagInactiveConversationsAsReadonly' },
     // Every night, trigger actions for courses that ended yesterday
     { cronTime: '00 00 10 * * *', name: 'notificationsEndedYesterday' },
+    // Every six hours look for unreachable achievements and delete them
+    { cronTime: '00 00 */6 * * *', name: 'deleteUnreachableAchievements' },
```
Jonasdoubleyou, line 68: (as an alternative, we could also hook it into notificationsEndedYesterday - there we anyways loop courses that ended. But a separate job is also fine)

jobs/periodic/delete-unreachable-achievements/courses.ts
```diff
@@ -0,0 +1,78 @@
...
+    for (const [relation, userIds] of Object.entries(relationToUserIds)) {
+        const [relationType, subcourseId] = parseRelation(relation);
+        if (relationType !== EventRelationType.Subcourse) {
+            logger.warn('unexpected relation type', {
+                relation,
+            });
+            continue;
+        }
+
+        const futureLectureCnt = await prisma.lecture.count({
```
Jonasdoubleyou, line 46: Use https://github.com/corona-school/backend/blob/33c9681b9deae57856ff37f0f3ce8c3cb5217c7b/common/courses/states.ts#L65 ? Then we have a single place to decide (also there is the plan to have "infinite courses" which never end)

Jonasdoubleyou (approved):
LGTM, some small suggestions

jobs/periodic/delete-unreachable-achievements/courses.ts
Jonasdoubleyou: Here you could also easily add `|| subcourseCancelled(...)` :)

jobs/periodic/delete-unreachable-achievements/courses.ts
dhenkel92: Actually, we even need to delete achieved achievements. When a user joins a course, the first achievement in the sequence will be marked as achieved, and the second will be created.  

This delete job will consider all courses that ended at most a day ago, ensuring that users can't achieve the second step in the sequence anymore.

jobs/periodic/delete-unreachable-achievements/courses.ts
dhenkel92: We need all the user ids later in the process. I could use your suggestion, but that would mean making N additional queries, one for each course, to retrieve them.

jobs/list.ts
dhenkel92: Interesting idea 👍 
I would keep it separate tho, for observability reasons.

jobs/periodic/delete-unreachable-achievements/courses.ts
dhenkel92: That's a great idea. I didn't know about these functions. I'll add them.

jobs/periodic/delete-unreachable-achievements/courses.ts
```diff
@@ -0,0 +1,74 @@
...
+        const [relationType, subcourseId] = parseRelation(relation);
+        if (relationType !== EventRelationType.Subcourse) {
+            logger.warn('unexpected relation type', {
+                relation,
+            });
+            continue;
+        }
+
+        const subcourse = await getSubcourse(subcourseId, false);
+        if (subcourse.cancelled || !(await subcourseOver(subcourse))) {
```
dhenkel92, line 49: Unfortunately, `subcourseOver` needs the whole subcourse even tho it only uses the subcourse.id.

I thought about refactoring it a little, but I don't know about potential side effects.

jobs/periodic/delete-unreachable-achievements/courses.ts
Jonasdoubleyou: Ah, I see, maybe put as comment into the code given that this is non obvious? :)

jobs/periodic/delete-unreachable-achievements/courses.ts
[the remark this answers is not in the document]
dhenkel92: Good Catch!

PR #1196, opened 2025-02-12
feat: add SSO_USER Role

JeangelLF:
## Ticket

Part of https://github.com/corona-school/project-user/issues/1429

## What was done?

- Added a `SSO_USER` Role. Will be used to change the app behavior for these users _(Hide some functionalities / adjust how it works)_

### DIFF
--- a/common/user/evaluate_roles.ts
+++ b/common/user/evaluate_roles.ts
@@ -3,6 +3,7 @@ import { prisma } from '../prisma';
 import { Role } from '../user/roles';
 import { getLogger } from '../logger/logger';
 import { getStudent, getPupil, getScreener, User } from '.';
+import { isSSOUser } from '../secret/idp';
 
 const logger = getLogger('Roles');
 
@@ -21,6 +22,10 @@ export async function evaluateUserRoles(user: User): Promise<Role[]> {
     if (user.screenerId) {
         const screener = await getScreener(user);
         evaluateScreenerRoles(screener, roles);
+    }
+
+    if (await isSSOUser(user.userID)) {
+        roles.push(Role.SSO_USER);
     }
 
     return roles;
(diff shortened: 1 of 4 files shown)


common/user/evaluate_roles.ts
```diff
@@ -23,6 +24,10 @@ export async function evaluateUserRoles(user: User): Promise<Role[]> {
         evaluateScreenerRoles(screener, roles);
     }
 
+    if (await isSSOUser(user.userID)) {
```
Jonasdoubleyou, line 27: Do we refresh the user session on loginPassword + register after the SSO secret was added?

Jonasdoubleyou (approved):
Hm, "not really a role" for authorizations - but okay I guess if we need this in many places

common/user/evaluate_roles.ts
Jonasdoubleyou: No, we do it before - so in createIdpLogin you would habe to refresh the session / manually add the SSO_USER role there

graphql/authentication.ts
```diff
@@ -214,6 +214,10 @@ export class AuthenticationResolver {
             // Now that the user confirmed their identity, we can complete linking process
             if (isSSO) {
                 await createIDPLogin(user.userID, sessionUser.idpSub, sessionUser.idpClientId);
+                await userSessions.set(context.sessionToken, {
+                    ...context.user,
+                    roles: context.user.roles.concat(Role.SSO_USER),
+                });
```
Jonasdoubleyou, line 220: Same for loginPassword?

JeangelLF: Sorry, what do you mean?

JeangelLF: I must be blind today, isn't that what I just changed?
Like here we complete the linking process and then I'm manually setting the new `SSO_USER` role?

Jonasdoubleyou: Ah, sorry, stupid me - thought this was in registration - but there we call loginAsUser afterwards already - sorry for the confusion

Jonasdoubleyou (approved):

PR #1197, opened 2025-02-15
fix: Do not expose supported hours if referred users is less than 2

Jonasdoubleyou:
It's nice that this was considered in the UI, but is still exposed to the user at the API. If we implement 2-anonymity, let's do it all the way.


JeangelLF (approved):
NIce, thank you!

PR #1207, opened 2025-03-21
Hotfix: Expose Zoom Meeting Password

JeangelLF:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1460

## What was done?

- Added a resolver for the Zoom meeting password. This will be used on our Zoom component view on the frontend
- Included the Meeting password (if any) in the `zoomMeetingUrl` resolver

In the meantime, I changed the settings again to not require a password since this feature wasn't built with that in mind so I'm not sure what other things it could break. 👀.

All of this is pretty weird. The only thing I could find was this post https://devforum.zoom.us/t/passcode-wrong-error/59739, but it wasn't entirely clear why it happened

### DIFF
--- a/common/zoom/scheduled-meeting.ts
+++ b/common/zoom/scheduled-meeting.ts
@@ -33,6 +33,7 @@ export type ZoomMeeting = {
     settings: {
         alternative_hosts: string;
     };
+    encrypted_password?: string;
 };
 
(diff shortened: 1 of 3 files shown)


Jonasdoubleyou (commented):
LGTM - how did we create meetings with passwords?

PR #1206, opened 2025-03-17
Feat: Track appointments attendance

JeangelLF:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1456

## What was done?

- Add a new `joinedBy` column in the `lecture` table
- Update the `trackUserJoinAppointmentMeeting` function to save when a user joins an appointment

PR #1210, opened 2025-03-22
Feat: privacy-conscious appointment stats

realmayus:
Only save select Zoom data to our database.

This required some very hacky workaround because I wasn't sure how to identify pupils/students in the meeting report other than separating them by having an ID and not having an ID.

Closes https://github.com/corona-school/project-user/issues/1422

### DIFF
--- a/jobs/manual/populate_appointment_stats.ts
+++ b/jobs/manual/populate_appointment_stats.ts
@@ -1,0 +1,25 @@
+import { getLogger } from '../../common/logger/logger';
+import { prisma } from '../../common/prisma';
+import { saveAppointmentStats } from '../../common/appointment/create';
+
+const logger = getLogger();
+
+export async function populateAppointmentStats() {
+    logger.info('Populating appointment stats');
+    const zoomMeetingReports = await prisma.lecture.findMany({
+        where: {
+            zoomMeetingReport: { hasSome: true },
+        },
+        select: {
+            zoomMeetingReport: true,
+            id: true,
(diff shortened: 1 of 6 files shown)


JeangelLF (commented):
Oh... I think we tried to partially solve the same thing 🙈.
Last week, I added https://github.com/corona-school/backend/pull/1206 trying to avoid the zoom meeting report because of the issues you described
- Not being able to identify users accurately
- Not having any infos on non-zoom meetings

jobs/manual/populate_appointment_stats.ts
```diff
@@ -0,0 +1,25 @@
+import { getLogger } from '../../common/logger/logger';
+import { prisma } from '../../common/prisma';
+import { saveAppointmentStats } from '../../common/appointment/create';
+
+const logger = getLogger();
+
+export async function populateAppointmentStats() {
+    logger.info('Populating appointment stats');
+    const zoomMeetingReports = await prisma.lecture.findMany({
+        where: {
```
JeangelLF, line 10: Shouldn't we also exclude here those that already have the stats populated? 👀

realmayus:
@JeangelLF I see, thanks for the notice. So all that remains for closing https://github.com/corona-school/project-user/issues/1422 is the meeting duration, afaict. This I can derive from the meeting reports without any hacks.

I'm gonna remove the appointment_stats data structure then and just add the `actualDuration` to the `lecture` entity.

PR #1219, opened 2025-04-08
feat: Add status column to students screenings

JeangelLF:
## Ticket

The backend part of https://github.com/corona-school/project-user/issues/1464

## What was done?

- I added a new `status` column to the `screening` / `instructor_screening`, which is similar to how we do with the pupil screenings. This will allow for student screenings in pending or in missed statuses.
- Update mutations to start saving infos there. For now, only `success` / `rejection`

[Hier I wrote a query to populate this column for the existing screenings based on the `success` column](https://github.com/corona-school/dev-db/commit/2bacce1929203fcabf663371a529f40c98d844a5) (which will be deprecated/removed in future PRs)

### DIFF
--- a/common/student/screening.ts
+++ b/common/student/screening.ts
@@ -1,4 +1,4 @@
-import { student as Student, screener as Screener, Prisma, PrismaClient } from '@prisma/client';
+import { student as Student, screener as Screener, Prisma, PrismaClient, student_screening_status_enum as ScreeningStatus } from '@prisma/client';
 import { prisma } from '../prisma';
 import * as Notification from '../notification';
 import { getLogger } from '../../common/logger/logger';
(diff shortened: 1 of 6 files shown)


Jonasdoubleyou (approved):

Jonasdoubleyou:
Maybe already switch the logic in https://github.com/corona-school/backend/blob/597ed481212a70453ffea8e52e15f256f9e78150/common/student/screening.ts#L86 to use the new enum? Otherwise pending/missed = rejected ...

JeangelLF:
Sure, I'll do it in a separate PR. For now, there is no way to create missed/pending student screenings.
My next step is to replace the usage of success with this new column and then add a way to create pending/missed screenings

PR #1220, opened 2025-04-09
Feat: Replace success with status in student screenings

JeangelLF:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1464

## Description

- Added migration to drop the `success` column and use the new `status` column. In Prod this column already has the right values for all screenings / instructor_screenings


graphql/statistics/fields.ts
```diff
@@ -704,7 +704,7 @@ export class StatisticsResolver {
                                              date_part('month', "createdAt"::date) AS month,
                                              "knowsCoronaSchoolFrom"               AS group
                                       FROM "screening"
-                                      WHERE "success" = TRUE
+                                      WHERE "status" = '1'
```
Jonasdoubleyou, line 707: Cant we use the enum here?

Jonasdoubleyou (approved):

PR #1221, opened 2025-04-14
Feat: send ICS calendar files as attachments

realmayus:
# Generate ICS files from appointments and attach them to emails
Closes https://github.com/corona-school/project-user/issues/1462.
- Introduce function for generating ICS file for group appointments and match appointments, for both instructors and participants
- Send out ICS files as notification attachments, in both initial appointment creation emails (participants only) and reminder emails sent 12 hours before appointment start (both instructors and participants)

## ICS properties:
```
            title: displayName,
            description: appointment.description ?? undefined,
            url: `https://app.lern-fair.de/appointment/${appointment.id}`,
            location: `https://app.lern-fair.de/appointment/${appointment.id}`,
            alarms: [
                {
                    action: 'display',
                    trigger: { hours: 1, minutes: 0, before: true },
                    description: `Erinnerung: ${displayName}`,
                },
            ],
            busyStatus: 'BUSY',
            transp: 'OPAQUE',

```
Some notes: 
- We use both `location` and `url` for the link. While `url` seems more fitting according to the iCal spec (see 4.8.4.6 of [RFC2445](https://www.ietf.org/rfc/rfc2445.txt)), calendar clients seem to disregard this entirely, whereas the `location` field supports URLs in most cases
- We can consider using the zoom link directly instead of the appointment link. I'm in favor of putting the appointment link here, as this might show more information (cancelled/other participants), but it's another indirection for the users.
- Alarms are not respected by Google Calendar, they are respected by other clients, including the native iOS mail app.
- I left out the `organizer` field on purpose, as gmail completely omits the event preview if you only put the organizer's "common name" there and leave out the email address (which we don't share). This shows "Unknown Organizer" on the Gmail web app.

## Previews in email clients
Gmail:
Screenshot 2025-04-16 at 13 32 11

Native iOS Mail app:

### DIFF
--- a/common/appointment/create.ts
+++ b/common/appointment/create.ts
@@ -12,9 +12,10 @@ import * as Notification from '../../common/notification';
 import { getStudentsFromList, User, userForPupil, userForStudent } from '../user';
 import { getMatch, getPupil, getStudent } from '../../graphql/util';
 import { PrerequisiteError, RedundantError } from '../../common/util/error';
-import { getContextForGroupAppointmentReminder, getContextForMatchAppointmentReminder } from './util';
+import { getContextForGroupAppointmentReminder, getContextForMatchAppointmentReminder, getIcsFile } from './util';
 import { getNotificationContextForSubcourse } from '../../common/courses/notifications';
 import { assertAllowed, Decision } from '../util/decision';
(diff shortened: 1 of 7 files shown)


JeangelLF (approved):
Woah this looks really nice! Thanks! 🙌

PR #1223, opened 2025-04-22
Fix logic in matching getViableUsers for test.*@lern-fair.de accounts

realmayus:
It turns out that if you specify multiple `NestedStringFilter`s inside the `not` filter, they get ORed instead of ANDed.

Previously we had `viable = !(startsWith("test") || endsWith("@lern-fair.de"))` = `!startsWith("test") && !endsWith("@lern-fair.de")`
We want: `viable = !(startsWith("test") && endsWith("@lern-fair.de")` = `!startsWith("test") || !endsWith("@lern-fair.de")`.

### DIFF
--- a/common/match/pool.ts
+++ b/common/match/pool.ts
@@ -55,7 +55,7 @@ const getViableUsers = (toggles: string[]) => {
     /* On production we want to avoid that our testusers test+prod-...@lern-fair.de
     are accidentally matched to real users */
     if (!isDev) {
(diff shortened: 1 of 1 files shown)


Jonasdoubleyou (approved):
Wow, thats surprising ... I hope we didnt have users with their email starting with test...

Jonasdoubleyou:
Do we use not: elsewhere? Might be worth checking

realmayus:
> Do we use not: elsewhere? Might be worth checking

Doesn't seem like it, all other instances of `not: ` are followed by just one filter

PR #1226, opened 2025-05-19
feat: Add subcourse grade range in subcourse context

JeangelLF:
## What was done?

- Added grade range in the subcourse context to be used in email templates

### DIFF
--- a/common/courses/notifications.ts
+++ b/common/courses/notifications.ts
@@ -5,7 +5,7 @@ import { userForPupil } from '../user';
 import * as Prisma from '@prisma/client';
 import { getFirstLecture } from './lectures';
 import { parseSubjectString } from '../util/subjectsutils';
-import { getCourseCapacity, getCourseImageURL, getGradeLabel } from './util';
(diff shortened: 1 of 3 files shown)


common/courses/notifications.ts
```diff
@@ -58,6 +58,8 @@ export async function getNotificationContextForSubcourse(course: { name: string;
         subcourse: {
             url: `https://app.lern-fair.de/single-course/${subcourse.id}`,
             id: '' + subcourse.id,
+            minGrade: getGradeLabel(subcourse.minGrade),
+            maxGrade: getGradeLabel(subcourse.maxGrade),
```
Jonasdoubleyou, line 62: So we'll show "10. Klasse - in Ausbildung", "in Ausbildung - in Ausbildung", "10. Klasse - 10. Klasse"?

Wouldn't it make more sense to directly translate the range in the backend, handling min == max + max == 14?

Jonasdoubleyou (approved):

common/courses/notifications.ts
JeangelLF: Yeah, sure, that makes way more sense! I've updated the PR accordingly.

PR #1229, opened 2025-05-20
Feat: Do not send notifications for Homeworkhelp course

JeangelLF:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1480

## What was done?

- Turned off appointment notifications for the Homeworkhelp course

### DIFF
--- a/common/appointment/create.ts
+++ b/common/appointment/create.ts
@@ -144,7 +144,12 @@ export function canCreateGroupAppointment(subcourse: Subcourse, hasInstructors: 
     return { allowed: true };
 }
 
-export const createGroupAppointments = async (subcourseId: number, appointmentsToBeCreated: AppointmentCreateGroupInput[], organizer: Student) => {
+export const createGroupAppointments = async (
+    subcourseId: number,
+    appointmentsToBeCreated: AppointmentCreateGroupInput[],
+    organizer: Student,
+    silent?: boolean
+) => {
     const participants = await prisma.subcourse_participants_pupil.findMany({ where: { subcourseId: subcourseId }, select: { pupil: true } });
     const mentors = await prisma.subcourse_mentors_student.findMany({ where: { subcourseId: subcourseId }, select: { student: true } });
     const instructors = await prisma.subcourse_instructors_student.findMany({ where: { subcourseId: subcourseId }, select: { student: true } });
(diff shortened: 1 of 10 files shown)


graphql/appointment/mutations.ts
```diff
@@ -113,7 +117,9 @@ export class MutateAppointmentResolver {
     async appointmentDecline(@Ctx() context: GraphQLContext, @Arg('appointmentId') appointmentId: number) {
         const appointment = await getLecture(appointmentId);
         await hasAccess(context, 'Lecture', appointment);
-        await declineAppointment(context.user, appointment);
+        const subcourse = await prisma.subcourse.findUnique({ where: { id: appointment.subcourseId }, include: { course: true } });
```
Jonasdoubleyou, line 120: What about match appointments? Shouldn't this only be silent if appointment type == course?

graphql/appointment/mutations.ts
```diff
@@ -123,8 +129,9 @@ export class MutateAppointmentResolver {
     async appointmentCancel(@Ctx() context: GraphQLContext, @Arg('appointmentId') appointmentId: number) {
         const appointment = await getLecture(appointmentId);
         await hasAccess(context, 'Lecture', appointment);
-
-        await cancelAppointment(context.user, appointment, false);
+        const subcourse = await prisma.subcourse.findUnique({ where: { id: appointment.subcourseId }, include: { course: true } });
```
Jonasdoubleyou, line 132: same here

graphql/appointment/mutations.ts
```diff
@@ -103,7 +105,9 @@ export class MutateAppointmentResolver {
     async appointmentUpdate(@Ctx() context: GraphQLContext, @Arg('appointmentToBeUpdated') appointmentToBeUpdated: AppointmentUpdateInput) {
         const appointment = await getLecture(appointmentToBeUpdated.id);
         await hasAccess(context, 'Lecture', appointment);
-        await updateAppointment(context.user, appointment, appointmentToBeUpdated);
+        const subcourse = await prisma.subcourse.findUnique({ where: { id: appointment.subcourseId }, include: { course: true } });
```
Jonasdoubleyou, line 108: same here

common/appointment/create.ts
```diff
@@ -188,34 +193,35 @@ export const createGroupAppointments = async (subcourseId: number, appointmentsT
...
-    }
 
-    for (const instructor of instructors) {
-        for (const appointment of createdGroupAppointments) {
-            if (subcourse.published) {
-                // For unpublished courses, this is deferred to a later point
-                await Notification.actionTakenAt(new Date(appointment.start), userForStudent(instructor.student), 'student_group_appointment_starts', {
+            // Send out reminders 12 hours before the appointment start
+            for (const appointment of createdGroupAppointments) {
+                await Notification.actionTakenAt(new Date(appointment.start), userForPupil(participant.pupil), 'pupil_group_appointment_starts', {
```
Jonasdoubleyou, line 207: Do we allow changing the course type? If yes this could cause a mess one day if a course is created, adding all the reminders and then not cancelling them properly after the course type was changed.
Maybe we should prohibit changing it?

Jonasdoubleyou (commented):
How about a method `isSubcourseSilent`? To me it looks as if the logic might change in the future, and then there is one place to change ...

common/appointment/create.ts
JeangelLF: In the case of Homeworkhelp courses, students are not allowed to create this course type. 
I'll update the edit mutation to also prevent students from changing an existing course to Homeworkhelp via the API _(This isn't possible through the UI, but it seems to be allowed via the API.)_

Jonasdoubleyou (approved):

PR #1231, opened 2025-05-22
Feat: Add calendly webhook

JeangelLF:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1366 (backend part)

## What was done?

- Added a new endpoint for our Calendly webhook `/calendly/event`
- Added handlers for two events
  - `onEventInviteeCreated`: It creates pending screenings for the user (Pupil or Student) who scheduled the meeting via Calendly. We also save some of this information in a new `screening_appointment` table
  - `onEventInviteeCanceled`: For now, it marks the appointment as cancelled so the user in the UI can book another one. Additionally, it writes some logs on the screenings (like the reason for the cancellation).
  - Expose screening appointment information that we'll show on the Screening blockers.

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -1004,7 +1004,10 @@ model screening_appointment {
   startAt               DateTime              @db.Timestamp(6)
   cancelledAt           DateTime?             @db.Timestamp(6)
   cancellationReason    String?               @db.VarChar
+  // Where the appointment takes place, e.g. 'Zoom', 'Google Meets', etc...
   joinUrl               String?               @db.VarChar
+  // Calendly Event URL, we can use this go get additional information about the appointment
+  // For calendly, it's also kind of an id
   eventUrl              String?               @db.VarChar
   pupilScreeningId      Int?
   pupilScreening        pupil_screening?      @relation(fields: [pupilScreeningId], references: [id], onDelete: NoAction, onUpdate: NoAction)
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -277,6 +277,7 @@ model instructor_screening {
   studentId             Int?                          @unique(map: "REL_e176665fa769d2e603d825f6fa")
   student               student?                      @relation(fields: [studentId], references: [id], onDelete: NoAction, onUpdate: NoAction, map: "FK_e176665fa769d2e603d825f6fa3")
   screener              screener?                     @relation(fields: [screenerId], references: [id], onDelete: NoAction, onUpdate: NoAction, map: "FK_ef1d3e862feda89b92fddcdbb34")
+  screening_appointment screening_appointment[]
 }
 
 // DEPRECATED: We once had a cooperation with Jugend Forscht ("jufo")
@@ -724,6 +725,7 @@ model screening {
   studentId             Int?                          @unique(map: "REL_dfb78fd7887c69e3c52e002083")
   screener              screener?                     @relation(fields: [screenerId], references: [id], onDelete: NoAction, onUpdate: NoAction, map: "FK_c0b20c6342ac95d3b66c31ac30e")
   student               student?                      @relation(fields: [studentId], references: [id], onDelete: NoAction, onUpdate: NoAction, map: "FK_dfb78fd7887c69e3c52e0020838")
+  screening_appointment screening_appointment[]
 }
 
 model student {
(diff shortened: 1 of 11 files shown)


web/controllers/calendlyController/index.ts
```diff
@@ -0,0 +1,61 @@
...
+    logger.info('Request at /calendly/event');
+    try {
+        const calendlySignature = req.get('Calendly-Webhook-Signature');
+        if (!calendlySignature) {
+            logger.error('Invalid Signature');
+            res.status(401).send({ error: 'Unauthorized' });
+            return;
+        }
+        // Calendly Signatures documentation https://developer.calendly.com/api-docs/4c305798a61d3-webhook-signatures
+        const { t, signature } = calendlySignature.split(',').reduce(
```
Jonasdoubleyou, line 22: would find `Object.fromEntries(calendlySignature.split(",").map(kv => kv.split("="))` more elegant ;)

web/controllers/calendlyController/index.ts
```diff
@@ -0,0 +1,61 @@
...
+                t: '',
+                signature: '',
+            }
+        );
+        if (!t || !signature) {
+            logger.error('Invalid Signature');
+            res.status(401).send({ error: 'Unauthorized' });
+            return;
+        }
+        const data = t + '.' + JSON.stringify(req.body);
```
Jonasdoubleyou, line 46: > https://datatracker.ietf.org/doc/html/rfc7519#section-3
> "This JSON object MAY contain
   whitespace and/or line breaks before or after any JSON values or
   structural characters"

Not sure if JSON.stringify(req.body) is always exactly the same as the original body sent with the request (and used to compute the signature.

While I am all for building things on our own and not throwing libraries on it, in this case maybe we should use a library like https://github.com/auth0/node-oauth2-jwt-bearer/tree/main/packages/express-oauth2-jwt-bearer?

prisma/schema.prisma
```diff
@@ -984,11 +986,35 @@ model pupil_screening {
...
   // so a COUNT(*) GROUP BY should work relatively well:
-  knowsCoronaSchoolFrom String? @db.VarChar
+  knowsCoronaSchoolFrom String?                 @db.VarChar
+  screening_appointment screening_appointment[]
+}
+
+// Users need to go through a screening process before they can use the app. 
+// For this they have to book an appointment in Calendly.
+// The appointment is then automatically stored here via a webhook.
+model screening_appointment {
```
Jonasdoubleyou, line 1000: Hmm, I wonder whether it is easier to reuse the generic "appointment"? We already have generic participants / hosts, cancellation handling and joinUrl, ... the only thing missing would be eventUrl and the relationship to screening. For the eventUrl I guess it would be fine to add it to appointment, there might be other cases where we want to offer scheduling via Calendly? ... And regarding screening, wouldn't it be enough to search all open screenings of all participants?

Jonasdoubleyou: Advantage would be that we could recycle some UI in the frontend?
But if you see good reasons to keep this separate I am also fine with the current approach.

Jonasdoubleyou (commented):
Uh nice, next big integration :)

prisma/schema.prisma
Jonasdoubleyou: I guess maybe for Screeners the "appointments" tab might just start working, AFAIR we never hid it ...

web/controllers/calendlyController/index.ts
JeangelLF: Thanks for the hint!
I'm not entirely sure how to apply that library in this case 👀 . That said, I noticed we might already have something similar in place to what we need. In our [server file](https://github.com/corona-school/backend/blob/c8e1cd2724edec7482faecb8bd70095616914b9f/web/server.ts#L52), we're extracting the rawBody, and it's already available on the request object. We're also using it for something quite similar in the [chat controller](https://github.com/corona-school/backend/blob/b5a7fecea65ee311cadeeac2a45f2b7a6f654e55/web/controllers/chatNotificationController/index.ts#L24).

What do you think?

prisma/schema.prisma
JeangelLF: Thank you for the feedback!

I actually received similar feedback from Marius as well. To be honest, I didn't think a lot about using the `lecture` table. Not sure if just because of the name that I immediately associated it with courses/learning partners 😅.

So, I just tried it and it was pretty smooth _(though I need to test it a bit better)_, I'll create a separate PR for it

PR #1232, opened 2025-05-25
Feat: Add calendly webhook (lecture version)

JeangelLF:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1366 (backend part)

## What was done?

- Added a new endpoint for our Calendly webhook `/calendly/event`
- Added handlers for two events
  - `onEventInviteeCreated`: It creates pending screenings for the user (Pupil or Student) who scheduled the meeting via Calendly. We save the appointment information in the `lecture` table with a new `screening` type
  - `onEventInviteeCanceled`: For now, it marks the appointment as cancelled so the user in the UI can book another one. Additionally, it writes some logs on the screenings (like the reason for the cancellation).
  - Expose screening appointment information that we'll show on the Screening blockers.


JeangelLF:
@Jonasdoubleyou Here is the implementation of the [Calendly webhook PR](https://github.com/corona-school/backend/pull/1231) but using the existing appointments table. In this PR, I'll only cover the webhook functionality to automatically create screenings/appointments. Let me know what you think.

In a following PR I'll add whatever is needed for the Screening appointments tab to work :)

common/calendly/index.ts
```diff
@@ -0,0 +1,296 @@
...
+    });
+    if (!appointment) {
+        logger.warn(`No ScreeningAppointment found for event URL: ${event.payload.scheduled_event.uri}`);
+        return;
+    }
+
+    await prisma.lecture.update({
+        where: { id: appointment.id },
+        data: {
+            isCanceled: true,
```
Jonasdoubleyou, line 222: In this case Calendly sends mails to the user so we dont need notifications?

Jonasdoubleyou (approved):
Looks good!

common/calendly/index.ts
JeangelLF: Yes, that's correct.

graphql/appointment/fields.ts
```diff
@@ -131,6 +143,15 @@ export class ExtendedFieldsLectureResolver {
...
+        if (appointment.pupilScreeningId) {
+            return await prisma.lecture.count({ where: { pupilScreeningId: appointment.pupilScreeningId, isCanceled: false } });
+        }
+        if (appointment.instructorScreeningId) {
+            return await prisma.lecture.count({ where: { instructorScreeningId: appointment.instructorScreeningId, isCanceled: false } });
+        }
+        if (appointment.tutorScreeningId) {
+            return await prisma.lecture.count({ where: { tutorScreeningId: appointment.tutorScreeningId, isCanceled: false } });
+        }
         throw new Error('Cannot determine total of loose appointment');
```
Jonasdoubleyou, line 155: I guess return 1; would also work here and be a good default, same for position. Dont think we need to implement it for all appointment types ...

graphql/appointment/fields.ts
```diff
@@ -207,6 +228,20 @@ export class ExtendedFieldsLectureResolver {
 
         return await getSubcourse(appointment.subcourseId);
     }
+
+    @FieldResolver((returns) => AppointmentActionsUrls, { nullable: true })
+    @Authorized(Role.APPOINTMENT_PARTICIPANT, Role.ADMIN)
+    async actionUrls(@Ctx() context: GraphQLContext, @Root() appointment: Appointment) {
```
Jonasdoubleyou, line 234: I wonder whether it would be easier if the backend calls these in appointmentEdit/Cancel?

Jonasdoubleyou (dismissed):

graphql/appointment/fields.ts
JeangelLF: Oh, those are client links that send the user to Calendly to complete the reschedule/cancellation process. We later get all the info with the webhooks.

PR #1237, opened 2025-06-02
fix: Allow Subcourse instructors to always add pupils to their courses

Jonasdoubleyou:
Previously this was "strict", i.e. enforcing course settings, correct grade, etc. As course instructors easily can weaken these settings to add pupils, it does not make sense to apply these restrictions to them. Now instructors can add pupils in the same way admins can (as long as they are on the waiting list or prospects).


JeangelLF (approved):
Nice, thank you!

PR #1238, opened 2025-06-02
fix: SSO_REGISTERING_USER is no longer its OWNER

Jonasdoubleyou:
Previously SSO_REGISTERING_USER (in case of link / register) would not include the USER role, however it would imply the OWNER role on the user entity. This means that the many resolvers guarded by OWNER could be run by such a temporary user. I am not certain that this correctly fails in all cases (in particular this runs into getUser(user.userID), which fails for a mock user-ID the sso user has), thus I entirely removed the OWNER role from such users by default, adding the TEMPORARY_OWNER role to allow the few cases where we want to allow this, i.e. the parts that are in the GetUser query the User-App uses globally

https://github.com/corona-school/project-user/issues/1488

### DIFF
--- a/common/user/roles.ts
+++ b/common/user/roles.ts
@@ -21,6 +21,9 @@ export enum Role {
     UNAUTHENTICATED = 'UNAUTHENTICATED',
     /* User owns the entity as defined in graphql/ownership */
     OWNER = 'OWNER',
+    /* Same as OWNER, but allows temporary users (i.e. without the USER Role) to hold it */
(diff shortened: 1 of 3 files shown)


Jonasdoubleyou:
@JeangelLF I am not a 100% sure whether I broke SSO login - as I do not have a local setup, I'd merge + test on staging ...

JeangelLF (approved):
It's looking good! I also tested it locally, and it's working fine :)

PR #1260, opened 2025-07-06
Feat/student zip code

Fabian5150:
## What was done?
- Added zipCode field for students and gave students and screeners write and read access to it


Fabian5150:
Integration test only fails because I added a new column to the student. (and directly changed it's type because I made a mistake...)
But I think it should not be a problem

Jonasdoubleyou:
Mmm, please delete the migrations and rerun db:create-migration to create a single migration. Then also the check should not fail.

Jonasdoubleyou (approved):

PR #1267, opened 2025-08-15
Feat: New registration adjustments

JeangelLF:
## Ticket

Backend part of https://github.com/corona-school/project-user/issues/1506

## What was done?

- Added an `age` column on the pupil model
- Added the new languages we want to show on the new Languages step
- Updated gradeAsInt to be nullable _(It's now asked after registration, so this prevents an error when fetching the user data)_

### DIFF
--- a/common/notification/actions.ts
+++ b/common/notification/actions.ts
@@ -100,6 +100,10 @@ const _notificationActions = {
         description: 'Pupil / Screening was rejected',
         sampleContext: {},
     },
+    pupil_screening_with_active_match_rejected: {
+        description: 'Pupil / Screening was rejected while having an active match',
+        sampleContext: {},
+    },
     pupil_screening_succeeded: {
         description: 'Pupil / Screening was successful',
         sampleContext: {},
--- a/common/pupil/screening.ts
+++ b/common/pupil/screening.ts
@@ -69,7 +69,8 @@ export async function updatePupilScreening(screener: Screener, pupilScreeningId:
             if (isFirstScreening) {
                 await Notification.actionTaken(asUser, 'pupil_screening_after_registration_rejected', {});
             } else {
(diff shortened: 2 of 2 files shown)


graphql/pupil/fields.ts
```diff
@@ -137,7 +137,7 @@ export class ExtendFieldsPupilResolver {
         });
     }
 
-    @FieldResolver((type) => Int)
+    @FieldResolver((type) => Int, { nullable: true })
     @Authorized(Role.ADMIN, Role.SCREENER, Role.OWNER, Role.TUTOR, Role.INSTRUCTOR)
     gradeAsInt(@Root() pupil: Required<Pupil>) {
         return gradeAsInt(pupil.grade);
```
Jonasdoubleyou, line 143: By the way, there are two other places in the code where null handling seems to be missing:
- https://github.com/corona-school/backend/blob/a026515beb9dabf034ae1b189c9d3c2ae6869608/common/match/matching.ts#L37
- https://github.com/corona-school/backend/blob/master/common/courses/participants.ts#L194

Jonasdoubleyou (approved):

graphql/pupil/fields.ts
JeangelLF: Shouldn't this work the same in this case? The gradeAsInt was already returning null when there wasn't a grade set

Jonasdoubleyou: Fair, I guess we can ignore it for now ... was just a "while at it" thing

common/daz/language.ts
```diff
@@ -1,5 +1,6 @@
 export enum Language {
     sq = 'Albanisch',
+    am = 'Amharisch',
```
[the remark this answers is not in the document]
Jonasdoubleyou: Well, lets see if some of our users speak them :)

PR #1276, opened 2025-08-28
Hotfix: Zoom mess

JeangelLF:
## Ticket

 Resolves https://github.com/corona-school/project-user/issues/1522

## What was done?

- Updated the `updateZoomMeeting` so it doesn't break all the Zoom appointments for a course after adding/removing an instructor.
- Fix the reschedule functionality so it actually reschedules the meeting in Zoom and it doesn't get deleted by them

### DIFF
--- a/common/appointment/update.ts
+++ b/common/appointment/update.ts
@@ -115,7 +115,7 @@ export async function updateAppointment(
         }
     }
     const zoomUpdate = {
-        start: newStart,
+        startTime: newStart,
         duration: newDuration,
         endDate: lastDate,
     };
--- a/common/zoom/scheduled-meeting.ts
+++ b/common/zoom/scheduled-meeting.ts
@@ -210,21 +210,36 @@ const getZoomMeetingReport = async (meetingId: string) => {
     return response.json();
 };
 
+interface PatchZoomMeetingBody {
+    start_time?: string;
+    duration?: number;
+    timezone?: string;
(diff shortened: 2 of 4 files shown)


common/appointment/update.ts
```diff
@@ -115,7 +115,7 @@ export async function updateAppointment(
         }
     }
     const zoomUpdate = {
-        start: newStart,
+        startTime: newStart,
```
Jonasdoubleyou, line 118: Huh, why didn't this even compile?

Jonasdoubleyou (approved):

common/appointment/update.ts
Jonasdoubleyou: Aah cause the type is not specified and startTime is optional and objects can have additional props :/

Jonasdoubleyou:
This is slightly embarrassing, did this ever work? This could explain some of the strange zoom bugs we had in the past

Jonasdoubleyou (approved):

JeangelLF:
> did this ever work? This could explain some of the strange zoom bugs we had in the past

I think this never worked as expected. It was really hard to figure out that something was wrong because our data was always fine. But on the Zoom side, everything was a mess and because of that we faced some really weird bugs...
 

> This is slightly embarrassing

Yeah ... exactly my feeling right now 🫠

PR #1280, opened 2025-09-05
fix: Cancel Subcourses correctly when deactivating a student

JeangelLF:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1526

## What was done?

- Use `cancelSubcourse` when a student is deactivated, so we also cancel appointments _(and its reminders for the participants)_

### DIFF
--- a/common/student/activation.ts
+++ b/common/student/activation.ts
@@ -10,6 +10,7 @@ import { isZoomFeatureActive } from '../zoom/util';
 import { userForStudent } from '../user';
 import { CertificateState } from '../certificate';
 import { removeAllPushSubcriptions } from '../notification/channels/push';
+import { cancelSubcourse } from '../courses/states';
 
 export async function deactivateStudent(
     student: Student,
@@ -57,6 +58,7 @@ export async function deactivateStudent(
         },
         include: {
             course_instructors_student: true,
+            subcourse: true,
         },
     });
 
@@ -80,18 +82,11 @@ export async function deactivateStudent(
                     id: courses[i].id,
                 },
                 data: {
-                    subcourse: {
-                        updateMany: {
-                            where: {},
-                            data: {
-                                cancelled: true,
-                            },
-                        },
-                    },
                     courseState: course_coursestate_enum.cancelled,
                 },
             });
-            // TODO Notify participants
+            const cancellations = courses[i].subcourse.map((subcourse) => cancelSubcourse(userForStudent(student), subcourse));
+            await Promise.all(cancellations);
         }
     }
 


common/student/activation.ts
```diff
@@ -57,6 +58,7 @@ export async function deactivateStudent(
         },
         include: {
             course_instructors_student: true,
+            subcourse: true,
```
Jonasdoubleyou, line 61: Shouldn't we query subcourses here instead? We actually care about "subcourse instructors" not "course instructors" - this code still assumes there is a one to one mapping between them ...

Jonasdoubleyou (dismissed):

common/student/activation.ts
```diff
@@ -46,66 +52,34 @@ export async function deactivateStudent(
...
-                    },
-                    courseState: course_coursestate_enum.cancelled,
-                },
-            });
-            // TODO Notify participants
+    for (const subcourse of subcourses) {
+        // There are multiple instructors, so just remove the student from the subcourse
+        if (subcourse.subcourse_instructors_student.length > 1) {
+            await deleteSubcourseInstructor(userForStudent(student), subcourse, student);
+        } else if (!(await subcourseOver(subcourse))) {
```
Jonasdoubleyou, line 74: If the course is over, we also do not want to remove the instructor if there are multiple instructors?

Jonasdoubleyou (dismissed):

common/student/activation.ts
JeangelLF: Not sure if it's because this PR is from last year 🫣 but I think your comment should be now covered?

Jonasdoubleyou: No?

Jonasdoubleyou: What is the point in updating courses that are over?

common/student/activation.ts
```diff
@@ -51,66 +58,49 @@ export async function deactivateStudent(
...
+        if (subcourse.subcourse_instructors_student.length > 1) {
+            await removeSubcourseInstructor(userForStudent(student), subcourse, student);
+        } else if (!(await subcourseOver(subcourse))) {
+            // there is only one instructor and the subcourse is not over yet, so cancel the subcourse
+            await cancelSubcourse(userForStudent(student), subcourse, true);
         }
     }
 
+    // Remove the student from any courses where they were mentors
+    const mentoredSubcourses = await prisma.subcourse.findMany({
```
Jonasdoubleyou, line 92: same here

common/student/activation.ts
```diff
@@ -32,7 +34,12 @@ export async function deactivateStudent(
 
     await Notification.cancelRemindersFor(userForStudent(student));
     // Setting 'active' to false will not send out any notifications during deactivation
-    student.active = false;
```
Jonasdoubleyou, line 35: Hm, are you sure there isn't any code further down that reads this?

common/student/activation.ts
JeangelLF: Ah I see, yeah you're right. That doesn't make much sense. 
I updated it to only update active subcourses

common/student/activation.ts
JeangelLF: I updated it to use everywhere the deactivated student instead of the one we get from the arguments

Jonasdoubleyou (approved):

PR #1283, opened 2025-09-24
feat: Prevent impersonation to update lastLogin

JeangelLF:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1536

## What was done?

Added a validation to prevent impersonated users to update the `lastLogin` field

### DIFF
--- a/common/secret/token.ts
+++ b/common/secret/token.ts
@@ -199,7 +199,7 @@ export async function verifyEmail(user: User) {
 }
 
 export function isImpersonationToken(tokenDescription: string) {
-    return tokenDescription.startsWith('Support');
+    return tokenDescription === SUPPORT_MARKER;
 }
 
 export const SUPPORT_MARKER = 'Support Week Access';


graphql/authentication.ts
```diff
@@ -231,7 +233,8 @@ export class AuthenticationResolver {
     async loginToken(@Ctx() context: GraphQLContext, @Arg('token') token: string, @Arg('deviceId', { nullable: true }) deviceId: string | null) {
         try {
             const user = await loginToken(token, deviceId);
-            await loginAsUser(user, context, deviceId);
+            const isImpersonation = await isImpersonationToken(token);
```
Jonasdoubleyou, line 236: Can't we do the check in loginToken above? That avoids doing two db roundtrips for every user login

Jonasdoubleyou: In loginToken, we retrieve the secret anyways including the description, so checking the description is for free

common/secret/token.ts
```diff
@@ -196,3 +196,14 @@ export async function verifyEmail(user: User) {
...
+
+export async function isImpersonationToken(token: string) {
+    const secret = await prisma.secret.findFirst({
+        where: {
+            secret: hashToken(token),
+            type: { in: [SecretType.TOKEN] },
+            OR: [{ expiresAt: null }, { expiresAt: { gte: new Date() } }],
+        },
+    });
+    return /^Support.*Week Access$/.test(secret?.description);
```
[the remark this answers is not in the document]
JeangelLF: We don't send a description when creating these tokens from Retool or from the App. So yeah, I think we can just do it like that. 👍

PR #1389, opened 2026-04-23
Feat: Abbreviate lastnames sus

JeangelLF:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1513

## What was done?

- Added a function to abbreviate pupils last names based on who is requesting it.


common/pupil/index.ts
```diff
@@ -16,3 +19,13 @@ export async function getParticipants(where: Prisma.pupilWhereInput) {
         gradeAsInt: gradeAsInt(pupil.grade),
     }));
 }
+
+export function normalizeLastName(pupil: { id: number; lastname: string; age: number }, context: GraphQLContext) {
+    const isOwnerOrElevated =
+        isElevated(context) || context.user.roles.includes(Role.TEMPORARY_OWNER) || (context.user.pupilId && context.user.pupilId == pupil.id);
```
Jonasdoubleyou: What is the purpose of includes(TEMPORARY_OWNER)? IIRC this is only used in the annotation but never stored in the user.roles and then resolved in authorization.ts (similar to owner)?

common/user/index.ts
```diff
@@ -23,6 +23,8 @@ export type User = {
     lastLogin: Date;
 
     pupilId?: number;
+    age?: number;
```
Jonasdoubleyou: not used?

graphql/user/fields.ts
```diff
@@ -67,8 +69,12 @@ export class UserFieldsResolver {
 
     @FieldResolver((returns) => String)
     @Authorized(Role.USER, Role.ADMIN, Role.TEMPORARY_OWNER)
-    lastname(@Root() user: User): string {
-        return user.lastname;
+    async lastname(@Root() user: User, @Ctx() context: GraphQLContext): Promise<string> {
+        if (!user.pupilId) {
+            return user.lastname;
+        }
+        const pupil = await getPupil(user);
```
Jonasdoubleyou: I wonder wether we can avoid the getPupil call here? Thinking of the user search, there we might return many users and then here would run hundreds of queries just to fetch the age ... 

I see two alternatives:
- Restrict this resolver from USER to OWNER, and return the full lastname here - I think we do not use user.lastname anywhere were users can see other users?
- Store age in user to avoid the fetch here?

Jonasdoubleyou (dismissed):

graphql/user/fields.ts
Jonasdoubleyou: ((1) and SCREENER and so on)

common/pupil/index.ts
JeangelLF: Oh, I meant to use `OWNER` but now that I think about that twice it doesn't make sense and checking if the context user is the same, should probably be enough.

graphql/user/fields.ts
JeangelLF: You're right... I reverted this change, it's true that we don't use `user.lastname` anywhere and this may not be needed here. I also adjusted the authorized roles

Jonasdoubleyou (approved):

PR #1422, opened 2026-06-19
Feat: Zoom Webhook

JeangelLF:
## What was done?

- Added a new webhook to listen to zoom events `meeting.ended` and `meeting.participant_joined` to collect more accurate information about the lectures.
- Added a new `actualDuration` column to the `lecture` table to store the time that at least two users spent in the meeting. I thought about re-using our `duration` in this case but it wasn't that clean because we use it as a "scheduledDuration" to calculate when we expect the meeting to end ... but it might be longer / shorter than that.
- Track more reliably when a user joins an appointment. So we also cover the cases where the student sends the zoom link to the pupil via chat, whatsapp _(this one shouldn't even be a use case but whatever)_, etc...

For now we only care about match appointments as this is a more recurring issue/need with these appointments.

### DIFF
--- a/common/zoom/webhook.ts
+++ b/common/zoom/webhook.ts
@@ -1,0 +1,161 @@
+import { lecture as Appointment } from '@prisma/client';
+import { trackUserJoinAppointmentMeeting } from '../appointment/tracking';
+import { getLogger } from '../logger/logger';
+import { prisma } from '../prisma';
+import { getUser } from '../user';
+import { getZoomMeetingReport } from './scheduled-meeting';
+
+interface ZoomMeetingEndedEvent {
+    event: 'meeting.ended';
+    event_ts: number;
+    payload: {
+        account_id: string;
+        object: {
+            id: string;
+            uuid: string;
+            host_id: string;
+            host_email: string;
+            topic: string;
+            type: number;
+            start_time: string;
+            timezone: string;
+            duration: number;
+            end_time: string;
+        };
+    };
+}
+
+interface ZoomMeetingParticipantJoinedEvent {
+    event: 'meeting.participant_joined';
+    event_ts: number;
+    payload: {
+        account_id: string;
+        object: {
+            id: string;
+            uuid: string;
+            host_id: string;
+            topic: string;
+            type: number;
+            start_time: string;
+            timezone: string;
+            duration: number;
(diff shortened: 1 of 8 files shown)


common/zoom/webhook.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    if (!sharedTime) {
+        return;
+    }
+    await prisma.lecture.update({ where: { id: appointment.id }, data: { actualDuration: { increment: Math.ceil(sharedTime / 60) } } });
+};
+
+const onEventMeetingParticipantJoined = async (event: ZoomMeetingParticipantJoinedEvent, appointment: Appointment) => {
+    const pupilId = appointment.participantIds.find((id) => id.startsWith('pupil/'));
+    const studentId = appointment.organizerIds.find((id) => id.startsWith('student/'));
```
Jonasdoubleyou, line 149: (only true if there is only one pupil and one student in a match appointment - but that should usually be the case)

prisma/schema.prisma
```diff
@@ -377,7 +377,10 @@ model lecture {
   createdAt             DateTime                     @default(now()) @db.Timestamp(6)
   updatedAt             DateTime                     @default(now()) @updatedAt @db.Timestamp(6)
   start                 DateTime                     @db.Timestamp(6)
+  // Scheduled duration of the appointment
   duration              Int // in minutes
+  // This is the duration in minutes in which the host + someone else were in the meeting together
```
Jonasdoubleyou: + only set for match appointments

web/controllers/zoomController/index.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const logger = getLogger('ZoomWebhook');
+
+const webhookSigningKey = process.env.ZOOM_WEBHOOK_SECRET_TOKEN;
+
+export const zoomRouter = Router();
+
+zoomRouter.post('/event', async (req: WithRawBody<Request>, res) => {
+    logger.info('Request at /zoom/event');
+    try {
```
Jonasdoubleyou, line 18: if (!webhookSigningKey) return; - i.e. prevent that the endpoint is open if the env is missing

Jonasdoubleyou (dismissed):

PR #1423, opened 2026-06-25
Feat: Script to correct old lectures data

JeangelLF:
## What was done?

- Added a script to correct the participant data for older lectures where our records were inaccurate. Zoom only provides us with participant data for meetings from a certain date onwards
- Since this migration affects a bit 2k lectures, I implemented a kind of conservative approach that processes lectures in chunks of 5
- Added retry logic to handle potential Zoom rate-limit responses
- Tested it locally and it worked nicely with around 50 lectures
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 

Happy to get some feedback / suggestions to improve this @Jonasdoubleyou! 🙌


jobs/manual/sync_old_match_lectures_with_zoom.ts
```diff
@@ -0,0 +1,113 @@
...
+import { getLogger } from '../../common/logger/logger';
+import { ZoomError } from '../../common/util/error';
+import { getSharedMeetingTimeInSeconds } from '../../common/zoom/util';
+
+const logger = getLogger('SyncOldMatchLecturesWithZoom');
+const CONCURRENCY = 5;
+
+export async function syncOldMatchLecturesWithZoom(): Promise<void> {
+    const lectures = (await prisma.$queryRaw`
+        SELECT lecture.*
```
Jonasdoubleyou: Do you need all fields? You currently fetch the whole lecture table into tue backend which is quite heavy ...

jobs/manual/sync_old_match_lectures_with_zoom.ts
```diff
@@ -0,0 +1,113 @@
...
+            AND student."zoomUserId" IS NOT NULL
+        ORDER BY lecture."start" ASC
+    `) as Appointment[];
+
+    logger.info(`Found ${lectures.length} lectures to sync with Zoom.`);
+    const migratedLectures: string[] = [];
+    const failedLectures: string[] = [];
+
+    for (let i = 0; i < lectures.length; i += CONCURRENCY) {
+        const chunk = lectures.slice(i, i + CONCURRENCY);
```
Jonasdoubleyou: You could use a cursor to fetch lectures in chunks too - https://www.prisma.io/docs/orm/prisma-client/queries/pagination

jobs/manual/sync_old_match_lectures_with_zoom.ts
```diff
@@ -0,0 +1,113 @@
...
+        FROM lecture
+        INNER JOIN match ON match.id = lecture."matchId"
+        INNER JOIN student ON student.id = match."studentId"
+        WHERE lecture."createdAt" >= '2025-04-01'
+            AND lecture."isCanceled" IS FALSE
+            AND lecture."appointmentType" = 'match'
+            AND lecture."zoomMeetingId" IS NOT NULL
+            AND array_length(lecture."declinedBy", 1) IS NULL
+            AND array_length(lecture."joinedBy", 1) = 1
+            AND student."zoomUserId" IS NOT NULL
```
Jonasdoubleyou: AND actualDuration IS NULL so that the job can be run multiple times?

Jonasdoubleyou (dismissed):

jobs/manual/sync_old_match_lectures_with_zoom.ts
JeangelLF: I see, in this case I'm using `queryRaw` to fetch the lectures, so I'll do more or less the same thing as in the example but manually

jobs/manual/sync_old_match_lectures_with_zoom.ts
```diff
@@ -0,0 +1,144 @@
...
+    });
+    logger.debug(`Lecture ${appointment.id} migrated with actualDuration: ${Math.ceil(sharedTime / 60)} minutes.`);
+    return { zoomMeetingId: appointment.zoomMeetingId, migrated: true };
+}
+
+function sleep(ms: number): Promise<void> {
+    return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+void syncOldMatchLecturesWithZoom().catch((error) => {
```
Jonasdoubleyou: Remove? This would potentially stop the web worker?

You can manually start the job via GraphQL ...

Jonasdoubleyou (dismissed):

jobs/manual/sync_old_match_lectures_with_zoom.ts
JeangelLF: Dang it 🤦‍♂️ how did I freaking forgot to remove this. Thanks for catching it!

PR #1425, opened 2026-07-08
chore: Cleanup deprecated tables

JeangelLF:
## What was done?

Removed the following deprecated tables we have in our schema:
- bbb_meeting
- course_attendance_log
- course_guest
- course_participation_certificate
- jufo_verification_transmission


common/transactionlog/log.ts
```diff
@@ -31,8 +25,8 @@ type LogData<Type extends LogType> = {
     cocCancel: { studentId: number };
     cancelledCourse: { id: number };
     cancelledSubcourse: { id: number };
-    createdCourseAttendanceLog: { courseAttendanceLog: CourseAttendanceLog };
-    bbbMeeting: { bbbMeeting: BBBMeeting };
+    createdCourseAttendanceLog: { courseAttendanceLog: any };
+    bbbMeeting: { bbbMeeting: any };
```
Jonasdoubleyou: could remove these events

Jonasdoubleyou (dismissed):

common/transactionlog/log.ts
JeangelLF: Alright, then this would be the script to run before the release, so the migration doesn't fail. https://github.com/corona-school/dev-db/commit/c9778cbb3beb644a937ceb0d28da00e3b60080f2

Jonasdoubleyou (approved):

PR #1437, opened 2026-07-30
Feat: New student fields for screeners

JeangelLF:
## What was done?

- Added two new fields `screeningTags` / `furtherTrainingsAttendedCount` that will be used in the screening tool
- Added `maxParallelMatches` to limit the amount of parallel matches a student can have.
- Updated the seed script to include more cooperation students / fix some small issues with it


graphql/student/mutations.ts
```diff
@@ -251,6 +263,16 @@ export async function updateStudent(
         throw new PrerequisiteError('cooperationId may only be changed by elevated users');
     }
 
+    if (screeningTags !== undefined && !isElevated(context)) {
```
Jonasdoubleyou: != undefined to include null - you also want to prohibit the user from deleting it :)

Jonasdoubleyou (dismissed):

graphql/student/mutations.ts
JeangelLF: Good catch, thank you!

PR #1444, opened 2026-08-12
feat: Add message flags

JeangelLF:
## What was done?

We want to flag users that send innappropiate messages in our chat or suggest switching to a different platform.
I decided to go with https://github.com/2Toad/Profanity which already has a decent amount of bad words in different languages, we'll just have to add/remove as needed to prevent false-positives.


web/controllers/chatNotificationController/util.ts
```diff
@@ -31,6 +36,17 @@ export async function verifyChatUser(user: User) {
     return false;
 }
 
+export function getMessageFlag(message: string) {
+    const EMAIL_REGEX = /\b[A-Za-z0-9._%+-]+\s*@\s*(?!lern-fair\s*\.\s*de\b)[A-Za-z0-9.-]+\s*\.\s*[A-Za-z]{2,}\b/i;
+    const EXTERNAL_PLATFORM_REGEX = /\b(whats\s?-?app|discord|telegram|skype|snapchat|instagram|tiktok|facebook|messenger|threema|imessage|facetime)\b/i;
```
Jonasdoubleyou, line 44: "Welche Messenger verwendest du so?"

... not sure about false-positive rate, maybe restrict to domains / links?

Jonasdoubleyou (dismissed):

web/controllers/chatNotificationController/util.ts
JeangelLF: Yeah that could be an option. We'd like to see how it behaves in a few days and if we get too many false-positives then that would be the way to go :)

PR #1445, opened 2026-08-19
fix: Add check when removing a match request

JeangelLF:
## What was done?

- Today I saw in our [logs](https://app.datadoghq.eu/logs?query=message%3A%22Deleted%20match%20request%20for%20student%2C%20now%20has%22&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1786957306364&to_ts=1787130106364&live=true) that somehow one of our screeners while testing managed to set the `openMatchRequestCount` to `-1` which kind of broke the UI. Here I'm adding an extra condition when doing the update. Unfortunately I wasn't able at all to reproduce the issue 
- Also updated some logs formatting that could be useful in the future


Jonasdoubleyou (approved):
someone had good timing ...

PR #1447, opened 2026-08-26
Feat: Microfeedback

JeangelLF:
## What was done?

- Added a new `lecture_feedback` table 
- Added needed fields, queries and mutations to submit, dismiss feedback requests and list their feedback. For now this only will be done for matches
- Create or delete microfeedback when (matches)-appointments are created or cancelled respectively


Jonasdoubleyou (approved):
nice!

